PR #4, opened 2020-08-28
go-fmt.sh: never run in a container

dhellmann:
Fix `make fmt` so it works in CI. We can redo the local container version
later, if someone really needs it.


dhellmann:
/cc @sadasu @deads2k 

We probably should have added the jobs before the code. One more fix...

PR #5, opened 2020-09-01
add tools for running unit tests in CI

dhellmann:
We need a wrapper script to download and install kubebuilder because
the CI image doesn't include those tools. These scripts are copied
from the machine-api-operator repository.


dhellmann:
/cc @andfasano 

The alternative to this is to have the CI job define a separate image, but this is the same approach being used by the MAO and seems less complicated.

PR #6, opened 2020-09-01
Adding minimal logic to Reconcile method and setting up Scheme for future use

sadasu:
Another step towards a minimally functioning operator to be deployed by CVO.

### DIFF
--- a/controllers/provisioning_controller.go
+++ b/controllers/provisioning_controller.go
@@ -38,7 +38,7 @@ type ProvisioningReconciler struct {
 	// that reads objects from the cache and writes to the apiserver
 	Client client.Client
 	Scheme *runtime.Scheme
-	Log    logr.Logger
+	Log logr.Logger
 }
 
 // +kubebuilder:rbac:groups=metal3.io,resources=provisionings,verbs=get;list;watch;create;update;patch;delete
@@ -55,21 +55,19 @@ func (r *ProvisioningReconciler) Reconcile(req ctrl.Request) (ctrl.Result, error
 		Name: "cluster",
 	}, infra)
 	if err != nil {
-		log.Info("Unable to determine Platform that the Operator is running on.")
+		log.Error(err, "unable to determine Platform")
 		return ctrl.Result{}, err
 	}
 
+	log.V(1).Info("CBO is running on Platform ", infra.Status.Platform)
+
 	// Disable ourselves on platforms other than bare metal
 	if infra.Status.Platform != osconfigv1.BareMetalPlatformType {
-		log.Info("Not Baremetal Platform. Setting CBO to disabled state.")
-		// set ClusterOperator status to Disabled.
-		if err != nil {
-			return ctrl.Result{}, err
-		}
+		log.V(1).Info("setting CBO to disabled state in current Platform")
+		// TODO: Set ClusterOperator status to Disabled.
 		// We're disabled; don't requeue
 		return ctrl.Result{}, nil
 	}
-	log.Info("CBO is running on Baremetal Platform.")
 
 	return ctrl.Result{}, nil
 }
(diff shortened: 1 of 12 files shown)


dhellmann:
/assign @dhellmann 

Let's get https://github.com/openshift/release/pull/11500 in place before we merge this.

controllers/provisioning_controller.go
```diff
@@ -22,15 +22,32 @@ import (
...
+
 // ProvisioningReconciler reconciles a Provisioning object
 type ProvisioningReconciler struct {
-	client.Client
-	Log    logr.Logger
+	// This client, initialized using mgr.Client() above, is a split client
+	// that reads objects from the cache and writes to the apiserver
+	Client client.Client
 	Scheme *runtime.Scheme
+	Log    logr.Logger
```
dhellmann, line 42: Does the logger need to be public? Is anything outside of the reconciler going to use it?

controllers/provisioning_controller.go
```diff
@@ -39,10 +56,29 @@ type ProvisioningReconciler struct {
...
-	_ = r.Log.WithValues("provisioning", req.NamespacedName)
+	ctx := context.Background()
+	log := r.Log.WithValues("provisioning", req.NamespacedName)
+
+	infra := &osconfigv1.Infrastructure{}
+	err := r.Client.Get(ctx, client.ObjectKey{
+		Name: "cluster",
+	}, infra)
+	if err != nil {
+		log.Info("Unable to determine Platform that the Operator is running on.")
```
dhellmann, line 67: This log message should include the error that caused the failure.

controllers/provisioning_controller.go
```diff
@@ -39,10 +56,29 @@ type ProvisioningReconciler struct {
...
+		return ctrl.Result{}, err
+	}
 
-	// your logic here
+	// Disable ourselves on platforms other than bare metal
+	if infra.Status.Platform != osconfigv1.BareMetalPlatformType {
+		log.Info("Not Baremetal Platform. Setting CBO to disabled state.")
```
dhellmann, line 73: This log message should include the value of the platform. If you log before testing, then the log message on line 81 isn't needed.

main.go
```diff
@@ -74,6 +74,18 @@ func main() {
 	}
 	// +kubebuilder:scaffold:builder
 
+	setupLog.Info("Registering Components.")
```
dhellmann, line 77: We should consider a style guide for log messages, so they're consistent. I notice this uses caps and punctuation, but line 89 doesn't. I like keeping them stripped down, myself, but could go either way. Thoughts?

controllers/provisioning_controller.go
```diff
@@ -39,10 +56,29 @@ type ProvisioningReconciler struct {
...
+		Name: "cluster",
+	}, infra)
+	if err != nil {
+		log.Info("Unable to determine Platform that the Operator is running on.")
+		return ctrl.Result{}, err
+	}
 
-	// your logic here
+	// Disable ourselves on platforms other than bare metal
+	if infra.Status.Platform != osconfigv1.BareMetalPlatformType {
```
dhellmann, line 72: This behavior could be tested with a unit test. Let's add a basic test. I know kubebuilder is recommending ginkgo, but we can just use normal test functions and packages like we do in the other metal3 components for now and consider ginkgo later.

dhellmann (changes requested):

controllers/provisioning_controller.go
sadasu: Yes, please check the latest now.

controllers/provisioning_controller.go
sadasu: It was put into Reconciler by kubebuilder. Just the formatting changed in this PR. I'll leave a TODO here for us to see if there is a reason why it needs to be part of the Reconciler struct.

controllers/provisioning_controller.go
sadasu: Done.

main.go
sadasu: This was added by me and the one in line 89 was added by kubebuilder. I'll make it consistent with the one added by kubebuilder.

PR #8, opened 2020-09-04
Add ClusterOperator resource

kirankt:
This PR adds the clusteroperator resource and corresponding functions to create/update the CBO ClusterOperator.


sadasu (changes requested):
We divided the CO related changes into 3 tasks. In the 1st task, all we need is to create the CO and set the CO in Disabled state when the platform is not Baremetal. We can add the other CO states later. Can we add unit tests to test this basic functionality?

kirankt:
> We divided the CO related changes into 3 tasks. In the 1st task, all we need is to create the CO and set the CO in Disabled state when the platform is not Baremetal. We can add the other CO states later. Can we add unit tests to test this basic functionality?

Sure thing. This was a first run commit. Right now I'm fighting go mod to vendor the correct packages and running into lots of issues. Once sorted, I'll add the tasks.

controllers/clusteroperator.go
```diff
@@ -0,0 +1,229 @@
...
+
+// StatusReason is a MixedCaps string representing the reason for a
+// status condition change.
+type StatusReason string
+
+const (
+	clusterOperatorName = "baremetal"
+
+	// CO Status Condition Types
+	Available   osconfigv1.ClusterStatusConditionType = osconfigv1.OperatorAvailable
```
dhellmann, line 28: Can we just use the existing constants from the osconfigv1 package instead of defining and exporting local versions of them here?

controllers/provisioning_controller.go
```diff
@@ -36,9 +38,11 @@ const (
 type ProvisioningReconciler struct {
 	// This client, initialized using mgr.Client() above, is a split client
 	// that reads objects from the cache and writes to the apiserver
-	Client client.Client
-	Scheme *runtime.Scheme
-	Log    logr.Logger
+	Client        client.Client
+	Scheme        *runtime.Scheme
+	Log           logr.Logger
+	osClient      osclientset.Interface
```
dhellmann, line 44: I don't see anything populating `osClient`. Where does that happen?

controllers/clusteroperator.go
```diff
@@ -0,0 +1,229 @@
...
+	}
+
+	co, err := r.client.ConfigV1().ClusterOperators().Create(context.Background(), defaultCO, metav1.CreateOptions{})
+	if err != nil {
+		return nil, err
+	}
+
+	co.Status = defaultCO.Status
+
+	return r.client.ConfigV1().ClusterOperators().UpdateStatus(context.Background(), co, metav1.UpdateOptions{})
```
dhellmann, line 95: This line uses `r.client` but line 100 has `r.osClient` also accessing the cluster operator type. Are the two clients the same? Do we need both?

kirankt: Just go all the make targets to compile correctly. There were a lot of errors. Still iterating, but this particular one should be resolved.

controllers/clusteroperator.go
kirankt: Sure. I think it made it easier to read but I'm indifferent. Will revert.

controllers/provisioning_controller.go
kirankt: This one needs to go when we initialize the reconciler. Not sure whether the reconcile code is in a sturdy state yet to add stuff there...

kirankt: Added this logic in the latest commit.

dhellmann (changes requested):
This does a lot more than I expected just to create the new ClusterOperator resource with an initial state.

Could you split the vendored code into a separate commit to make it easier to review?

kirankt:
> This does a lot more than I expected just to create the new ClusterOperator resource with an initial state.
> 
> Could you split the vendored code into a separate commit to make it easier to review?

Sure thing. The vendored code is required or else the tests were failing. I hadn't anticipated that either. I'll split the commits into 2 for better review.

main.go
```diff
@@ -19,21 +19,31 @@ import (
...
 
 	osconfigv1 "github.com/openshift/api/config/v1"
 )
 
 var (
-	scheme   = runtime.NewScheme()
-	setupLog = ctrl.Log.WithName("setup")
+	scheme             = runtime.NewScheme()
+	setupLog           = ctrl.Log.WithName("setup")
+	componentNamespace = "openshift-machine-api"
```
sadasu, line 42: Already defined in provisioning_controller.go here: https://github.com/openshift/cluster-baremetal-operator/pull/8/files#diff-1002c9504e54348b5431f65cbd2cc108R34.

main.go
```diff
@@ -19,21 +19,31 @@ import (
...
+)
+
+const (
+	componentName = "cluster-baremetal-operator"
```
sadasu, line 46: This is also defined in provisioning_controller.go

controllers/clusteroperator.go
```diff
@@ -0,0 +1,222 @@
...
+
+	conds := []osconfigv1.ClusterOperatorStatusCondition{
+		setStatusCondition(osconfigv1.OperatorAvailable, osconfigv1.ConditionTrue, string(ReasonComplete), message),
+	}
+
+	return r.syncStatus(co, conds)
+}
+
+// updateCOStatusDegraded updates the ClusterOperator's status to Degraded
+func (r *ProvisioningReconciler) updateCOStatusDegraded() error {
```
sadasu, line 188: Can we add this method here when we are ready to call it from Reconcile()? We can also add unit tests at that time.

controllers/clusteroperator_test.go
```diff
@@ -0,0 +1,55 @@
...
+		for _, expectedCondition := range tc.expectedConditions {
+			ok := v1helpers.IsStatusConditionPresentAndEqual(
+				gotCO.Status.Conditions, expectedCondition.Type, expectedCondition.Status,
+			)
+			if !ok {
+				assert.Regexp(t, tc.expected, ok)
+			}
+		}
+	}
+}
```
sadasu, line 55: We can keep updateCOStatusAvailable() in clusteroperator.go. Can we also add a test case to test the available state?

controllers/provisioning_controller.go
```diff
@@ -78,7 +82,12 @@ func (r *ProvisioningReconciler) Reconcile(req ctrl.Request) (ctrl.Result, error
...
 	}
 	if !enabled {
-		// TODO: Set ClusterOperator status to Disabled.
+		// set ClusterOperator status to disabled=true, available=true
+		err = r.updateCOStatusDisabled()
+		if err != nil {
+			return ctrl.Result{}, err
+		}
+
 		// We're disabled; don't requeue
```
sadasu, line 91: If updateCOStatusAvailable() is going to be part of this commit, we should set the CO status to Available here when the Platform type is Baremetal.

controllers/clusteroperator.go
kirankt: These snuck back into the latest commit after I split the commits a short while ago. I'll remove this since we are not using it in reconcile at this time.

controllers/clusteroperator.go
```diff
@@ -0,0 +1,222 @@
...
+	for _, c := range conds {
+		v1helpers.SetStatusCondition(&co.Status.Conditions, c)
+	}
+
+	_, err := r.OSClient.ConfigV1().ClusterOperators().UpdateStatus(context.Background(), co, metav1.UpdateOptions{})
+	return err
+}
+
+// updateCOStatusProgressing updates the ClusterOperator's status to Progressing
+func (r *ProvisioningReconciler) updateCOStatusProgressing() error {
```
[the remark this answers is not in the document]
kirankt: These snuck back into the latest commit after I split the commits a short while ago. I'll remove this since we are not using it in reconcile at this time.

main.go
kirankt: This one is not needed in this file. Can be removed.

main.go
kirankt: But it starts with lower-case 'c', which is not available outside the package. But I'll convert it to 'ComponentName' in the controllers package.

controllers/clusteroperator_test.go
kirankt: removing available from this commit

config/crd/bases/metal3.io_provisionings.yaml
```diff
@@ -4,7 +4,7 @@ apiVersion: apiextensions.k8s.io/v1beta1
 kind: CustomResourceDefinition
 metadata:
   annotations:
-    controller-gen.kubebuilder.io/version: v0.2.5
+    controller-gen.kubebuilder.io/version: v0.2.4
```
kirankt, line 7: @dhellmann This change was due to the fact that we're asking make generate to fetch v0.2.4 of controller-gen, if none was found locally. Do you want me to change this to v0.2.5 in the Makefile?

kirankt: I also think we should not look for a local copy of controller-gen in the Makefile. We should just stick to our blessed vendored version. Or else this value might get changed a lot from PR to PR. What do you think?

kirankt:
Rebased code. Please review when you get a chance.
cc: @sadasu, @dhellmann

config/crd/bases/metal3.io_provisionings.yaml
sadasu: Can this change be left out of this PR?

kirankt: +1. Will remove it. Thanks.

controllers/clusteroperator.go
```diff
@@ -0,0 +1,152 @@
...
+// getOperandVersions returns current version
+func getOperandVersions() []osconfigv1.OperandVersion {
+	operandVersions := []osconfigv1.OperandVersion{}
+	if releaseVersion := os.Getenv("RELEASE_VERSION"); len(releaseVersion) > 0 {
+		operandVersions = append(operandVersions, osconfigv1.OperandVersion{Name: "operator", Version: releaseVersion})
+	}
+	return operandVersions
+}
+
+//syncStatus applies the new condition to the mao ClusterOperator object.
```
sadasu, line 125: mao -> cbo

controllers/clusteroperator.go
```diff
@@ -0,0 +1,152 @@
...
+	if err != nil {
+		return nil, err
+	}
+
+	co.Status = defaultCO.Status
+	return r.OSClient.ConfigV1().ClusterOperators().UpdateStatus(context.Background(), co, metav1.UpdateOptions{})
+}
+
+// getOrCreateClusterOperator gets the existing CO, failing which it creates a new CO.
+func (r *ProvisioningReconciler) getOrCreateClusterOperator() (*osconfigv1.ClusterOperator, error) {
```
sadasu, line 89: Can we add a unit -test for the case where we are able to get an existing CO when it exists?

controllers/clusteroperator.go
kirankt: Doh.

controllers/clusteroperator.go
kirankt: Done. Thanks.

controllers/clusteroperator_test.go
```diff
@@ -0,0 +1,191 @@
...
+			osClient = fakeconfigclientset.NewSimpleClientset()
+		}
+		reconciler := newFakeProvisioningReconciler(setUpSchemeForReconciler(), &osconfigv1.Infrastructure{})
+		reconciler.OSClient = osClient
+
+		co, err := reconciler.getOrCreateClusterOperator()
+		if err != nil {
+			t.Fatalf("unexpected error: %v", err)
+		}
+		normalizeTransitionTimes(co.Status, tc.expectedCO.Status)
```
sadasu, line 173: Did not expect that we would have to do that!

sadasu (approved):

PR #10, opened 2020-09-08
Read Provisioning CR

sadasu:


controllers/provisioning_controller_test.go
```diff
@@ -97,3 +98,65 @@ func TestReconcile(t *testing.T) {
...
+					APIVersion: "v1",
+				},
+				ObjectMeta: metav1.ObjectMeta{
+					Name: baremetalProvisioningCR,
+				},
+			},
+			expectedError:  false,
+			expectedConfig: true,
+		},
+		{
```
dhellmann, line 125: What about a test for a missing resource to handle the delete case?

sadasu: Test case "EmptyCR" is testing that case. The  call to Client.Get does return an apierrors.IsNotFound(err) which would be the case when the CR is missing. The test case could be named better to indicate the case we are testing (rather that the way were are achieving it with an empty CR.)

dhellmann: Is the MissingCR case really different from the InvalidName case as it is implemented? Don't we want the test to not create any Provisioning CRs at all for this test case?

dhellmann: I guess the query is still asking for the "right" name so the fact that it isn't there because we don't create that one is a good test.

PR #12, opened 2020-09-11
Set fixed controller gen

kirankt:
This PR ensures we use a fixed, vendored controller-gen rather than a fetched or local copy. It also maintains parity with MAO vendored controller-gen tool.

The generate-check target also checks the manifests target rather than the code generation target.


dhellmann:
/approve

It looks like this needs to be rebased.

kirankt:
Rebase done. PTAL. Thanks.
cc: @dhellmann , @sadasu

PR #15, opened 2020-09-16
Validate contents of Provisioning CR

sadasu:
Basic validations for different ProvisioningNetwork modes.


controllers/baremetal_config.go
```diff
@@ -0,0 +1,84 @@
...
+		config.Spec.ProvisioningNetworkCIDR == "" ||
+		config.Spec.ProvisioningDHCPRange == "" ||
+		config.Spec.ProvisioningOSDownloadURL == "" {
+		return fmt.Errorf("configuration missing in Managed ProvisioningNetwork mode in config resource %s", baremetalProvisioningCR)
+	}
+	return nil
+}
+
+func validateUnmanagedConfig(config *metal3iov1alpha1.Provisioning) error {
+	// All values except ProvisioningDHCPRange must be provided in the provisioning CR
```
kirankt, line 66: Change this comment.

controllers/baremetal_config.go
```diff
@@ -0,0 +1,84 @@
...
+		config.Spec.ProvisioningIP == "" ||
+		config.Spec.ProvisioningNetworkCIDR == "" ||
+		config.Spec.ProvisioningOSDownloadURL == "" {
+		return fmt.Errorf("configuration missing in Unmanaged ProvisioningNetwork mode in config resource %s", baremetalProvisioningCR)
+	}
+	return nil
+}
+
+func validateDisabledConfig(config *metal3iov1alpha1.Provisioning) error {
+	// All values except ProvisioningDHCPRange must be provided in the provisioning CR
```
kirankt, line 77: Same here. Comments need to be changed.

controllers/baremetal_config.go
```diff
@@ -0,0 +1,84 @@
...
+		glog.V(1).Info("Provisioning Network Disabled")
+		return validateDisabledConfig(config)
+	}
+	return nil
+}
+
+func validateManagedConfig(config *metal3iov1alpha1.Provisioning) error {
+	// All values must be provided in the provisioning CR
+	if config.Spec.ProvisioningInterface == "" ||
+		config.Spec.ProvisioningIP == "" ||
```
kirankt, line 56: We should also check whether the value specified is a valid IP. Something simple like:
e.g. https://github.com/openshift/installer/blob/b894e4c734366194b4773ca4bbcf90a249c2a7e4/pkg/types/baremetal/validation/platform.go#L126

controllers/baremetal_config.go
```diff
@@ -0,0 +1,84 @@
...
+		config.Spec.ProvisioningNetworkCIDR == "" ||
```
kirankt, line 57: Same here for CIDR. We should check at least whether the input is in the form of a CIDR in addition to an empty string.
Something simple like this:
https://github.com/openshift/installer/blob/b894e4c734366194b4773ca4bbcf90a249c2a7e4/pkg/types/baremetal/validation/platform.go#L36

controllers/baremetal_config.go
kirankt, line 56: We should also check whether the value specified is a valid IP. Something simple like:
e.g. https://github.com/openshift/installer/blob/b894e4c734366194b4773ca4bbcf90a249c2a7e4/pkg/types/baremetal/validation/platform.go#L126

controllers/baremetal_config.go
kirankt, line 66: Change this comment.

controllers/baremetal_config.go
kirankt, line 77: Change this comment also.

controllers/baremetal_test.go
```diff
@@ -0,0 +1,291 @@
...
+					ProvisioningIP:            "172.30.20.3",
+					ProvisioningNetworkCIDR:   "172.30.20.0/24",
+					ProvisioningDHCPRange:     "172.30.20.11, 172.30.20.101",
+					ProvisioningOSDownloadURL: "http://172.22.0.1/images/rhcos-44.81.202001171431.0-openstack.x86_64.qcow2.gz?sha256=e98f83a2b9d4043719664a2be75fe8134dc6ca1fdbde807996622f8cc7ecd234",
+					ProvisioningNetwork:       "Managed",
+				},
+			},
+			expectedError: true,
+			expectedMode:  provisioningNetworkManaged,
+		},
```
kirankt, line 123: We should probably add test cases for broken IP and CIDR inputs here.

kirankt (changes requested):
Left a few comments. In short we need to validate IPs and CIDRs in addition to empty strings and also test them. Let me know what you think.

controllers/baremetal_config.go
```diff
@@ -0,0 +1,84 @@
...
+		config.Spec.ProvisioningDHCPRange == "" ||
+		config.Spec.ProvisioningOSDownloadURL == "" {
```
kirankt, line 59: We should also validate for correct URL syntax.

controllers/baremetal_config.go
sadasu: Removed comment.

dhellmann (changes requested):

controllers/baremetal_config.go
```diff
@@ -0,0 +1,84 @@
...
+
+const (
+	provisioningNetworkManaged   = "Managed"
+	provisioningNetworkUnmanaged = "Unmanaged"
+	provisioningNetworkDisabled  = "Disabled"
+)
+
+func validateBaremetalProvisioningConfig(config *metal3iov1alpha1.Provisioning) error {
+	if config.Spec.ProvisioningNetwork == "" {
+		config.Spec.ProvisioningNetwork = provisioningNetworkManaged
```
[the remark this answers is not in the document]
sadasu: Yes, there is a need to have a Config type that wraps the CRD type. Since, these changes have only one instance (support for DHCP external and default behavior) I did not add it in the PR. The containers within metal3 need other config like the ironic inspector endpoint, that would be provided by this wrapper Config type using values in the CRD type. 
Can the Config type be handled in a follow up PR?

controllers/baremetal_test.go
```diff
@@ -0,0 +1,291 @@
...
+
+	metal3iov1alpha1 "github.com/openshift/cluster-baremetal-operator/api/v1alpha1"
+	"github.com/stretchr/testify/assert"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+func TestValidateManagedProvisioningConfig(t *testing.T) {
+	tCases := []struct {
+		name          string
+		baremetalCR   *metal3iov1alpha1.Provisioning
```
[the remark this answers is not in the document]
sadasu: Agree.

controllers/provisioning_controller.go
```diff
@@ -128,6 +128,12 @@ func (r *ProvisioningReconciler) Reconcile(req ctrl.Request) (ctrl.Result, error
 		// Cannot proceed wtih metal3 deployment.
 		return ctrl.Result{}, nil
 	}
+	if err := validateBaremetalProvisioningConfig(baremetalConfig); err != nil {
+		// Provisioning configuration is not valid.
+		// Requeue request.
+		return ctrl.Result{}, err
```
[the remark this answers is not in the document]
sadasu: Ok. Yes, we will put the ClusterOperator in a degraded state which is coming as a follow-up PR. I'll just log an error and return no error for this change.

controllers/baremetal_test.go
```diff
@@ -0,0 +1,291 @@
...
+	}
+	for _, tc := range tCases {
+		t.Run(tc.name, func(t *testing.T) {
+			t.Logf("Testing tc : %s", tc.name)
+			err := validateBaremetalProvisioningConfig(tc.baremetalCR)
+			if !tc.expectedError && err != nil {
+				t.Errorf("unexpected error: %v", err)
+				return
+			}
+			assert.Equal(t, tc.expectedMode, tc.baremetalCR.Spec.ProvisioningNetwork, "enabled results did not match")
```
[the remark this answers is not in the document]
sadasu: Ok.Will update code to not update the Spec and modify test accordingly.

controllers/baremetal_config.go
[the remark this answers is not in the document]
dhellmann: This is more than a little nit, so I think we should fix it as part of this PR. That may mean a simple function for now, or, since nothing uses the value outside of this function just using a local variable instead of updating the Spec field.

controllers/baremetal_config.go
```diff
@@ -0,0 +1,84 @@
+/*
```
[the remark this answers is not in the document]
sadasu: Since we expect the number of validations to grow, I can move the validations to its own package.

controllers/baremetal_config.go
```diff
@@ -0,0 +1,84 @@
...
+}
+
+func validateManagedConfig(config *metal3iov1alpha1.Provisioning) error {
+	// All values must be provided in the provisioning CR
+	if config.Spec.ProvisioningInterface == "" ||
+		config.Spec.ProvisioningIP == "" ||
+		config.Spec.ProvisioningNetworkCIDR == "" ||
+		config.Spec.ProvisioningDHCPRange == "" ||
+		config.Spec.ProvisioningOSDownloadURL == "" {
+		return fmt.Errorf("configuration missing in Managed ProvisioningNetwork mode in config resource %s", baremetalProvisioningCR)
```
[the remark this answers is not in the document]
sadasu: Done.

controllers/baremetal_config.go
```diff
@@ -0,0 +1,84 @@
...
+
+import (
+	"fmt"
+
+	"github.com/golang/glog"
+	metal3iov1alpha1 "github.com/openshift/cluster-baremetal-operator/api/v1alpha1"
+)
+
+const (
+	provisioningNetworkManaged   = "Managed"
```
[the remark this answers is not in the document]
sadasu: @dhellmann I added enum values for ProvisioningNetwork in commit sadasu@de28b9f. As expected, this resulted in a change in the CRD. I am not sure if this will cause any issues during upgrade from the CRD in MAO in 4.6 to this CRD in 4.7.

controllers/baremetal_config.go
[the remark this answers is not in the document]
sadasu: Resolved as part of commit https://github.com/sadasu/cluster-baremetal-operator/commit/b73ff586f75bdf424553bc8f5bfe6c92716fc04e.

controllers/baremetal_config.go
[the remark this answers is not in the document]
dhellmann: Maybe for now we should just define them as string constants, then? Eventually we'll have to update that CRD, and we can start making a list of the improvements like this that we want to include when the time comes.

provisioning/baremetal_config.go
```diff
@@ -0,0 +1,84 @@
...
+const (
+	// BaremetalProvisioningCR is the name of the provisioning config resource.
+	BaremetalProvisioningCR = "provisioning-configuration"
+)
+
+// ValidateBaremetalProvisioningConfig validates the contents of the baremetal Provisioning CR.
+func ValidateBaremetalProvisioningConfig(prov *metal3iov1alpha1.Provisioning) error {
+	provisioningNetworkMode := getProvisioningNetworkMode(prov)
+	switch provisioningNetworkMode {
+	case metal3iov1alpha1.ProvisioningNetwork("Managed"):
```
dhellmann, line 35: You can refer to the const values directly.

```suggestion
	case metal3iov1alpha1.Managed:
```

That may mean that we want to make those names more verbose. `ProvisioningNetworkManaged`, for example.

provisioning/baremetal_config.go
```diff
@@ -0,0 +1,84 @@
...
+		glog.V(1).Info("Provisioning Network Managed")
```
dhellmann, line 36: This log call could be done once outside of the switch statement with a %s to fill in the actual value.

provisioning/baremetal_config.go
```diff
@@ -0,0 +1,84 @@
...
+	for _, item := range slice {
+		if item == value {
+			return true
+		}
+	}
+	return false
+}
+
+func validateConfig(prov *metal3iov1alpha1.Provisioning, optionalConfig []string) error {
+	specConfig := reflect.ValueOf(prov.Spec)
```
dhellmann, line 75: Using reflection here feels like an unnecessary obfuscation of what the function is doing. The earlier structure with inline references to the fields was fine, it just needed to be restructured to allow it to report the name of the missing value. That could be done clearly using a slice of a struct with a field name and a field value.

```

for _, toTest := range []struct{
    Name string
    Value string
  }{
    {Name: "ProvisioningInterface", Value: prov.Spec.ProvisioningInterface},
  }{
  if toTest.Value == "" {
    return fmt.Errorf("%s is required but is empty", toTest.Name)
  }
}
```

go.mod
```diff
@@ -5,6 +5,7 @@ go 1.13
 require (
 	github.com/go-logr/logr v0.2.1-0.20200730175230-ee2de8da5be6
 	github.com/go-logr/zapr v0.2.0 // indirect
+	github.com/golang/glog v0.0.0-20160126235308-23def4e6c14b
```
dhellmann, line 8: We already have 2 logging packages. Do we need this new one?

kirankt: I think this one can be removed.

sadasu: zapr is indirect so glog would be the 2nd actually. I found logr to be tricky while using file level logging vs glog. So, just didn't spend enough time on it. Object level logging using logr in controller_provisioning.go was a bit simpler.

provisioning/baremetal_config.go
sadasu: Not sure how I missed that. Done now.

provisioning/baremetal_config.go
sadasu: Done.

provisioning/baremetal_config.go
sadasu: The reason I decided to use reflect are the following: 
If and when we add new fields to the Spec, it could be of any type (not just string). For example, our earlier change to make ProvisioningNetwork into a enum type instead of string might break above code (haven't actually tested it). Using reflect would allow us to introspect our own data structure and do the right thing based on the type of the field in the struct.
Also, when the Provisioning API changes (and specifically the Spec field), all we will have to do is change the allowedAbsent slice per mode.

If you still feel that this is unnecessary, I am happy to use something similar to what is suggested.

dhellmann: If we end up having to compare different types, I think I would rather just use separate conditional statements outside of a loop. Kiran pointed out on an earlier review that there are additional checks we need to do besides ensuring that values are present, and when we start adding those it doesn't seem likely that using reflection would be necessary or helpful.

Let's try to keep the implementation as simple as possible, even if it means a little repetition.

sadasu: Made it simpler by having a separate validation function for each ProvisioningNetwork mode.

go.mod
sadasu: Imported controller-runtime for its logger so we don't pull in glog as a new logging package.

provisioning/baremetal_config.go
```diff
@@ -0,0 +1,116 @@
...
+import (
+	"fmt"
+
+	metal3iov1alpha1 "github.com/openshift/cluster-baremetal-operator/api/v1alpha1"
+	ctrl "sigs.k8s.io/controller-runtime"
+)
+
+const (
+	// BaremetalProvisioningCR is the name of the provisioning resource
+	BaremetalProvisioningCR = "provisioning-configuration"
```
dhellmann, line 27: Requiring this specific name is a characteristic of the controller, rather than the resource type, right? It seems like the only use inside this package are for the logger, and we could skip that, and the unit tests where it shouldn't matter. I think I would leave it back in the controller package.

provisioning/baremetal_config.go
```diff
@@ -0,0 +1,116 @@
...
+
+var (
+	log = ctrl.Log.WithName("provisioning").WithName(BaremetalProvisioningCR)
+)
+
+// ValidateBaremetalProvisioningConfig validates the contents of the provisioning resource
+func ValidateBaremetalProvisioningConfig(prov *metal3iov1alpha1.Provisioning) error {
+	provisioningNetworkMode := getProvisioningNetworkMode(prov)
+	log.V(1).Info("Final Provisioning Network", "Mode", provisioningNetworkMode)
+	switch provisioningNetworkMode {
```
dhellmann, line 38: We don't specify an enum, yet, so this value isn't guaranteed to be one of the 3 valid values. We'll need a default handler here to report an error, in case someone uses an invalid value.

provisioning/baremetal_config.go
sadasu: It has been moving around. Yes, it makes sense that the name of the resource is a function of the controller. Made the necessary changes.

provisioning/baremetal_config.go
sadasu: Added validation for Managed as the default case.

sadasu:
@kirankt I renamed the test file as you requested.

provisioning/baremetal_config.go
sadasu: After making ProvisioningNetwork and enum in the CRD, we can go back to not having a default case.

dhellmann:
/approve

This looks good. @kirankt had some comments, too, so I'll leave it open for him to review.

kirankt:
> /approve
> 
> This looks good. @kirankt had some comments, too, so I'll leave it open for him to review.

Looks good to me. Thanks.

PR #16, opened 2020-09-21
Allow ClusterOperator to be set to Degraded state

sadasu:
Set CO to Degraded state when provisioning CR validations fail

### DIFF
--- a/controllers/clusteroperator.go
+++ b/controllers/clusteroperator.go
@@ -149,3 +149,23 @@ func (r *ProvisioningReconciler) updateCOStatusDisabled() error {
 
 	return r.syncStatus(co, conds)
 }
+
+// updateCOStatusDegraded updates the ClusterOperator's Degraded
+// degradedReason should contain the current reason for the Operator to be marked in that state
+func (r *ProvisioningReconciler) updateCOStatusDegraded(degradedReason string, detailedError string) error {
+	degradedMessage := "Operator is Degraded"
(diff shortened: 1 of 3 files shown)


controllers/clusteroperator.go
```diff
@@ -149,3 +149,24 @@ func (r *ProvisioningReconciler) updateCOStatusDisabled() error {
...
+
+// updateCOStatusDegraded updates the ClusterOperator's Degraded
+// degradedReason should contain the current reason for the Operator to be marked in that state
+func (r *ProvisioningReconciler) updateCOStatusDegraded(degradedReason string, detailedError string) error {
+	degradedMessage := "Operator is Degraded"
+	progressingMessage := "Operator is Degraded while Progressing"
+
+	co, err := r.getOrCreateClusterOperator()
+	if err != nil {
+		r.Log.Error(err, "failed to get or create ClusterOperator")
```
dhellmann, line 161: Is logging here before returning the error going to mean the error is logged multiple times? Should this just wrap the error with more detail and then return it?

dhellmann (commented):
This looks good. I have one question about error handling inline.

controllers/clusteroperator.go
sadasu: Removed this error msg. getOrCreateClusterOperator() is returning a good error message that can be wrapped with one  from Reconcile() and returned from there. Doing the same for updateCOStatusDisabled().

PR #17, opened 2020-09-22
Generate the CRD with v1 (not v1alpha1)

asalkeld:
Originally done to get "oc explain Provisioning" to print more information.
(https://bugzilla.redhat.com/show_bug.cgi?id=1880787)

Update:
So I think we should merge this, just because v1alpha1 is deprecated.. 

However this is not helping 'oc explain Provisioning'.
if I "oc apply -f manifests/0000_30_cluster-baremetal-operator_00_metal3provisioning.crd.yaml"
then 'oc explain Provisioning', I still get the same <empty> response.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -3,8 +3,6 @@
 IMG ?= controller:latest
 # Controller-gen tool
 CONTROLLER_GEN ?= go run vendor/sigs.k8s.io/controller-tools/cmd/controller-gen/main.go
-# Produce CRDs that work back to Kubernetes 1.11 (no version conversion)
-CRD_OPTIONS ?= "crd:trivialVersions=true"
 
 # Get the currently used golang install path (in GOPATH/bin, unless GOBIN is set)
 ifeq (,$(shell go env GOBIN))
--- a/hack/gen-crd.sh
+++ b/hack/gen-crd.sh
@@ -4,11 +4,14 @@ set -euo pipefail
 REPO_ROOT=$(dirname "${BASH_SOURCE}")/..
 source $REPO_ROOT/hack/utils.sh
 
+# Produce CRDs that work back to Kubernetes 1.11 (no version conversion)
+CRD_OPTIONS="crd:trivialVersions=true,crdVersions=v1"
+
 # Generate manifests
-go run vendor/sigs.k8s.io/controller-tools/cmd/controller-gen/main.go "crd:trivialVersions=true" rbac:roleName=manager-role webhook paths=./... output:crd:artifacts:config=config/crd/bases
+go run vendor/sigs.k8s.io/controller-tools/cmd/controller-gen/main.go $CRD_OPTIONS rbac:roleName=manager-role webhook paths=./... output:crd:artifacts:config=config/crd/bases
 sed -i '/^    controller-gen.kubebuilder.io\/version: (devel)/d' config/crd/bases/*
 
 # Copy crds into the manifests
 install_crd \
   "config/crd/bases/metal3.io_provisionings.yaml" \
-  "manifests/0000_31_cluster-baremetal-operator_00_metal3provisioning.crd.yaml"
+  "manifests/0000_31_cluster-baremetal-operator_02_metal3provisioning.crd.yaml"
--- a/config/crd/bases/metal3.io_provisionings.yaml
+++ b/config/crd/bases/metal3.io_provisionings.yaml
@@ -1,6 +1,6 @@
 
 ---
-apiVersion: apiextensions.k8s.io/v1beta1
+apiVersion: apiextensions.k8s.io/v1
 kind: CustomResourceDefinition
 metadata:
   annotations:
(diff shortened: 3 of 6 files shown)


JoelSpeed (commented):
Since this changes your resource to include a status subresource, have you checked that you are handling updates to the status using a status client within the controller? If you aren't, your controller would stop being able to update the status once this is merged

Makefile
```diff
@@ -4,7 +4,7 @@ IMG ?= controller:latest
 # Controller-gen tool
 CONTROLLER_GEN ?= go run vendor/sigs.k8s.io/controller-tools/cmd/controller-gen/main.go
 # Produce CRDs that work back to Kubernetes 1.11 (no version conversion)
-CRD_OPTIONS ?= "crd:trivialVersions=true"
+CRD_OPTIONS ?= "crd:trivialVersions=true,crdVersions=v1"
```
dhellmann, line 7: Unfortunately, we need to keep this consistent with what is in the machine-api-operator for the time being. https://github.com/openshift/machine-api-operator/blob/master/install/0000_30_machine-api-operator_04_metal3provisioning.crd.yaml#L1 shows that is still using v1beta1.

I think that means deferring the fix for this until later in the 4.7 cycle, when we have completed the work to remove the metal3 management logic from the MAO. Unless we also need to backport the fix to 4.6, in which case I would expect a patch in the MAO repository to land before we make the change here.

api/v1alpha1/provisioning_types.go
```diff
@@ -100,10 +100,9 @@ type ProvisioningStatus struct {
 // This CR is a singleton, created by the installer and currently only
 // consumed by the cluster-baremetal-operator to bring up and update
 // containers in a metal3 cluster.
-// +kubebuilder:subresource:status
-// +kubebuilder:resource:path=provisionings,scope=Cluster
-
 // +kubebuilder:object:root=true
+// +kubebuilder:resource:path=provisionings,scope=Cluster
+// +kubebuilder:subresource:status
```
dhellmann, line 105: These changes should be OK, since they match the MAO version.

dhellmann (changes requested):

sadasu:
> Since this changes your resource to include a status subresource, have you checked that you are handling updates to the status using a status client within the controller? If you aren't, your controller would stop being able to update the status once this is merged

@JoelSpeed by status client in the controller, do you mean the ClusterOperator tracking the status of the Operator? Moreover, this PR is changing the order of the kubebuilder comments and not adding a new one. Should this result in a status change somewhere?

JoelSpeed:
@sadasu If you look at the generated CRD in this repo, it doesn't currently include the status subresource, so this generation is adding the status subresource to the CRD in the eyes of the API server.

However, if you look in the [machine-api-operator](https://github.com/openshift/machine-api-operator/blob/f304d771cd3bc40305430b6a3f1ff8ca9a56e91d/install/0000_30_machine-api-operator_04_metal3provisioning.crd.yaml#L13-L14) repo, you can see the subresource is already there in that version, so as long as that is the one people are using, there's no breaking changes or anything here and my previous comment can be ignored

Makefile
zaneb: I think the point here is to copy this into the MAO, since the explain output is currently broken in 4.6.

JoelSpeed: I believe it will be safe to copy what is currently generated in this PR over to MAO within the 4.6 timeline, can we can a PR raised for that and attached to the same bug?

sadasu: This generated CRD is different from the one in MAO specifically around https://github.com/openshift/machine-api-operator/blob/master/install/0000_30_machine-api-operator_04_metal3provisioning.crd.yaml#L98-L101. Either we have to update our code here to generate the enum or make the changes in MAO independent of these changes.

JoelSpeed: The enum you've highlighted appears to be present in the version in this PR as well? 
https://github.com/openshift/cluster-baremetal-operator/blob/2c5c0e6e9e6d8a117a802e0b9d4307af71142df5/config/crd/bases/metal3.io_provisionings.yaml#L48-L51

sadasu: Yes, https://github.com/openshift/cluster-baremetal-operator/pull/19 took care of that.

JoelSpeed: Ok, so it should be safe to copy this generated file over to MAO and make sure they're synced?

hack/gen-crd.sh
```diff
@@ -1,14 +1,12 @@
...
-source $REPO_ROOT/hack/utils.sh
+controllerGen="go run vendor/sigs.k8s.io/controller-tools/cmd/controller-gen/main.go"
 
-# Generate manifests
-go run vendor/sigs.k8s.io/controller-tools/cmd/controller-gen/main.go "crd:trivialVersions=true" rbac:roleName=manager-role webhook paths=./... output:crd:artifacts:config=config/crd/bases
-sed -i '/^    controller-gen.kubebuilder.io\/version: (devel)/d' config/crd/bases/*
+# generate the CRD
+CRD_FILE=manifests/0000_30_cluster-baremetal-operator_00_metal3provisioning.crd.yaml
+CRD_OPTIONS="crd:trivialVersions=true,crdVersions=v1"
+$controllerGen $CRD_OPTIONS paths=./... output:stdout | sed '/^    controller-gen.kubebuilder.io\/version: (devel)/d' > $CRD_FILE
```
andfasano, line 9: Any particular reason for dropping `install_crd`?

asalkeld: just seemed unnecessary to copy and have it in 2 palaces?
happy to revert this

andfasano: I think it's ok for now.
As a general comment, my personal feeling is that it would be better to keep the _raw_ generated manifests separated from the _rendered_ ones in the `/manifests` folder, anyhow we could revisit this strategy later I suppose.

asalkeld: I have reverted it for now

dhellmann:
Comparing the CRD file in this PR with the version currently in the MAO show no differences.

/approve

dhellmann:
> Comparing the CRD file in this PR with the version currently in the MAO show no differences.
> 
> /approve

Well, it does show differences if I compare the right files. :head-desk:

The file in this PR looks **correct**.

PR #18, opened 2020-09-22
Bring the CRD closer to what it should be

asalkeld:
The grouping of the +kubebuilder comments was
causing the scope and subresource to not work.


sadasu (commented):
/lgtm

sadasu (approved):

PR #19, opened 2020-09-22
Make ProvisioningNetwork values an enum

sadasu:


api/v1alpha1/provisioning_types.go
```diff
@@ -21,6 +21,16 @@ import (
 	operatorv1 "github.com/openshift/api/operator/v1"
 )
 
+// ProvisioningNetwork is the boot mode of the system
+// +kubebuilder:validation:Enum=Managed;Unmanaged;Disabled
+type ProvisioningNetwork string
+
+const (
+	provisioningNetworkManaged   = "Managed"
```
dhellmann, line 29: These consts can be defined as being of the new type. I think if you do that, you don't need line 25.

config/crd/bases/metal3.io_provisionings.yaml
```diff
@@ -43,6 +43,10 @@ spec:
               type: string
             provisioningNetwork:
               description: ProvisioningNetwork provides a way to indicate the state of the underlying network configuration for the provisioning network. This field can have one of the following values - `Managed`- when the provisioning network is completely managed by the Baremetal IPI solution. `Unmanaged`- when the provsioning network is present and used but the user is responsible for managing DHCP. Virtual media provisioning is recommended but PXE is still available if required. `Disabled`- when the provisioning network is fully disabled. User can bring up the baremetal cluster using virtual media or assisted installation. If using metal3 for power management, BMCs must be accessible from the machine networks. User should provide two IPs on the external network that would be used for provisioning services.
+              enum:
+              - Managed
+              - Unmanaged
+              - Disabled
               type: string
```
sadasu, line 50: Making the change in https://github.com/openshift/cluster-baremetal-operator/pull/19#discussion_r493020627 made these changes in 46 - 49 disappear.

api/v1alpha1/provisioning_types.go
```diff
@@ -84,7 +94,7 @@ type ProvisioningSpec struct {
 	// installation. If using metal3 for power management, BMCs must be
 	// accessible from the machine networks. User should provide two IPs on
 	// the external network that would be used for provisioning services.
-	ProvisioningNetwork string `json:"provisioningNetwork,omitempty"`
+	ProvisioningNetwork ProvisioningNetwork `json:"provisioningNetwork,omitempty"`
 }
```
sadasu, line 98: Here the type of ProvisioningNetwork has changed but in the resulting yaml file it is still a string.

sadasu:
Rebased over @asalkeld 's changes. Matches MAO https://github.com/openshift/machine-api-operator/blob/master/install/0000_30_machine-api-operator_04_metal3provisioning.crd.yaml, afaik.

api/v1alpha1/provisioning_types.go
```diff
@@ -100,12 +107,7 @@ type ProvisioningStatus struct {
 // This CR is a singleton, created by the installer and currently only
 // consumed by the cluster-baremetal-operator to bring up and update
 // containers in a metal3 cluster.
-
 // +kubebuilder:object:root=true
-// +kubebuilder:resource:path=provisionings,scope=Cluster
-// +kubebuilder:subresource:status
-
```
asalkeld, line 107: any reason for moving these back up?

sadasu: Moving this back did not affect https://github.com/openshift/cluster-baremetal-operator/blob/640297f920e13f2308d97d3b835ecf083a86cebf/config/crd/bases/metal3.io_provisionings.yaml#L16-L18.

This change was needed for the yaml file to have https://github.com/openshift/machine-api-operator/blob/master/install/0000_30_machine-api-operator_04_metal3provisioning.crd.yaml#L17-L24.

api/v1alpha1/provisioning_types.go
sadasu: I removed these constants from the latest changes because they are not necessary to generate the right yaml file. Will add them as part of the validation code that is in review.

asalkeld (approved):

PR #20, opened 2020-09-23
Install Provisioning CRD

andfasano:
This PR takes care of installing properly, as requested by the CVO, the Provisioning CRD

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,27 +1,12 @@
-# Build the manager binary
-FROM golang:1.13 as builder
+# When bumping the Go version, don't forget to update the configuration of the
+# CI jobs in openshift/release.
+FROM registry.svc.ci.openshift.org/ocp/builder:rhel-8-golang-1.15-openshift-4.6 AS builder
+WORKDIR /go/src/github.com/openshift/cluster-baremetal-operator
+COPY . .
+RUN make manager
 
-WORKDIR /workspace
-# Copy the Go Modules manifests
-COPY go.mod go.mod
-COPY go.sum go.sum
-# cache deps before building and copying source so that we don't need to re-download as much
-# and so that source changes don't invalidate our downloaded layer
-RUN go mod download
-
-# Copy the go source
-COPY main.go main.go
-COPY api/ api/
-COPY controllers/ controllers/
-
-# Build
-RUN CGO_ENABLED=0 GOOS=linux GOARCH=amd64 GO111MODULE=on go build -a -o manager main.go
-
-# Use distroless as minimal base image to package the manager binary
-# Refer to https://github.com/GoogleContainerTools/distroless for more details
-FROM gcr.io/distroless/static:nonroot
-WORKDIR /
-COPY --from=builder /workspace/manager .
-USER nonroot:nonroot
-
-ENTRYPOINT ["/manager"]
+FROM registry.svc.ci.openshift.org/ocp/4.6:base
+COPY --from=builder /go/src/github.com/openshift/cluster-baremetal-operator/bin/manager /usr/bin/cluster-baremetal-operator
+COPY --from=builder /go/src/github.com/openshift/cluster-baremetal-operator/manifests /manifests
+LABEL io.openshift.release.operator=true
+ENTRYPOINT ["/usr/bin/cluster-baremetal-operator"]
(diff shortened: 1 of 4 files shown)


Dockerfile
```diff
@@ -1,27 +1,12 @@
...
+# When bumping the Go version, don't forget to update the configuration of the
+# CI jobs in openshift/release.
+FROM registry.svc.ci.openshift.org/ocp/builder:rhel-8-golang-1.15-openshift-4.6 AS builder
+WORKDIR /go/src/github.com/openshift/cluster-baremetal-operator
+COPY . .
+RUN make manager
+
+FROM registry.svc.ci.openshift.org/ocp/4.6:base
+COPY --from=builder /go/src/github.com/openshift/cluster-baremetal-operator/bin/manager /usr/bin/cluster-baremetal-operator
+COPY --from=builder /go/src/github.com/openshift/cluster-baremetal-operator/manifests /manifests
```
kirankt, line 10: It would be preferable to use ...cluster-baremetal-operator/install for generated/run-leveled manifests just like MAO, MCO, do this today. I think the naming makes it clear that these manifests are needed to install the project.

hack/gen-crd.sh
```diff
@@ -0,0 +1,21 @@
...
+}
+
+# Generate manifests
+go run vendor/sigs.k8s.io/controller-tools/cmd/controller-gen/main.go "crd:trivialVersions=true" rbac:roleName=manager-role webhook paths=./... output:crd:artifacts:config=config/crd/bases
+sed -i '/^    controller-gen.kubebuilder.io\/version: (devel)/d' config/crd/bases/*
+
+# Copy crds into the manifests
+install_crd \
+  "config/crd/bases/metal3.io_provisionings.yaml" \
+  "manifests/0000_30_cluster-baremetal-operator_00_metal3provisioning.crd.yaml"
```
kirankt, line 20: Continuing from comment above. here as an example, I would put it in install/0000_30... Because ultimately all the installation manifests (service accounts, rbac, etc) end up here...

kirankt (changes requested):
Just a minor nit pick with directory naming but it looks good to me.

Dockerfile
andfasano: I don't have a strong preference here, I just checked how the other cluster operators organize the manifests, and most of them used `/manifests` - but I don't think there's a consolidated convention for that

kirankt:
/approve

No actual powers but I approve.

Dockerfile
```diff
@@ -1,27 +1,12 @@
...
+LABEL io.openshift.release.operator=true
```
sadasu, line 11: This line should be added in 4.7 when we actually want to be part of the release. I am not sure if this is the only part of the requirements and leaving here may not add us to the 4.7 release. Can we make this a comment with the instructions to un-comment when 4.7 starts?

Dockerfile
sadasu: I have seen both /install and /manifests being used. Let me check to see if it is just a matter of preference.

hack/gen-crd.sh
sadasu: Do we want 2 copies of metal3provisioning.crd.yaml in the repo? At line 15, we could directly add the generated CRD to the (install/manifests)/0000_30_cluster-baremetal-operator_00_metal3provisioning.crd.yaml.

andfasano: I think it depends on how much additional manipulation is required after the raw generation. In such case I noticed that there was an initial replace, that I kept as it was defined, and the renaming part. In general probably we could split into two separate actions (generate raw CRD _and_ copy them into the manifest folder)

Dockerfile
andfasano: I don't think that this is sufficient to be included as part of the official release, alternatively it would be probably easier to comment out the `promotion` section of the CI job I guess

hack/gen-crd.sh
kirankt: I am in favor of two copies as well (generated and copy/renamed one).

Dockerfile
sadasu: Yes, see https://mojo.redhat.com/docs/DOC-1179058#jive_content_id_Dont_be_Hasty_with_labels

Dockerfile
kirankt: I just brought the name up because MAO was using it, but I'm not really hung up about it. I'm OK with manifests or for that matter anything really.

hack/gen-crd.sh
sadasu: Lets keep the 2 copies. At this point, we don't have any modifications that we need to make to the generated CRD. So, OK to just copy it over.

PR #22, opened 2020-09-23
Fix the CRD added to /manifests

sadasu:


sadasu:
This file being out-of-sync might be a one-time thing. Let me know if you think otherwise.

asalkeld (commented):
/lgtm

dhellmann:
Can we set up a CI job to ensure they stay in sync? Maybe that should be part of the existing generate-check job?

PR #23, opened 2020-09-24
Add BMH CRD into manifests directory with appropriate run-level

kirankt:
This PR adds the BaremetalHost CRD from openshift/baremetal-operator into the manifests directory. The file is renamed with the correct runlevel nomenclature required by the CVO.

### DIFF
--- a/hack/gen-crd.sh
+++ b/hack/gen-crd.sh
@@ -1,14 +1,8 @@
 #!/bin/bash
 set -euo pipefail
 
-function install_crd {
-  local SRC="$1"
-  local DST="$2"
-  if ! diff -Naup "$SRC" "$DST"; then
-    cp "$SRC" "$DST"
-    echo "installed CRD: $SRC => $DST"
-  fi
-}
+REPO_ROOT=$(dirname "${BASH_SOURCE}")/..
+source $REPO_ROOT/hack/utils.sh
 
 # Generate manifests
 go run vendor/sigs.k8s.io/controller-tools/cmd/controller-gen/main.go "crd:trivialVersions=true" rbac:roleName=manager-role webhook paths=./... output:crd:artifacts:config=config/crd/bases
@@ -18,4 +12,3 @@ sed -i '/^    controller-gen.kubebuilder.io\/version: (devel)/d' config/crd/base
 install_crd \
   "config/crd/bases/metal3.io_provisionings.yaml" \
   "manifests/0000_30_cluster-baremetal-operator_00_metal3provisioning.crd.yaml"
-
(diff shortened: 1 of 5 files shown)


hack/gen-crd.sh
```diff
@@ -19,3 +19,10 @@ install_crd \
   "config/crd/bases/metal3.io_provisionings.yaml" \
   "manifests/0000_30_cluster-baremetal-operator_00_metal3provisioning.crd.yaml"
 
+# install bmh crd
+tmpfile="$(mktemp bmhcrd-XXXXXXXX)"
+curl -o $tmpfile https://raw.githubusercontent.com/openshift/baremetal-operator/master/deploy/crds/metal3.io_baremetalhosts_crd.yaml
```
andfasano, line 24: I would move the above two lines into a separate script to be invoked manually

kirankt: Curious why we want to keep it separate.

hack/gen-crd.sh
```diff
@@ -19,3 +19,10 @@ install_crd \
...
+if [ $? -eq 0 ]; then
+	install_crd $tmpfile "manifests/0000_30_cluster-baremetal-operator_08_baremetalhost.crd.yaml"
```
sadasu, line 26: This CRD can be installed right after the Provisioning CRD, so we can call it 0000_30_cluster-baremetal-operator_01_baremetalhost.crd.yaml

sadasu (changes requested):
Left one minor comment inline.

hack/gen-crd.sh
andfasano: Well, this script is part of the build system, and as a general approach I'd prefer to ensure the reliability of the build, given that the internet access (or even the target repo itself) could not be granted to be always  available when building (for example, from your disconnected laptop). Decoupling it will make the relationship loosely coupled, so with fewer chances of affecting the build process. The CRD also is not updated very frequently, so a manual triggering of the script could be enough in this case.

sadasu: @andfasano why do you think BMH CRD should be updated manually?

andfasano: Just to clarify, I meant that the update script (to copy the BMH CRD from the BMO repo into the `/manifests` folder) should be decoupled from the the `hack/gen-crd.sh`. Not necessarily invoked manually, there could be ie a periodic job on the CI for invoking it.

hack/gen-crd.sh
kirankt: Thanks. I've made the change in the latest commit.

hack/gen-crd.sh
kirankt: OK. Makes sense. Let me know what you guys think of the latest commit.

andfasano:
It looks fine for me

PR #24, opened 2020-09-28
improve logging practices

dhellmann:
Logging and then returning an error produces multiple messages, so
change a few spots to just return and let the caller do the logging.

Tighten up the language in a few log messages so they are shorter.

Wait until an action is done or failed to log, so we have "did action"
or "failed action" messages but no "doing action" messages. This
ensures that all of the information we need to have is in 1 log
message, instead of split across several.

### DIFF
--- a/controllers/clusteroperator.go
+++ b/controllers/clusteroperator.go
@@ -5,11 +5,12 @@ import (
 	"fmt"
 	"os"
 
-	"k8s.io/apimachinery/pkg/api/errors"
-	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"github.com/pkg/errors"
 
 	osconfigv1 "github.com/openshift/api/config/v1"
 	"github.com/openshift/library-go/pkg/config/clusteroperator/v1helpers"
+	k8serrors "k8s.io/apimachinery/pkg/api/errors"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 )
 
 // StatusReason is a MixedCaps string representing the reason for a
@@ -77,8 +78,10 @@ func (r *ProvisioningReconciler) createClusterOperator() (*osconfigv1.ClusterOpe
 
 	co, err := r.OSClient.ConfigV1().ClusterOperators().Create(context.Background(), defaultCO, metav1.CreateOptions{})
 	if err != nil {
-		return nil, err
+		return nil, errors.Wrap(err, fmt.Sprintf("failed to create ClusterOperator %s",
+			clusterOperatorName))
 	}
+	r.Log.V(1).Info("created ClusterOperator", "name", clusterOperatorName)
 
 	co.Status = defaultCO.Status
 	return r.OSClient.ConfigV1().ClusterOperators().UpdateStatus(context.Background(), co, metav1.UpdateOptions{})
@@ -87,8 +90,7 @@ func (r *ProvisioningReconciler) createClusterOperator() (*osconfigv1.ClusterOpe
 // getOrCreateClusterOperator gets the existing CO, failing which it creates a new CO.
 func (r *ProvisioningReconciler) getOrCreateClusterOperator() (*osconfigv1.ClusterOperator, error) {
 	existing, err := r.OSClient.ConfigV1().ClusterOperators().Get(context.Background(), clusterOperatorName, metav1.GetOptions{})
-	if errors.IsNotFound(err) {
-		r.Log.V(1).Info("cluster baremetal operator does not exist, creating a new one.")
+	if k8serrors.IsNotFound(err) {
 		return r.createClusterOperator()
 	}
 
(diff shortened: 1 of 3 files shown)


controllers/clusteroperator.go
```diff
@@ -5,11 +5,11 @@ import (
 	"fmt"
 	"os"
 
-	"k8s.io/apimachinery/pkg/api/errors"
-	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
-
 	osconfigv1 "github.com/openshift/api/config/v1"
 	"github.com/openshift/library-go/pkg/config/clusteroperator/v1helpers"
+	"github.com/pkg/errors"
```
kirankt, line 10: For included packages, can we stick to the following format? I believe we're already doing this in all other files.

stdlibs

outside packages

openshift packages

kirankt (changes requested):

controllers/clusteroperator.go
dhellmann: Good idea, I'll update

controllers/provisioning_controller.go
```diff
@@ -90,11 +88,9 @@ func (r *ProvisioningReconciler) readProvisioningCR(req ctrl.Request) (*metal3io
 	instance := &metal3iov1alpha1.Provisioning{}
 	if err := r.Client.Get(ctx, req.NamespacedName, instance); err != nil {
 		if apierrors.IsNotFound(err) {
-			r.Log.V(1).Info("Provisioning CR not found")
```
sadasu, line 93: We wanted to differentiate the case where the provisioning CR is not present vs the case where it is present but ill-formed so cannot be read.
Removing this error will not give the user enough information to diagnose a provisioning CR problem.

controllers/provisioning_controller.go
```diff
@@ -127,6 +123,7 @@ func (r *ProvisioningReconciler) Reconcile(req ctrl.Request) (ctrl.Result, error
 	if baremetalConfig == nil {
 		// Provisioning configuration not available at this time.
 		// Cannot proceed wtih metal3 deployment.
+		r.Log.V(1).Info("Provisioning CR not found")
```
sadasu, line 126: We could end up here if the Provisioning CR is not found or if it ill-formed and cannot be read. The log msg here only mentions that the provisioning CR is not found.

honza:
nit: the first line of the commit message should start with an imperative verb

controllers/clusteroperator.go
```diff
@@ -87,8 +90,7 @@ func (r *ProvisioningReconciler) createClusterOperator() (*osconfigv1.ClusterOpe
 // getOrCreateClusterOperator gets the existing CO, failing which it creates a new CO.
 func (r *ProvisioningReconciler) getOrCreateClusterOperator() (*osconfigv1.ClusterOperator, error) {
 	existing, err := r.OSClient.ConfigV1().ClusterOperators().Get(context.Background(), clusterOperatorName, metav1.GetOptions{})
-	if errors.IsNotFound(err) {
-		r.Log.V(1).Info("cluster baremetal operator does not exist, creating a new one.")
```
sadasu, line 91: This msg is useful to figure out if an existing CO was used or a new one was created. I agree with the general idea of not logging and the returning error. But, if the a log message indicates an intermediate state that then caused action1 or action2 to be taken, then I think it would be helpful to log it at that time before moving on.

controllers/provisioning_controller.go
sadasu, line 93: When the provisioning CR is not present, we donot return an error so removing this log message would mean we will not know from our logs if the Provisioning CR was present or not.

sadasu (commented):
Thanks for cleaning up the log and error messages. Just a minor nit inline.

controllers/clusteroperator.go
dhellmann: As things are now, we get 2 log messages when we create a CO. This removes the one that anticipates creating it, in favor of the success message that comes in createClusterOperator() on line 84.

dhellmann (commented):
> nit: the first line of the commit message should start with an imperative verb

Suggestions?

controllers/clusteroperator.go
sadasu: Yes, either one of these msgs should do. And I agree that the msg after creating one is better than the one anticipating creating of one.

sadasu:
> > nit: the first line of the commit message should start with an imperative verb
> 
> Suggestions?

@honza is there some documentation about this that we can add to the yet-to-be-created "CONTRIBUTING.md"?

honza:
I usually recommend [this article on the topic of git commit messages](https://chris.beams.io/posts/git-commit/).

Some suggestions for the subject line of  this particular patch:

* Make logging more consistent
* Improve logging
* Smooth out logging semantics

The body of the commit message can easily remain the same. 

I'm sorry if this is being too pedantic!

dhellmann:
I think I've addressed everyone's comments with the latest draft.

PR #25, opened 2020-09-29
Allow CVO to deploy CBO

andfasano:
This PR adds the manifests required by CVO for the deployment of CBO, as specified in the related [CVO documentation](https://github.com/openshift/cluster-version-operator/blob/master/docs/dev/operators.md#what-do-i-put-in-manifests)


manifests/0000_30_cluster-baremetal-operator_03_serviceaccount.yaml
```diff
@@ -0,0 +1,5 @@
+apiVersion: v1
+kind: ServiceAccount
+metadata:
+  name: cluster-baremetal-operator
+  namespace: cluster-baremetal-operator
```
kirankt, line 5: Are we sure we want to use this namespace? Why not using openshift-machine-api, I wonder? If we do want to use the cbo namespace, we should have a manifest to create it as well before we create the service account.

manifests/0000_30_cluster-baremetal-operator_04_deployment.yaml
```diff
@@ -0,0 +1,55 @@
...
+      k8s-app: cluster-baremetal-operator
+  template:
+    metadata:
+      labels:
+        k8s-app: cluster-baremetal-operator
+    spec:
+      priorityClassName: system-node-critical
+      serviceAccountName: cluster-baremetal-operator
+      containers:
+      - name: cluster-baremetal-operator
```
dhellmann, line 23: The container spec is going to need an image to be valid.

What would it take to get a quay.io image for the CBO, so we could use these files to test deploying it ourselves locally?

manifests/0000_30_cluster-baremetal-operator_03_serviceaccount.yaml
andfasano: Thanks, my mistake, I'll fix it!

manifests/0000_30_cluster-baremetal-operator_04_deployment.yaml
andfasano: That's a point I wanted to clarify (the image references), I will follow up on this topic

andfasano: Ok, I think that this manifest added by @sadasu will fix the problem: https://github.com/openshift/cluster-baremetal-operator/blob/f2f3f9e213becd309d973350052aaef39c434de9/manifests/image-references#L5

manifests/0000_30_cluster-baremetal-operator_05_deployment.yaml
```diff
@@ -0,0 +1,55 @@
+apiVersion: apps/v1
+kind: Deployment
+metadata:
+  name: cluster-baremetal-operator
+  namespace: openshift-machine-api
+  labels:
+    k8s-app: cluster-baremetal-operator
+  annotations:
+    exclude.release.openshift.io/internal-openshift-hosted: "true"
```
sadasu, line 9: I am not sure what the significance of this annotation is. Some of our manifests have it and others don't. Is there some documentation about this annotation?

sadasu (commented):
I have a generic question inline. Other than that this looks good.

manifests/0000_30_cluster-baremetal-operator_05_deployment.yaml
andfasano: Looks like it's the _cluster profile_ is used to exclude operators from being applied when the control plane is externally hosted (such as in the case of IBM Cloud, as described in [1]; see also [2]).
So I think we should keep it.

[1] https://github.com/openshift/enhancements/pull/200
[2] https://github.com/openshift/cluster-version-operator/pull/269

manifests/0000_31_cluster-baremetal-operator_06_clusteroperator.cr.yaml
```diff
@@ -0,0 +1,8 @@
+apiVersion: config.openshift.io/v1
```
dhellmann, line 1: This should probably come before the deployment, since the deployment uses it.

manifests/0000_31_cluster-baremetal-operator_05_deployment.yaml
```diff
@@ -0,0 +1,55 @@
...
+      labels:
+        k8s-app: cluster-baremetal-operator
+    spec:
+      priorityClassName: system-node-critical
+      serviceAccountName: cluster-baremetal-operator
+      containers:
+      - name: cluster-baremetal-operator
+        env:
+        - name: RELEASE_VERSION
+          value: "0.0.1-snapshot"
```
dhellmann, line 26: Does something in the build process change this value?

manifests/0000_31_cluster-baremetal-operator_06_clusteroperator.cr.yaml
kirankt: I'm not sure whether it matters. Shouldn't CBO create the CO deployement if not found anyways> FYI, MAO uses the same runlevel.

manifests/0000_31_cluster-baremetal-operator_05_deployment.yaml
kirankt: Not 100% sure. But this value is used in the CO to signal upgrades.

manifests/0000_31_cluster-baremetal-operator_06_clusteroperator.cr.yaml
dhellmann: Based on the note in [this section of the docs](https://github.com/openshift/cluster-version-operator/blob/master/docs/dev/clusteroperator.md#how-clusterversionoperator-handles-clusteroperator-in-release-image) I think the ClusterOperator resource has to come after the Deployment. The CVO doesn't use the manifest to actually create the resource, it only uses it to know which CO resources it should wait for before continuing with the next step of the install/upgrade. So, I think this is correct.

manifests/0000_31_cluster-baremetal-operator_05_deployment.yaml
dhellmann: [This section of the docs](https://github.com/openshift/cluster-version-operator/blob/master/docs/dev/clusteroperator.md#what-should-be-the-contents-of-clusteroperator-custom-resource-in-manifests) includes an example with a comment that the value "is substituted in the manifests when the payload is built" so I guess something about building the release image modifies the value.

PR #26, opened 2020-09-29
Bug 1906102: Add standard metrics support for CBO

sadasu:
According to https://github.com/openshift/cluster-version-operator/blob/master/docs/dev/operators.md#how-do-i-get-added-as-a-special-run-level, the certificate authority service is started at run level 09 and hence CBO's service monitor manifest should be created at that run level also.

For Prometheus to be able to scrape CBO metrics:
1. Added kubeProxyContainer to the CBO deployment
2. Created a ClusterIP Service (cluster-baremetal-operator-service) to expose CMO metrics port
3. Created a Service Monitor (cluster-baremetal-operator-servicemonitor) in order to be discovered automatically by the prometheus operator
4. Created ServiceAccount and Role with appropriate RBACs for prometheus to scrape CBO metrics endpoint

### DIFF
--- a/manifests/0000_31_cluster-baremetal-operator_05_kube-rbac-proxy-config.yaml
+++ b/manifests/0000_31_cluster-baremetal-operator_05_kube-rbac-proxy-config.yaml
@@ -1,0 +1,16 @@
+apiVersion: v1
+data:
+  config-file.yaml: |+
+    authorization:
+      resourceAttributes:
+        apiVersion: v1
+        resource: namespace
+        subresource: metrics
+        namespace: openshift-machine-api
+
+kind: ConfigMap
+metadata:
+  annotations:
+    include.release.openshift.io/self-managed-high-availability: "true"
+  name: kube-rbac-proxy
+  namespace: openshift-machine-api
--- a/manifests/0000_31_cluster-baremetal-operator_05_prometheus_rbac.yaml
+++ b/manifests/0000_31_cluster-baremetal-operator_05_prometheus_rbac.yaml
@@ -1,0 +1,41 @@
+apiVersion: rbac.authorization.k8s.io/v1
+kind: RoleBinding
+metadata:
+  annotations:
+    include.release.openshift.io/self-managed-high-availability: "true"
+  name: prometheus-k8s-cluster-baremetal-operator
+  namespace: openshift-machine-api
+roleRef:
+  apiGroup: rbac.authorization.k8s.io
+  kind: Role
+  name: prometheus-k8s-cluster-baremetal-operator
+subjects:
+- kind: ServiceAccount
+  name: prometheus-k8s
+  namespace: openshift-monitoring
+---
+apiVersion: rbac.authorization.k8s.io/v1
+kind: Role
+metadata:
+  annotations:
+    include.release.openshift.io/self-managed-high-availability: "true"
+  name: prometheus-k8s-cluster-baremetal-operator
+  namespace: openshift-machine-api
+rules:
+- apiGroups:
+  - ""
+  resources:
+  - namespace/metrics
+  verbs:
+  - get
+- apiGroups:
+  - ""
+  resources:
+  - services
+  - endpoints
+  - pods
+  verbs:
+  - get
+  - list
+  - watch
+---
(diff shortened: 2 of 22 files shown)


asalkeld:
looks good, but i don't know much about the service monitor..

dhellmann:
Reviewers can compare the work here to the MAO version in https://github.com/openshift/machine-api-operator/pull/671

sadasu:
> Reviewers can compare the work here to the MAO version in [openshift/machine-api-operator#671](https://github.com/openshift/machine-api-operator/pull/671)

Thanks @dhellmann ! In that case, this PR is not complete yet. I'll put it back in WIP then.

sadasu:
@dhellmann @asalkeld could you PTAL?

asalkeld:
@sadasu is this not the way to do it? https://github.com/openshift/cluster-baremetal-operator/blob/master/config/profiles/default/kustomization.yaml#L13-L14

sadasu:
> @sadasu is this not the way to do it? https://github.com/openshift/cluster-baremetal-operator/blob/master/config/profiles/default/kustomization.yaml#L13-L14
> 
> I've taken a quick stab at this here https://github.com/sadasu/cluster-baremetal-operator/compare/metrics...asalkeld:pull-26-but-moved-into-config?expand=1 (totally untested, I just moved your good work into config/). If you like this approach, please just copy my changes.

I knew there was a better way to do this using kustomize. :-) Thanks!

sadasu:
@asalkeld I see that with these changes the CBO deployment is split into 2 files namely config/cluster-baremetal-operator/cluster-baremetal-operator.yaml and config/profiles/default/manager_auth_proxy_patch.yaml. Is there a way to really patch manager_auth_proxy_patch.yaml into the final deployment manifest (manifests/0000_31_cluster-baremetal-operator_06_deployment.yaml) ?

sadasu:
@asalkeld I think that there are enough changes to convert to the style you proposed to warrant a separate PR. I have just started working on https://github.com/openshift/cluster-baremetal-operator/pull/103 to work through that. Could you please treat this PR as a first step towards that ?

config/cluster-baremetal-operator/cluster-baremetal-operator.yaml
```diff
@@ -20,6 +20,29 @@ spec:
       priorityClassName: system-node-critical
       serviceAccountName: cluster-baremetal-operator
       containers:
```
asalkeld, line 22: so personally i'd prefer these changes in the config/promethus directory

manifests/0000_31_cluster-baremetal-operator_05_kube-rbac-proxy-config.yaml
```diff
@@ -0,0 +1,16 @@
+apiVersion: v1
```
asalkeld, line 1: i think add this to config/promethus
the point of having the config/profiles/<> is that if we made one called
config/profiles/localtest for running on your laptop and we disabled prometus you could then do as below and it would work
```
kustomize build config/profiles/localtest | kubectl -f
```

manifests/0000_31_cluster-baremetal-operator_05_prometheus_rbac.yaml
```diff
@@ -0,0 +1,43 @@
+apiVersion: rbac.authorization.k8s.io/v1
+kind: RoleBinding
```
asalkeld, line 2: same with this

sadasu: Done. Do we still need the above change in controllers/provisioning_controller.go then?

manifests/0000_31_cluster-baremetal-operator_05_kube-rbac-proxy-config.yaml
sadasu: Ah! Thanks for that pointer. I made this change.

manifests/0000_31_cluster-baremetal-operator_05_prometheus_rbac.yaml
asalkeld: no, i think you can remove it.

manifests/0000_31_cluster-baremetal-operator_05_kube-rbac-proxy-config.yaml
sadasu: Done.

config/cluster-baremetal-operator/cluster-baremetal-operator.yaml
sadasu: Done.

PR #27, opened 2020-09-29
Add a manifest and ImageStream info for images needed by the CBO

sadasu:


andfasano (commented):
Probably it's better to have it just after the namespace, so I'd suggest to rename it to 0000_30_cluster-baremetal-operator_**01**_images.configmap.yaml. I'll reorder the other files in https://github.com/openshift/cluster-baremetal-operator/pull/25 accordingly

manifests/0000_30_cluster-baremetal-operator_01_images.configmap.yaml
```diff
@@ -0,0 +1,17 @@
+piVersion: v1
```
kirankt, line 1: This needs fixing. :smile:

manifests/image-references
```diff
@@ -0,0 +1,32 @@
+kind: ImageStream
+apiVersion: image.openshift.io/v1
```
kirankt, line 2: Not strictly needed but it would be nice (and consistent) to start all yamls files with:

```
---
apiVersion:
```

manifests/0000_30_cluster-baremetal-operator_01_images.configmap.yaml
```diff
@@ -0,0 +1,18 @@
+---
+apiVersion: v1
+kind: ConfigMap
+metadata:
+        name: cluster-baremetal-operator-images
```
kirankt, line 5: Sorry one last thing. Can we fix the formatting here.

sadasu: Not sure how I missed that!

PR #28, opened 2020-09-29
Ensure ClusterOperator status is set to available=true when platform is baremetal

kirankt:
If the platform is baremetal, the status of CO is changed to available=true, disabled=false.

### DIFF
--- a/controllers/clusteroperator.go
+++ b/controllers/clusteroperator.go
@@ -1,9 +1,25 @@
+/*
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
 package controllers
 
 import (
 	"context"
 	"fmt"
 	"os"
+	"strings"
 
 	"github.com/pkg/errors"
 
@@ -25,13 +41,13 @@ const (
 
 	// ReasonEmpty is an empty StatusReason
 	ReasonEmpty StatusReason = ""
-	// ReasonComplete is an empty StatusReason
+	// ReasonComplete signals status completion
 	ReasonComplete StatusReason = "DeployComplete"
-	// ReasonSyncing is an complete StatusReason
+	// ReasonSyncing indicates that we are currently syncing
 	ReasonSyncing StatusReason = "SyncingResources"
-	// ReasonSyncFailed is an failed StatusReason
+	// ReasonSyncFailed means we failed while syncing
 	ReasonSyncFailed StatusReason = "SyncingFailed"
-	// ReasonUnsupported is an unsupported StatusReason
+	// ReasonUnsupported means we have an unsupported platform
 	ReasonUnsupported StatusReason = "UnsupportedPlatform"
 )
 
@@ -114,11 +130,14 @@ func setStatusCondition(conditionType osconfigv1.ClusterStatusConditionType,
 	}
 }
 
-// getOperandVersions returns current version
-func getOperandVersions() []osconfigv1.OperandVersion {
+// getOperandVersions returns the operand version
+func (r *ProvisioningReconciler) getOperandVersions() []osconfigv1.OperandVersion {
 	operandVersions := []osconfigv1.OperandVersion{}
 	if releaseVersion := os.Getenv("RELEASE_VERSION"); len(releaseVersion) > 0 {
 		operandVersions = append(operandVersions, osconfigv1.OperandVersion{Name: "operator", Version: releaseVersion})
+	} else {
+		err := fmt.Errorf("env variable: RELEASE_VERSION was not set")
+		r.Log.Error(err, "failed to get OperandVersion")
 	}
 	return operandVersions
 }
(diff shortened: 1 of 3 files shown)


controllers/clusteroperator_test.go
```diff
@@ -266,3 +281,46 @@ func TestUpdateCOStatusDegraded(t *testing.T) {
...
 }
+
+func TestUpdateCOStatusAvailable(t *testing.T) {
+	tCases := []struct {
+		name               string
+		expectedConditions []osconfigv1.ClusterOperatorStatusCondition
+		expected           bool
+	}{
+		{
+			name: "Incorrect Condition",
```
sadasu, line 292: This test case seems a little artificial to me. In this case, it might be fine just to test if the condition is what it is supposed to be.
I felt the same about my test code for https://github.com/openshift/cluster-baremetal-operator/pull/29. So, I skipped creating multiple test cases.

kirankt: I don't think its artificial. The fact that the correct output is not expected is a valid test. If I change the expected result to true, then the test will fail.

controllers/clusteroperator.go
```diff
@@ -169,3 +203,25 @@ func (r *ProvisioningReconciler) updateCOStatusDegraded(degradedReason string, d
...
+		return err
+	}
+
+	// Write the operand versions when available
+	co.Status.Versions = getOperandVersions()
+
+	conds := []osconfigv1.ClusterOperatorStatusCondition{
+		setStatusCondition(osconfigv1.OperatorAvailable, osconfigv1.ConditionTrue, string(ReasonEmpty),
+			fmt.Sprintf("Cluster Baremetal Operator is available at %s", r.printOperandVersions())),
+		setStatusCondition(osconfigv1.OperatorProgressing, osconfigv1.ConditionFalse, "", ""),
```
dhellmann, line 220: I'm not sure using this function is really clearer than instantiating the resource inline here, but I don't feel so strongly that I'd block the PR over it.

controllers/clusteroperator_test.go
```diff
@@ -266,3 +281,46 @@ func TestUpdateCOStatusDegraded(t *testing.T) {
...
+		},
+	}
+
+	reconciler := newFakeProvisioningReconciler(setUpSchemeForReconciler(), &osconfigv1.Infrastructure{})
+	co, _ := reconciler.createClusterOperator()
+	reconciler.OSClient = fakeconfigclientset.NewSimpleClientset(co)
+
+	for _, tc := range tCases {
+		reconciler.updateCOStatusAvailable()
+		gotCO, _ := reconciler.OSClient.ConfigV1().ClusterOperators().Get(context.Background(), clusterOperatorName, metav1.GetOptions{})
```
dhellmann, line 315: We don't want to ignore unexpected errors in tests. They lead to weird failures, and it's better to have a fatal  log message with the error in that case.

controllers/clusteroperator_test.go
```diff
@@ -266,3 +281,46 @@ func TestUpdateCOStatusDegraded(t *testing.T) {
...
+
+		for _, expectedCondition := range tc.expectedConditions {
+			ok := v1helpers.IsStatusConditionPresentAndEqual(
+				gotCO.Status.Conditions, expectedCondition.Type, expectedCondition.Status,
+			)
+			if !ok {
+				assert.Regexp(t, tc.expected, ok)
```
dhellmann, line 322: This seems to be using a regex to compare 2 booleans?

controllers/provisioning_controller.go
```diff
@@ -119,6 +119,11 @@ func (r *ProvisioningReconciler) Reconcile(req ctrl.Request) (ctrl.Result, error
 		// We're disabled; don't requeue
 		return ctrl.Result{}, nil
 	}
+	// set ClusterOperator status to disabled=false, available=true
+	err = r.updateCOStatusAvailable()
+	if err != nil {
+		return ctrl.Result{}, err
```
dhellmann, line 125: Wrap the error?

controllers/clusteroperator.go
```diff
@@ -112,7 +128,7 @@ func setStatusCondition(conditionType osconfigv1.ClusterStatusConditionType,
 	}
 }
 
-// getOperandVersions returns current version
+// getOperandVersions returns the operand version
 func getOperandVersions() []osconfigv1.OperandVersion {
 	operandVersions := []osconfigv1.OperandVersion{}
 	if releaseVersion := os.Getenv("RELEASE_VERSION"); len(releaseVersion) > 0 {
```
dhellmann, line 134: Should we report an error if we can't figure out our version?

controllers/clusteroperator.go
```diff
@@ -121,6 +137,24 @@ func getOperandVersions() []osconfigv1.OperandVersion {
...
+	for _, operand := range getOperandVersions() {
+		versionsOutput = append(versionsOutput, fmt.Sprintf("%s: %s", operand.Name, operand.Version))
+	}
+	return strings.Join(versionsOutput, ", ")
+}
+
+// getCurrentVersions returns current version
+func (r *ProvisioningReconciler) getCurrentVersions() []osconfigv1.OperandVersion {
+	co, err := r.getOrCreateClusterOperator()
+	if err != nil {
```
dhellmann, line 152: I guess returning the defaults is expected, but should we at least log here?

controllers/clusteroperator.go
```diff
@@ -24,9 +40,9 @@ const (
 
 	// ReasonEmpty is an empty StatusReason
 	ReasonEmpty StatusReason = ""
-	// ReasonComplete is an empty StatusReason
+	// ReasonComplete is an complete StatusReason
```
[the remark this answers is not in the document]
kirankt: They were placeholders to shut up linters. But I'll try my best to add more descriptive and grammatically correct comments.

controllers/clusteroperator.go
```diff
@@ -121,6 +137,24 @@ func getOperandVersions() []osconfigv1.OperandVersion {
 	return operandVersions
 }
 
+// printOperandVersion prints current version as a string
+func (r *ProvisioningReconciler) printOperandVersions() string {
```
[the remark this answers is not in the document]
kirankt: We do use it in: https://github.com/kirankt/cluster-baremetal-operator/blob/c656c97ec758afc5a6f2fa0c00cc0918ff1cee95/controllers/clusteroperator.go#L219

controllers/clusteroperator.go
kirankt: Perhaps a little bit more readable but I understand your point.

controllers/clusteroperator_test.go
kirankt: Yes. Will fix it.

controllers/clusteroperator.go
kirankt: +1. I think this is env value required and will be passed to the CBO deployment. Nevertheless, we should do something about it if it goes missing.

controllers/clusteroperator.go
kirankt: +1. Fixed it. Thanks.

controllers/provisioning_controller.go
kirankt: Thanks. WIll do.

controllers/clusteroperator_test.go
kirankt: +1. This is addressed in the latest commit.

controllers/clusteroperator.go
sadasu: I don't see getCurrentVersions() being called from anywhere.

kirankt: Yep. Not needed in the PR. Going to remove it.

sadasu:
Thanks for taking care of all my comments.

controllers/clusteroperator.go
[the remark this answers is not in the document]
asalkeld: I think what Doug is saying is to remove the "print" from function name if it is not printing.
it's more of a operandVersionsString()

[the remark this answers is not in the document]
kirankt: A heck. I removed the function and just inlined the code. Easier than coming up with a new name. :-)

controllers/clusteroperator_test.go
```diff
@@ -266,3 +289,53 @@ func TestUpdateCOStatusDegraded(t *testing.T) {
 		}
 	}
 }
+
+func TestUpdateCOStatusAvailable(t *testing.T) {
```
sadasu, line 293: We are adding code to get the CO's name and version. Could we make sure they are set correctly in this test?

kirankt: Looks like this code will change significantly (and better). Thinking of closing this PR.

kirankt:
Going to close this PR in favor of a more comprehensive implementation for CO: https://github.com/openshift/cluster-baremetal-operator/pull/32

PR #29, opened 2020-09-30
Add secrets required to create metal3 Deployment

sadasu:


controllers/provisioning_controller.go
```diff
@@ -141,6 +143,11 @@ func (r *ProvisioningReconciler) Reconcile(req ctrl.Request) (ctrl.Result, error
 		// Temporarily not requeuing request
 		return ctrl.Result{}, nil
 	}
+	//Create Secrets needed for Metal3 deployment
+	if err := provisioning.CreateMetal3PasswordSecrets(r.kubeClient.CoreV1(), ComponentNamespace); err != nil {
+		r.Log.V(1).Info("Unable to create one or more Secrets")
+		return ctrl.Result{}, err
```
andfasano, line 149: As per recent discussion on logging style, I think this should be:
`return ctrl.Result{}, errors.Wrap(err, "unable to create one or more Secrets")}`

provisioning/baremetal_secrets_test.go
```diff
@@ -0,0 +1,87 @@
...
+const testNamespace = "test-namespce"
+
+func TestGenerateRandomPassword(t *testing.T) {
+	pwd, err := generateRandomPassword()
+	if err != nil {
+		t.Errorf("Unexpected error: %s", err)
+	}
+	if pwd == "" {
+		t.Errorf("Expected a valid string but got null")
+	}
```
andfasano, line 21: I think that an interesting test for a random password generator method would be to verify that two consecutive calls produce different passwords

provisioning/baremetal_secrets_test.go
```diff
@@ -0,0 +1,87 @@
...
+}
+
+//Testing the case where the password already exists
+func TestCreateMariadbPasswordSecret(t *testing.T) {
```
andfasano, line 25: Personally I am more inclined to simpler tests. This one also seems to combine together different cases, so I'd recommend splitting them also to increase readability.

provisioning/baremetal_secrets_test.go
```diff
@@ -0,0 +1,87 @@
...
+		t.Fatal("Failure reading second Mariadb password from Secret.")
+	}
+	if oldPassword != newPassword {
+		t.Fatalf("Both passwords do not match.")
+	} else {
+		t.Logf("First Mariadb password is being preserved over re-creation as expected.")
+	}
+}
+
+func TestCreateMetal3PasswordSecrets(t *testing.T) {
```
andfasano, line 62: The same for this test

provisioning/baremetal_secrets.go
```diff
@@ -0,0 +1,131 @@
...
+	password, err := generateRandomPassword()
+	if err != nil {
+		return err
+	}
+	hash, err := bcrypt.GenerateFromPassword([]byte(password), 5) // Use same cost as htpasswd default
+	if err != nil {
+		return err
+	}
+	// Change hash version from $2a$ to $2y$, as generated by htpasswd.
+	// These are equivalent for our purposes.
```
dhellmann, line 91: A link to some docs would help here. It feels iffy to be modifying the hash produced by bcrypt ourselves, is that normal? There's not a flag we can pass to bcrypt to produce the value we need?

controllers/provisioning_controller.go
sadasu: Interesting because I did try to follow what https://github.com/openshift/cluster-baremetal-operator/pull/24 is doing.

Do I really need to wrap this error again? What we are getting from CreateMetal3PasswordSecrets() is a specific error msg about which password creation failed with the error from the underlying call.

provisioning/baremetal_secrets.go
```diff
@@ -0,0 +1,131 @@
...
+			},
+		},
+		metav1.CreateOptions{},
+	)
+	log.V(1).Info("created password for", "name", name)
+	return err
+}
+
+// CreateMetal3PasswordSecrets creates all Secrets that a metal3 pod would need
+func CreateMetal3PasswordSecrets(client coreclientv1.SecretsGetter, targetNamespace string) error {
```
dhellmann, line 120: Making this function the public API for the package is going to make it more difficult to change the behavior later if we change the metal3 deployment and don't need some of the secrets. I think I would just expose each of the other functions and eliminate this one. That will also make it easier to implement the reconcile logic for re-creating a secret if one is removed.

Should we also separate the step of creating the `Secret` in memory and the step of passing that to the client, since for unit tests we may want to verify that a `Secret` has particular members?

dhellmann (changes requested):
In addition to the inline comments, I think we want to get into the habit of separating CBO logic and vendor updates into different commits. That will make it easier to review large PRs like this one, since in the UI you can look at just the commit with CBO logic in it and leave comments there.

provisioning/baremetal_secrets_test.go
sadasu: What this test is trying to test here is that fact that when mariadbPassword already exists, another call to this method would not create a new password. To test this behavior we need to make 2 calls to createMariadbPasswordSecret(). We are testing only one scenario here. 
I am open to splitting or making it simpler but not sure how. I can add comments within the code to explain what the test is trying to accomplish.

provisioning/baremetal_secrets_test.go
sadasu: Not sure how. We want to test that the top level function CreateMetal3PasswordSecrets() generates 3 passwords.

go.mod
```diff
@@ -11,6 +11,8 @@ require (
 	github.com/openshift/library-go v0.0.0-20200910214143-887092e305c1
```
kirankt, line 11: This file should end up in the vendoring commit. :frowning_face:

provisioning/baremetal_secrets.go
```diff
@@ -0,0 +1,131 @@
...
+auth_type = http_basic
+username = %s
+password = %s
+`,
+					configSection, username, password),
+			},
+		},
+		metav1.CreateOptions{},
+	)
+	log.V(1).Info("created password for", "name", name)
```
kirankt, line 115: Can we combine the first two strings into one? And also shouldn't we be using Infof?

kirankt (changes requested):
Couple of nits mentioned inline. Let me know what you think. Thanks.

controllers/provisioning_controller.go
dhellmann: In a concurrent system like this, we want each event to have 1 log message and for that message to include all of the information someone would need to understand what's going on. So we almost never want to log and then return an error that the caller will also log, because then we end up with 2 log messages on separate lines. The log aggregation tool won't know they are related, so information will be harder to find.

The message in the log line here, "Unable to create one or more secrets," could be made more helpful by adding context about the secret(s) that couldn't be created -- names, purpose, etc. Splitting that function into multiple calls will make it easier to provide that context, because the message can say "could not create foo secret".

provisioning/baremetal_secrets.go
sadasu: This a reproduction of code added by @zaneb as part of https://github.com/openshift/machine-api-operator/pull/657.

controllers/provisioning_controller.go
sadasu: The error message returned by CreateMetal3PasswordSecrets() tells exactly which secret had the error and that would definitely be part of the err returned here. 
I'll remove the log message to avoid logging and then returning an error.

provisioning/baremetal_secrets.go
sadasu: Couple of issues listed in the comment above -
1. Support creating a subset of the secrets due to changes to CBO
2. Create only the secret that is not present during the reconcile loop
Suggestion is that the solution to both these problems is that we should get rid of the top level function and be able to call individual methods separately. 

Having the top level functions abstracts away the many constants in play while generating the password. This should not be within the reconcile logic. Also, these functions already check if the Secret is already there and will create a new one **only** if not present. So, that is a non-issue.

Last part of code is asking for code re-organization to help with unit testing. The unit tests can be augmented to check for more fields within the Secret without creating the Secret in memory 1st.

go.mod
sadasu: Done.

provisioning/baremetal_secrets.go
sadasu: Hopefully, addressed this in the latest changes.

provisioning/baremetal_secrets_test.go
sadasu: Added test case to try 2 consecutive calls. Ideally, the code to generate random numbers of a reasonable length should not repeat for n tries where n is a non-trivial number.

provisioning/baremetal_secrets_test.go
sadasu: Updated comment for the test to explain in detail what the test does.

provisioning/baremetal_secrets_test.go
sadasu: Updated these tests after comment to remove the top level function to create all 3 passwords together.

provisioning/baremetal_secrets.go
sadasu: Referring to https://github.com/openshift/cluster-baremetal-operator/blob/master/controllers/provisioning_controller.go#L71 and https://github.com/go-logr/logr/blob/master/examples/usage_example.go, this seems like it is OK.

sadasu:
@dhellmann, am I missing something? https://github.com/openshift/cluster-baremetal-operator/pull/29#pullrequestreview-500483341 suggests I should move vendor updates to a diff commit. That is how the current PR is organized, isn't it?

provisioning/baremetal_secrets.go
sadasu: We should be restarting a Pod(s) when the secrets change. 
Also, there is currently no need for a method to delete a secret that was created. But, the current way the code is organized makes it easy to add those methods when we need them.
At this point we do not need to pass provisioning config to these methods because currently all secrets are needed for all ProvisioningNetwork modes. When that changes, we can always come back and update these methods.

provisioning/baremetal_secrets_test.go
```diff
@@ -0,0 +1,87 @@
...
+	oldPassword, ok := oldMaridbPassword.StringData[baremetalSecretKey]
+	if !ok || oldPassword == "" {
+		t.Fatal("Failure reading first Mariadb password from Secret.")
+	}
+
+	// The pasword definitely exists. Try creating again.
+	if err := createMariadbPasswordSecret(kubeClient.CoreV1(), testNamespace); err != nil {
+		t.Fatal("Failure creating second Mariadb password.")
+	}
+	newMaridbPassword, err := client.Secrets(testNamespace).Get(context.Background(), baremetalSecretName, metav1.GetOptions{})
```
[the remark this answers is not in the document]
sadasu: Done.

provisioning/baremetal_secrets_test.go
```diff
@@ -0,0 +1,87 @@
...
+func TestCreateMariadbPasswordSecret(t *testing.T) {
+	kubeClient := fakekube.NewSimpleClientset(nil...)
+	client := kubeClient.CoreV1()
+
+	// First create a mariadb password secret
+	if err := createMariadbPasswordSecret(kubeClient.CoreV1(), testNamespace); err != nil {
+		t.Fatalf("Failed to create first Mariadb password. %s ", err)
+	}
+	// Read and get Mariadb password from Secret just created.
+	oldMaridbPassword, err := client.Secrets(testNamespace).Get(context.Background(), baremetalSecretName, metav1.GetOptions{})
```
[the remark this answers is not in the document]
sadasu: Done.

provisioning/baremetal_secrets.go
dhellmann: Let's put all of that in a comment here.

provisioning/baremetal_secrets_test.go
andfasano: Let me try put down an example

andfasano: Here's the sample PR: https://github.com/sadasu/cluster-baremetal-operator/pull/2

dhellmann:
> @dhellmann, am I missing something? [#29 (review)](https://github.com/openshift/cluster-baremetal-operator/pull/29#pullrequestreview-500483341) suggests I should move vendor updates to a diff commit. That is how the current PR is organized, isn't it?

Ugh, sorry. Github's UI is terrible. I looked at ab6a3f8f33b6128fc90a4cc4168c5ff30d9a5d75 and saw "changes 206 files", but that's the summary of the PR not the commit.

provisioning/baremetal_secrets.go
```diff
@@ -0,0 +1,126 @@
...
+	return createIronicSecret(client, targetNamespace, ironicSecretName, ironicUsername, "ironic")
+}
+
+// CreateInspectorPasswordSecret creates a Secret for the Ironic Inspector Password
+func CreateInspectorPasswordSecret(client coreclientv1.SecretsGetter, targetNamespace string) error {
+	return createIronicSecret(client, targetNamespace, inspectorSecretName, inspectorUsername, "inspector")
+}
+
+func createIronicSecret(client coreclientv1.SecretsGetter, targetNamespace string, name string, username string, configSection string) error {
+	_, err := client.Secrets(targetNamespace).Get(context.Background(), name, metav1.GetOptions{})
```
dhellmann, line 86: Should we be using the pattern that the MAO uses, of having utility functions to create resources and then generic functions to ensure those resources exist?

https://github.com/openshift/machine-api-operator/blob/master/pkg/operator/sync.go#L320

dhellmann: Maybe that wouldn't apply in the case of the secrets, since we don't want to change them every time?

sadasu: Maybe we could. I don't understand everything that is happening in  https://github.com/openshift/library-go/blob/master/pkg/operator/resource/resourceapply/core.go#L237 tbh.

provisioning/baremetal_secrets_test.go
sadasu: Thanks for the test you provided. I think this is a good addition to the existsing test so I have added it and kept the existing test that verifies functionality when we call the CreateMariadbPasswordSecret() multiple times when it already exists.

andfasano: Thanks Sandhya but honestly I don't see the point in keeping both the approaches (from a readability/maintenance point of viewe). In general, if there's a case missing it should be added in the list.

provisioning/baremetal_secrets_test.go
```diff
@@ -0,0 +1,151 @@
...
+	kubeClient := fakekube.NewSimpleClientset(nil...)
+	client := kubeClient.CoreV1()
+
+	err := CreateIronicPasswordSecret(client, testNamespace)
+	if err != nil {
+		t.Errorf("unexpected error: %v", err)
+		return
+	}
+	// Check if Ironic secret exits
+	secret, err := client.Secrets(testNamespace).Get(context.Background(), ironicSecretName, metav1.GetOptions{})
```
andfasano, line 127: I think it would be better to use ` kubeClient.Tracker().Get` in test code to verify that the mock registered correctly the new stored secret

provisioning/baremetal_secrets_test.go
sadasu: Yes, I was able to add it as a case to this test.

provisioning/baremetal_secrets.go
sadasu: Couple of things to consider:
1. Do we want to make all 3 Secrets immutable?
https://github.com/kubernetes/api/blob/master/core/v1/types.go#L5624
(I don't understand the scenario where we would want to just change a secret and want metal3 or any other to be re-deployed)
2. I could use https://github.com/openshift/library-go/blob/master/pkg/operator/resource/resourceapply/core.go#L237 to actually create the Secrets. The pattern pointed to earlier (https://github.com/openshift/cluster-baremetal-operator/pull/29#discussion_r498977454) is not possible exactly because the concept of generations doesn't seem to exist for Secret types.

dhellmann: How will making the secrets immutable affect upgrades from clusters that do not have immutable secrets?

Using `ApplySecret()` looks like it's probably a good idea.

dhellmann: Maybe we can create a separate ticket to track the idea of making those secrets immutable, to give us time to do the research without blocking this work?

zaneb: I don't see any reason not to make them immutable (provided that feature gate is actually enabled in OpenShift). It won't affect existing secrets but at least new ones will be. Note that the Secret can still be deleted, at which point we'll create a new one, so it doesn't change a lot.

`ApplySecret()` seems to be about updating secrets, so not what we want.

sadasu: Yes, ApplySecrets() does not work for us. (Its name is misleading because as @zaneb mentioned, it is about updating secrets.)
This is why:
1. We 1st have to create a Secret type
2. Call ApplySecret() with this Secret
3. ApplySecret would get an existing Secret with the same name as the one passed in. If found, it updates it with the new value.
4. If not found, it applies the Secret passed in.

Step 3 is problematic for the way we want our secrets to be manged. So, sticking with the current implementation.

provisioning/baremetal_secrets_test.go
sadasu: Done.

asalkeld:
looks good to me

sadasu:
Updated Makefile to ignore gosec rules that look for hardcoded credentials.
https://fossies.org/linux/buildah/tests/tools/vendor/github.com/golangci/gosec/README.md

kirankt:
/lgtm

Since e2e-metal-ipi is not really essential at this moment.

PR #30, opened 2020-10-05
Add bmh as a relatedobject in the clusteroperator

kirankt:
This PR adds bmh as a relatedobject in the clusteroperator. This will enable the user to run:  `oc get -o yaml clusteroperator baremetal` and view that "baremetalhosts" is in the list of `relatedObjects`.

### DIFF
--- a/controllers/clusteroperator.go
+++ b/controllers/clusteroperator.go
@@ -56,6 +56,12 @@ func relatedObjects() []osconfigv1.ObjectReference {
 			Resource: "namespaces",
 			Name:     ComponentNamespace,
 		},
+		{
+			Group:     "metal3.io",
+			Resource:  "baremetalhosts",
+			Name:      "baremetalhosts.metal3.io",
+			Namespace: ComponentNamespace,
+		},
 	}
 }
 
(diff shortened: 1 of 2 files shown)


andfasano:
Note: this will also help to define an openshift conformance test fro CBO here: https://github.com/openshift/openshift-tests/blob/292dfd1dc2d170bd8b5f2d4dfb2414ef657ff22b/test/extended/operators/clusteroperators.go

controllers/clusteroperator.go
```diff
@@ -56,6 +56,12 @@ func relatedObjects() []osconfigv1.ObjectReference {
 			Resource: "namespaces",
 			Name:     ComponentNamespace,
 		},
+		{
+			Group:     "metal3.io",
+			Resource:  "baremetalhosts",
+			Name:      "baremetalhosts.metal3.io",
```
zaneb, line 62: This doesn't look right to me. There's unlikely to be a BareMetalHost named `baremetalhosts.metal3.io`, and even if there were there's no reason it should be a related object?
I presume the intention here was to reference the baremetalhost **CRD**, in which case I think the Group should be `apiextensions.k8s.io` and the Resource should be `CustomResourceDefinition`.

zaneb: @kirankt pointed out that the [equivalent code in the MAO](https://github.com/openshift/machine-api-operator/blob/71aeab4ff840991c8569baf3817e8640e03a2dfe/pkg/operator/status.go#L202-L207) just has the name as an empty string. Presumably that means that _all_ `baremetalhosts` would be treated as related, and that was the intention here (i.e. not to mark the CRD as related). I think we should drop the `Name`.

PR #31, opened 2020-10-06
Use golangci lint for linting

asalkeld:


controllers/clusteroperator.go
```diff
@@ -114,15 +113,6 @@ func setStatusCondition(conditionType osconfigv1.ClusterStatusConditionType,
...
 
-// getOperandVersions returns current version
-func getOperandVersions() []osconfigv1.OperandVersion {
-	operandVersions := []osconfigv1.OperandVersion{}
-	if releaseVersion := os.Getenv("RELEASE_VERSION"); len(releaseVersion) > 0 {
-		operandVersions = append(operandVersions, osconfigv1.OperandVersion{Name: "operator", Version: releaseVersion})
-	}
-	return operandVersions
-}
-
```
asalkeld, line 125: unused function

controllers/clusteroperator_test.go
```diff
@@ -5,15 +5,15 @@ import (
 	"testing"
 	"time"
 
+	osconfigv1 "github.com/openshift/api/config/v1"
```
kirankt, line 8: Why was this moved? Usually we're keeping the imports in the following format:

std libs

outside libs

openshift libs

controllers/clusteroperator.go
kirankt: Is the linter complaining about this?

controllers/clusteroperator_test.go
```diff
@@ -45,7 +45,10 @@ func TestUpdateCOStatusDisabled(t *testing.T) {
 	reconciler.OSClient = fakeconfigclientset.NewSimpleClientset(co)
 
 	for _, tc := range tCases {
-		reconciler.updateCOStatusDisabled()
+		err := reconciler.updateCOStatusDisabled()
```
kirankt, line 48: Is this linting related? If not, we should not worry about it in this PR.

controllers/provisioning_controller.go
```diff
@@ -20,15 +20,15 @@ import (
 	"fmt"
 
 	"github.com/go-logr/logr"
+	osconfigv1 "github.com/openshift/api/config/v1"
```
kirankt, line 23: Same here. Need to move the openshift libs down.

controllers/provisioning_controller_test.go
```diff
@@ -3,24 +3,24 @@ package controllers
 import (
 	"testing"
 
+	configv1 "github.com/openshift/api/config/v1"
```
kirankt, line 6: Ditto.

main.go
```diff
@@ -19,6 +19,8 @@ import (
 	"flag"
 	"os"
 
+	osconfigv1 "github.com/openshift/api/config/v1"
```
kirankt, line 22: Another one.

kirankt (changes requested):
Few minor nits and also a question about changes not specifically related to this linting change. Please let me know what you think.

controllers/clusteroperator.go
asalkeld: yes, don't worry tho' when we need it, it will be in git history..

controllers/clusteroperator_test.go
asalkeld: yes, it forces you to check errors

controllers/clusteroperator_test.go
asalkeld: aaah, OK so goimports does
```
std libs

outside libs

local libs
```
I can adjust the definition of "local" to github.com/openshift

asalkeld: aaah, OK so goimports does
```
std libs

outside libs

local libs
```
I can adjust the definition of "local" to github.com/openshift

kirankt (approved):
Thanks for addressing changes. Looks fine to me. I'll let either @dhellmann, @sadasu or @andfasano  take a look and lgtm this PR.

PR #32, opened 2020-10-06
Add correct condition states

asalkeld:
This PR aims to implement the following logic.
```
0. When the cluster platform is NOT "baremetal", the ClusterOperator has status Progressing=false, Available=true, Disabled=true, Upgradeable=true, Degraded=false

When the cluster platform is "baremetal" and...
1. the Provisioning CR is valid and metal3 is being deployed, the ClusterOperator has status Progressing=true, Available=false, Disabled=false, Upgradeable=true, Degraded=false
2. the Provisioning CR is valid and metal3 is running, the ClusterOperator has status Progressing=false, Available=true, Disabled=false, Upgradeable=true, Degraded=false
3. the Provisioning CR is valid and metal3 has not successfully launched after 20 minutes, the ClusterOperator has status Progressing=true, Available=false, Disabled=false, Upgradeable=true, Degraded=true
4. the Provisioning CR is valid and metal3 has not successfully launched and is in a crash loop, the ClusterOperator has status Progressing=false, Available=false, Disabled=false, Upgradeable=true, Degraded=true
```
the above reasons for state changes are made to match the condition reason as well.
One function updateCOStatus() is used to implement the status changes.

### DIFF
--- a/controllers/clusteroperator.go
+++ b/controllers/clusteroperator.go
@@ -24,12 +24,22 @@ const (
 
 	// ReasonEmpty is an empty StatusReason
 	ReasonEmpty StatusReason = ""
-	// ReasonComplete is an empty StatusReason
+
+	// ReasonComplete the deployment of required resources is complete
 	ReasonComplete StatusReason = "DeployComplete"
-	// ReasonSyncing is an complete StatusReason
+
+	// ReasonSyncing means that resources are deploying
 	ReasonSyncing StatusReason = "SyncingResources"
-	// ReasonSyncFailed is an failed StatusReason
-	ReasonSyncFailed StatusReason = "SyncingFailed"
+
+	// ReasonInvalidConfiguration indicates that the configuration is invalid
+	ReasonInvalidConfiguration StatusReason = "InvalidConfiguration"
+
+	// ReasonDeployTimedOut indicates that the deployment timedOut
+	ReasonDeployTimedOut StatusReason = "DeployTimedOut"
+
+	// ReasonDeplymentCrashLooping indicates that the deployment is crashlooping
+	ReasonDeplymentCrashLooping StatusReason = "DeploymentCrashLooping"
+
 	// ReasonUnsupported is an unsupported StatusReason
 	ReasonUnsupported StatusReason = "UnsupportedPlatform"
 )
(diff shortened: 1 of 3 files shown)


controllers/clusteroperator_test.go
```diff
@@ -2,41 +2,36 @@ package controllers
...
-			},
-			expected: false,
-		},
 		{
 			name: "Correct Condition",
 			expectedConditions: []osconfigv1.ClusterOperatorStatusCondition{
+				setStatusCondition(osconfigv1.OperatorDegraded, osconfigv1.ConditionFalse, "", ""),
 				setStatusCondition(osconfigv1.OperatorAvailable, osconfigv1.ConditionTrue, "", ""),
-				setStatusCondition(OperatorDisabled, osconfigv1.ConditionTrue, "", ""),
+				setStatusCondition(OperatorDisabled, osconfigv1.ConditionTrue, "UnsupportedPlatform", "Operator is non functional"),
```
dhellmann, line 31: If this "disabled" behavior is meant to be part of the API of an operator, should we add it to the same package where the other condition constants are defined?

controllers/clusteroperator_test.go
```diff
@@ -192,6 +186,40 @@ func normalizeTransitionTimes(got, expected osconfigv1.ClusterOperatorStatus) {
 	}
 }
 
+// getStatusDiff this is the same as v1helpers.GetStatusDiff except for:
```
dhellmann, line 189: We don't need the behaviors that are removed in this version?

controllers/provisioning_controller.go
```diff
@@ -107,7 +107,7 @@ func (r *ProvisioningReconciler) Reconcile(req ctrl.Request) (ctrl.Result, error
 	}
 	if !enabled {
 		// set ClusterOperator status to disabled=true, available=true
-		err = r.updateCOStatusDisabled()
+		err = r.updateCOStatus(ReasonUnsupported, "Operator is non functional", "")
```
dhellmann, line 110:
```suggestion
		err = r.updateCOStatus(ReasonUnsupported, "Operator is non-functional", "")
```

dhellmann (commented):
> should mostly implement KNIDEPLOY-2300 when the calls to updateCOStatus()
> are in place.

Our team jira isn't public, so we usually try to describe the tickets here on GitHub instead of just including a reference.

controllers/clusteroperator_test.go
sadasu: The other states are defined here; https://github.com/openshift/api/blob/master/config/v1/types_cluster_operator.go#L141. There was a discussion earlier with @deads2k to add support for the "Disabled" state there and the answer then was a No. At this time, there are only a very small number of operators that use this state, that might have something to do with that decision.

controllers/clusteroperator.go
```diff
@@ -133,40 +143,32 @@ func (r *ProvisioningReconciler) syncStatus(co *osconfigv1.ClusterOperator, cond
...
+		v1helpers.SetStatusCondition(&conds, setStatusCondition(osconfigv1.OperatorProgressing, osconfigv1.ConditionFalse, string(newReason), progressMsg))
+	case ReasonInvalidConfiguration, ReasonDeployTimedout:
+		v1helpers.SetStatusCondition(&conds, setStatusCondition(osconfigv1.OperatorDegraded, osconfigv1.ConditionTrue, string(newReason), msg))
+		v1helpers.SetStatusCondition(&conds, setStatusCondition(osconfigv1.OperatorAvailable, osconfigv1.ConditionTrue, string(ReasonEmpty), ""))
+		v1helpers.SetStatusCondition(&conds, setStatusCondition(osconfigv1.OperatorProgressing, osconfigv1.ConditionTrue, string(newReason), progressMsg))
+	case ReasonDeplymentCrashLooping:
+		v1helpers.SetStatusCondition(&conds, setStatusCondition(osconfigv1.OperatorDegraded, osconfigv1.ConditionTrue, string(newReason), msg))
+		v1helpers.SetStatusCondition(&conds, setStatusCondition(osconfigv1.OperatorAvailable, osconfigv1.ConditionFalse, string(newReason), msg))
+		v1helpers.SetStatusCondition(&conds, setStatusCondition(osconfigv1.OperatorProgressing, osconfigv1.ConditionFalse, string(newReason), progressMsg))
 	}
```
sadasu, line 172: Should the getOperandVersions() be implemented in some form (see https://github.com/openshift/cluster-baremetal-operator/pull/28/files#diff-e86bbec1cec0825c6c6f7f39bad704abR134) ?

controllers/clusteroperator_test.go
asalkeld: we could add back relatedobjects and version when we add unit tests for them, but i am not using them .

controllers/clusteroperator.go
asalkeld: @sadasu I am happy to add it. Can I do it in a separate PR?

controllers/clusteroperator_test.go
asalkeld: I have updated the name and comment to better indicate what it does

controllers/clusteroperator.go
asalkeld: https://github.com/openshift/cluster-baremetal-operator/pull/36

controllers/clusteroperator.go
```diff
@@ -24,12 +24,22 @@ const (
...
+	// ReasonSyncing means that resources are deploying
 	ReasonSyncing StatusReason = "SyncingResources"
-	// ReasonSyncFailed is an failed StatusReason
-	ReasonSyncFailed StatusReason = "SyncingFailed"
+
+	// ReasonInvalidConfiguration indicates that the configuration is invalid
+	ReasonInvalidConfiguration StatusReason = "InvalidConfiguration"
+
+	// ReasonDeployTimedout indicates that the deployment timeedout
+	ReasonDeployTimedout StatusReason = "DeployTimedout"
```
dtantsur, line 38: nit: "out" is a separate word, so `TimedOut`?

sadasu (commented):
Looks good. Ready to merge after https://github.com/openshift/cluster-baremetal-operator/pull/32#discussion_r501683064 is taken care of.

sadasu (commented):
/lgtm

Adding it back after the rebase.

controllers/clusteroperator.go
```diff
@@ -144,40 +154,32 @@ func (r *ProvisioningReconciler) syncStatus(co *osconfigv1.ClusterOperator, cond
...
-	co, err := r.getOrCreateClusterOperator()
-	if err != nil {
-		return err
-	}
-
-	conds := []osconfigv1.ClusterOperatorStatusCondition{
-		setStatusCondition(osconfigv1.OperatorDegraded, osconfigv1.ConditionTrue, degradedReason, degradedMessage),
-		setStatusCondition(osconfigv1.OperatorProgressing, osconfigv1.ConditionTrue, detailedError, progressingMessage),
-		setStatusCondition(osconfigv1.OperatorAvailable, osconfigv1.ConditionFalse, "", ""),
+	conds := defaultStatusConditions()
```
zaneb, line 164: Starting from the defaults every time (rather than the current status) means that the "lastUpdated" timestamp will get updated every time, even when the condition value hasn't changed.

asalkeld: don't think there is anything to worry about: https://github.com/openshift/cluster-baremetal-operator/blob/master/controllers/clusteroperator.go#L192
https://github.com/openshift/cluster-baremetal-operator/blob/master/vendor/github.com/openshift/library-go/pkg/config/clusteroperator/v1helpers/status.go#L18
so that copies only changes across, so i don't think there is a problem

PR #36, opened 2020-10-08
Add operandVersions to the ClusterOperator's Status

asalkeld:
Done as per request: https://github.com/openshift/cluster-baremetal-operator/pull/32#discussion_r500446673

### DIFF
--- a/controllers/clusteroperator.go
+++ b/controllers/clusteroperator.go
@@ -64,6 +64,21 @@ func relatedObjects() []osconfigv1.ObjectReference {
 	}
 }
 
+// operandVersions returns the current list of OperandVersions for the
+// ClusterOperator objects's status.
+func (r *ProvisioningReconciler) operandVersions() []osconfigv1.OperandVersion {
+	operandVersions := []osconfigv1.OperandVersion{}
+
+	if r.ReleaseVersion != "" {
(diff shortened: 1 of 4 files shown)


controllers/clusteroperator.go
```diff
@@ -58,6 +58,20 @@ func relatedObjects() []osconfigv1.ObjectReference {
 	}
 }
 
+// operandVersions returns the current list of OperandVersions for the
+// ClusterOperator objects's status.
+func (r *ProvisioningReconciler) operandVersions() []osconfigv1.OperandVersion {
+	if r.ReleaseVersion == "" {
```
kirankt, line 64: Why can't we define an empty OperandVersion before we check for the env var. If the env var is defined we add values to the previously declared OperandVersion. That way, we only have to return once. Something like this perhaps?

```
        operandVersion := []osconfigv1.OperandVersion{}
        if r.releaseVersion != "" {
                operandVersions = append(operandVersions, osconfigv1.OperandVersion{Name: "operator", Version: r.releaseVersion})
        } 
        return operandVersion
```

kirankt (changes requested):
Let a small nit. Please let me know what you think.

controllers/clusteroperator.go
asalkeld: sure, i can do that

PR #38, opened 2020-10-08
Move the lint exclude rule to the config file

asalkeld:
This is so they are all in one place.

@sadasu note we can also disable rules in the code line-by-line, this is also a nice option if you want that rule checked but just have one instance of it ignored.
```go
thePassword = "srotet" // #nosec
```

### DIFF
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -21,8 +21,6 @@ linters:
 linters-settings:
   goimports:
     local-prefixes: "github.com/openshift"
-  govet:
-    check-shadowing: false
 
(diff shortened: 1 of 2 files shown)


sadasu:
I didn't find a way to disabled per line so went with the approach to exclude a rule. I would prefer keeping checking on for the rule but disable per line when we know it is ok. Thanks for the information.

sadasu:
@asalkeld what do you think of this? https://github.com/openshift/cluster-baremetal-operator/pull/41

PR #40, opened 2020-10-09
Add Owns(Secret, Deployment and ClusterOperator) to be watched

asalkeld:
### Main objective

Make sure that if owned resources are deleted or modified, the Reconcile is run to fix these resources.
Owned resources are
- deployment
- secrets
- clusterOperator

## Note to reviewers
This may make sense to review by commit.

[This commit](https://github.com/openshift/cluster-baremetal-operator/pull/40/commits/cef91263b2214abdb04c658da64eb43442dbb470) does the following
- add Owns() on ClusterOperator
- reworks how the ClusterOperator is created as it now needs a baremetalConfig (for the ownership)
- uses bindata to use the exact copy of the clusterOperator CR in manifests (This was does as the code was missing the required annotation and it does not seem right have multiple places for the same entity).
- It also improves how the initial status and ownership is applied. There was a case when the CO was created by the manifest, it would not have the same details as when created in code.


provisioning/baremetal_secrets.go
```diff
@@ -62,6 +64,16 @@ func CreateMariadbPasswordSecret(client coreclientv1.SecretsGetter, targetNamesp
 			ObjectMeta: metav1.ObjectMeta{
 				Name:      baremetalSecretName,
 				Namespace: targetNamespace,
+				OwnerReferences: []metav1.OwnerReference{
```
dhellmann, line 67: Is there any reason not to use SetControllerReference here, too? I guess that would mean moving this Secret to a variable, but that seems OK.

asalkeld: it just means passing in r.Schema as well. it just felt messy, i can change it and see what you think

sadasu (commented):
This is clever. I have a couple of minor questions inline.

provisioning/baremetal_secrets.go
```diff
@@ -45,7 +48,7 @@ func generateRandomPassword() (string, error) {
 }
 
 // CreateMariadbPasswordSecret creates a Secret for Mariadb password
-func CreateMariadbPasswordSecret(client coreclientv1.SecretsGetter, targetNamespace string) error {
+func CreateMariadbPasswordSecret(client coreclientv1.SecretsGetter, targetNamespace string, baremetalConfig *metal3iov1alpha1.Provisioning, scheme *runtime.Scheme) error {
 	_, err := client.Secrets(targetNamespace).Get(context.Background(), baremetalSecretName, metav1.GetOptions{})
```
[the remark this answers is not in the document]
asalkeld: ok, I can do that.

sadasu (approved):
/lgtm
Thanks for taking care of my comments.

PR #42, opened 2020-10-09
Generate config that are not provided in the Provisioning Spec

sadasu:


provisioning/baremetal_config.go
```diff
@@ -110,3 +117,80 @@ func validateDisabledConfig(prov *metal3iov1alpha1.Provisioning) error {
...
+		return &baremetalConfig.ProvisioningInterface
+	case "DEPLOY_KERNEL_URL":
+		return getDeployKernelUrl(baremetalConfig)
+	case "DEPLOY_RAMDISK_URL":
+		return getDeployRamdiskUrl(baremetalConfig)
+	case "IRONIC_ENDPOINT":
+		return getIronicEndpoint(baremetalConfig)
+	case "IRONIC_INSPECTOR_ENDPOINT":
+		return getIronicInspectorEndpoint(baremetalConfig)
+	case "HTTP_PORT":
```
asalkeld, line 187: import 	"k8s.io/utils/pointer"
return pointer.StringPtr(baremetalHttpPort)

provisioning/baremetal_config.go
```diff
@@ -110,3 +117,80 @@ func validateDisabledConfig(prov *metal3iov1alpha1.Provisioning) error {
...
+func getProvisioningOSDownloadURL(config *metal3iov1alpha1.ProvisioningSpec) *string {
+	if config.ProvisioningOSDownloadURL != "" {
+		return &(config.ProvisioningOSDownloadURL)
+	}
+	return nil
+}
+
+func getMetal3DeploymentConfig(name string, baremetalConfig *metal3iov1alpha1.ProvisioningSpec) *string {
+	configValue := ""
+	switch name {
```
asalkeld, line 174: these names should probably be consts

provisioning/baremetal_config.go
sadasu: Is importing a new library to manage pointers giving me anything other than avoiding the use of a temp variable?

sadasu: Done.

provisioning/baremetal_config.go
sadasu: Done.

kirankt (approved):
The only thing I'll mention is that we should get into a habit of writing a description for each PR. It'll help with the context for people not familiar with what the PR is about.

dhellmann:
It would be good to have some tests for these functions using IPv6 addresses.

PR #44, opened 2020-10-12
Binary rename

andfasano:
- Renaming initial kubebuilder `manager` resources to `cluster-baremetal-operator`
- Fixed makefile, Dockerfile and deployment manifest command


config/cluster-baremetal-operator/cluster-baremetal-operator.yaml
```diff
@@ -24,7 +24,7 @@ spec:
     spec:
       containers:
       - command:
-        - /manager
+        - /cluster-baremetal-operator
         args:
         - --enable-leader-election
         image: controller:latest
```
sadasu, line 30: Doe the name in the follwoing line have to change too?

andfasano: Probably it's not very relevant as I'm not sure we're going to use this yaml to generate the deployment file, anyhow I'll revert it back

sadasu: Lets make a change to lines 27 and 31 to be consistent everywhere.

andfasano:
> Is there actually any advantage to doing things differently to how kubebuilder generates them?

Are you referring to anything specific? We can keep the term "manager" for the operator but for the bin it's a pretty generic name. We can leave the renaming in the Dockerfile as I initially implemented if it's preferred.

PR #45, opened 2020-10-13
Create and apply the Metal3 Deployment object

sadasu:
- Create all Init containers required for the metal3 deployment
- Create all Containers required for the metal3 deployment
- Create Metal3 deployment object


provisioning/baremetal_pod.go
```diff
@@ -0,0 +1,353 @@
...
+			Value: *value,
+		}
+	} else {
+		return corev1.EnvVar{
+			Name: name,
+		}
+	}
+}
+
+func setMariadbPassword() corev1.EnvVar {
```
dhellmann, line 65: Could these environment variable definitions be constants, like the volume mounts are?

provisioning/baremetal_pod_test.go
```diff
@@ -0,0 +1,214 @@
...
+		t.Run(tc.name, func(t *testing.T) {
+			t.Logf("Testing tc : %s", tc.name)
+			actualContainers := newMetal3InitContainers(&images, &managedSpec)
+			assert.Equal(t, len(tc.expectedContainers), len(actualContainers), fmt.Sprintf("%s : Expected number of Init Containers : %d Actual number of Init Containers : %d", tc.name, len(tc.expectedContainers), len(actualContainers)))
+		})
+	}
+
+}
+
+func TestNewMetal3Containers(t *testing.T) {
```
andfasano, line 154: I think that such test (and from a broader point of view the one on the Deployment, not yet present) deserves all the possible attentions, since it represents a core feature of the operator: an error in one of the generated fields could potentially compromise the entire metal3 deployment. 
So, from this point of view, it could be worth to "enrich" such kind of tests by taking care that all the fields contain the expected value - instead of just counting the number of containers. Thus, it seems a perfect fit for the golden file testing approach: I've prepared a sample PR to illustrate the approach (see https://github.com/sadasu/cluster-baremetal-operator/pull/3). It could work well directly on the generated overall deployment

provisioning/baremetal_pod.go
sadasu: Done

provisioning/baremetal_pod_test.go
sadasu: Thanks for this suggestion @andfasano. I was looking for ways to better test the Deployment object. I'll use your suggest technique for the Deployment object.

provisioning/baremetal_pod.go
dhellmann: What I meant was, move lines 66-75 up to the top of the file (say around line 50) and define them as variables like `inspectorCredentialsMount` instead of using a function that returns a new object created from constant data each time.

provisioning/baremetal_pod.go
```diff
@@ -351,3 +356,118 @@ func createContainerMetal3StaticIpManager(images *Images, config *metal3iov1alph
...
+					Items: []corev1.KeyToPath{
+						{Key: ironicUsernameKey, Path: ironicUsernameKey},
+						{Key: ironicPasswordKey, Path: ironicPasswordKey},
+						{Key: ironicConfigKey, Path: ironicConfigKey},
+					},
+				},
+			},
+		},
+		{
+			Name: inspectorCredentialsVolume,
```
dhellmann, line 382: It looks like this same volume definition is included twice.

provisioning/baremetal_pod.go
```diff
@@ -351,3 +356,118 @@ func createContainerMetal3StaticIpManager(images *Images, config *metal3iov1alph
 	}
 	return container
 }
+
+func newMetal3Volumes() []corev1.Volume {
```
dhellmann, line 360: Is this list of volumes static? Do we need a function to return it, or could we define it in a variable, too?

provisioning/baremetal_pod.go
dhellmann: Oh, never mind, "ironic" vs. "inspector".

provisioning/baremetal_pod.go
sadasu: Done. Added vars for marladbPassword and metal3Volumes instead of methods.

provisioning/baremetal_pod.go
sadasu: Done. https://github.com/openshift/cluster-baremetal-operator/pull/45#discussion_r510465229

dhellmann:
This looks correct and fairly clean now. We can continue to iterate on it next sprint, if we need to.

/lgtm

dhellmann:
It occurred to me after saying lgtm that landing this without the change to annotate an existing deployment might put the MAO and CBO into a race, and since the CVO is installing CBO now that might cause issues for other users of the baremetal platform. Let's talk about it before we merge this to make sure we have things happening in the right order.

dhellmann:
This version looks safe, since it leaves the existing deployment alone.

/lgtm

PR #47, opened 2020-10-19
Add missing KubeClient initialization

sadasu:
The Client added by KubeBuilder reads resource values from a cache but writes to the API server. Since the code that creates resources for the metal3 deployment creates resources just before using them, we are taking the prudent approach of adding a regualr k8s client that reads and writes to the API Server. 
Some more experimentation is required before we can confidently use the Client to create and update the metal3 Deployment object and the resources that it depends on.


controllers/provisioning_controller.go
```diff
@@ -54,7 +54,7 @@ type ProvisioningReconciler struct {
 	Log            logr.Logger
 	OSClient       osclientset.Interface
 	EventRecorder  record.EventRecorder
-	kubeClient     kubernetes.Interface
+	KubeClient     kubernetes.Interface
```
dhellmann, line 57: What is the difference between this client and the one on line 52?

sadasu: KubeClient always writes and reads from the API server but the Client reads from a cache but writes to the API server. While creating the Secrets, I think either clients would be OK to do the job. But when we need to read the Secrets to eventually create the Deployment object, using a client that did not read from the local cache seemed like a more predictable approach to take.

dhellmann: Interesting. Is that what other operators do?

sadasu: Yes, other operators directly use the K8s client. The "Client" was added/generated by KubeBuilder and it *might* work even with the caching. The current focus is on getting metal3 deployment working using the k8s client. Once we have that, we could experiment using this client over the k8s client.

controllers/provisioning_controller.go
```diff
@@ -159,13 +159,13 @@ func (r *ProvisioningReconciler) Reconcile(req ctrl.Request) (ctrl.Result, error
 	}
 
 	//Create Secrets needed for Metal3 deployment
-	if err := provisioning.CreateMariadbPasswordSecret(r.kubeClient.CoreV1(), ComponentNamespace); err != nil {
+	if err := provisioning.CreateMariadbPasswordSecret(r.KubeClient.CoreV1(), ComponentNamespace); err != nil {
```
dhellmann, line 162: If the client wasn't configured at all before, how were these calls working in the test suite?

sadasu: We initialize them as Fake clients in the test suite.

sadasu: This was an oversight on my end. I knew it had to be done but somehow missed it the the commit.

dhellmann (changes requested):
The code changes look good, but we had a bit of discussion about why they are needed that would make useful detail in the commit messages.

sadasu:
> The code changes look good, but we had a bit of discussion about why they are needed that would make useful detail in the commit messages.

Updated the commit message. Please let me know if that is a good summary of the discussion.

dhellmann:
The pull request message doesn't actually go into the git history as a commit message. I was suggesting that you amend the commit that instantiates the client with those details. What you've written looks good.

AFAIK, the cache client does update the local cache when a resource is written, so I suspect we don't need this, but if it's what the other operators do then it's OK.

sadasu:
> The pull request message doesn't actually go into the git history as a commit message. I was suggesting that you amend the commit that instantiates the client with those details. What you've written looks good.
> 
> AFAIK, the cache client does update the local cache when a resource is written, so I suspect we don't need this, but if it's what the other operators do then it's OK.

I suspect that is the case. That would be one of the 1st things we will be trying when the deployment works e2e with the KubeClient.

PR #50, opened 2020-10-20
CVO-CBO integration

andfasano:
This PR adds the roles currently needed by CBO.


Dockerfile
```diff
@@ -8,6 +8,6 @@ RUN make cluster-baremetal-operator
 FROM registry.svc.ci.openshift.org/ocp/4.6:base
 COPY --from=builder /go/src/github.com/openshift/cluster-baremetal-operator/bin/cluster-baremetal-operator /usr/bin/cluster-baremetal-operator
 COPY --from=builder /go/src/github.com/openshift/cluster-baremetal-operator/manifests /manifests
-# Uncomment when ready for the release so that CVO could manage the operator
-# LABEL io.openshift.release.operator=true
+
+LABEL io.openshift.release.operator=true
```
andfasano, line 12: This is currently required in the PR to force the CVO handling the CBO manifest in the ephemeral release paylod

stbenjam: Do you know if that's all we need to do? I've never added an SLO.  Containers further up the chain need a reference in image-references to get pulled into a release payload, but CVO doesn't have one. 

For example, the installer is included in the release payload by cluster-samples-operator: https://github.com/openshift/cluster-samples-operator/blob/master/manifests/image-references

andfasano: Yes, we also added other manifests in the `/manifests` folders, as per CVO requirements. Without the label anyhow CVO will ignore the CBO images and will not pickup its manifests. Since we're are not sure that enabling it will impact the release, we'll keep it disable until the integration with the ART release will be completed (just activated in this PR for testing purpouses).

stbenjam: Right, but the manifests/ directory in CBO is for it's images. I was referring to cluster-baremetal-operator itself being picked up as part of the release payload. I guess the label is what should do it for second-level operators.

manifests/0000_31_cluster-baremetal-operator_05_rbac.yaml
```diff
@@ -0,0 +1,431 @@
...
+      - create
+
+  - apiGroups:
+      - authorization.k8s.io
+    resources:
+      - subjectaccessreviews
+    verbs:
+      - create
+
+# TODO(vikasc): Remove extensions/daemonsets permissions once all controllers have bumped kubernetes-drain
```
sadasu, line 142: We should remove this section because we are not deploying any Daemonsets yet.

andfasano: Yes, and there's a lot more to remove. This is just a test to verify that the deployment will work using essentially the same MAO rbac roles. If so, we can start to remove the excess. For example, it's not need any role on the `machine.openshift.io` or `openshift-config` namespaces

sadasu:
/lgtm

Matches my current understanding of the Roles required by the CBO.

PR #51, opened 2020-10-22
add current profile annotations to CVO manifests

guillaumerose:
This is matches openshift/enhancements#414 and doesn't change existing behavior

### DIFF
--- a/manifests/0000_31_cluster-baremetal-operator_01_images.configmap.yaml
+++ b/manifests/0000_31_cluster-baremetal-operator_01_images.configmap.yaml
@@ -4,6 +4,8 @@ kind: ConfigMap
 metadata:
   name: cluster-baremetal-operator-images
   namespace: openshift-machine-api
+  annotations:
+    include.release.openshift.io/self-managed-high-availability: "true"
 data:
   images.json: >
     {
--- a/manifests/0000_31_cluster-baremetal-operator_02_metal3provisioning.crd.yaml
+++ b/manifests/0000_31_cluster-baremetal-operator_02_metal3provisioning.crd.yaml
@@ -4,6 +4,7 @@ apiVersion: apiextensions.k8s.io/v1
 kind: CustomResourceDefinition
 metadata:
   annotations:
+    include.release.openshift.io/self-managed-high-availability: "true"
   creationTimestamp: null
   name: provisionings.metal3.io
 spec:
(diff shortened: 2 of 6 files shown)


andfasano:
Hi @guillaumerose I have a question: the enhancement proposes that manifests must contain at least one inclusion annotation, and the default profile is `self-managed-highly-available`. I was looking for a more formal definition in the enh proposal for `self-managed-highly-available` but didn't found anything (except for [this](https://github.com/openshift/enhancements/blob/6eb55f8a78318b977420c4bddc1e5a2474ad50ac/enhancements/update/cluster-profiles.md#design)). Do you have any link with additional info and/or can clarify? Thanks

guillaumerose:
Hi @andfaso. The naming discussion started in this comment https://github.com/openshift/enhancements/pull/414#issuecomment-675491233. I don't have much than this thread. The main point is that they prefer explicit over implicit inclusion and `default` as a name was not really a good name.
This change is also tracked here: https://bugzilla.redhat.com/show_bug.cgi?id=1871890

andfasano:
> Hi @andfaso. The naming discussion started in this comment [openshift/enhancements#414 (comment)](https://github.com/openshift/enhancements/pull/414#issuecomment-675491233). I don't have much than this thread. The main point is that they prefer explicit over implicit inclusion and `default` as a name was not really a good name.
> This change is also tracked here: https://bugzilla.redhat.com/show_bug.cgi?id=1871890

Thanks. Some of the existing resources already had an exclusion profile define (`exclude.release.openshift.io/internal-openshift-hosted: "true"`), do you maybe have any idea if it's ok to have both the exclusion _and_ inclusion annotations set (I think so)?

dhellmann:
> > Hi @andfaso. The naming discussion started in this comment [openshift/enhancements#414 (comment)](https://github.com/openshift/enhancements/pull/414#issuecomment-675491233). I don't have much than this thread. The main point is that they prefer explicit over implicit inclusion and `default` as a name was not really a good name.
> > This change is also tracked here: https://bugzilla.redhat.com/show_bug.cgi?id=1871890
> 
> Thanks. Some of the existing resources already had an exclusion profile define (`exclude.release.openshift.io/internal-openshift-hosted: "true"`), do you maybe have any idea if it's ok to have both the exclusion _and_ inclusion annotations set (I think so)?

Yes, it should be fine. Only the annotation for the profile being used is examined. See https://github.com/openshift/enhancements/blob/master/enhancements/update/cluster-profiles.md for details.

/approve

PR #52, opened 2020-10-26
use "make manifests" to generate crd,rbac,deploment in manifests/

asalkeld:
Now "make manifests" generates the following into manifests/:
- crd
- rbac
- deployment

Note: to change the roles a developer needs to edit the markup in the controller.


dhellmann:
Can we use this same approach with the other manifests, too? It would be nice to be able to be consistent.

asalkeld:
> Can we use this same approach with the other manifests, too? It would be nice to be able to be consistent.

@dhellmann should be done now. I have checked the generated roles against the original as best I can.. sorry the diff is
not pretty as kustomize sorts everything.

manifests/0000_31_cluster-baremetal-operator_05_rbac.yaml
```diff
@@ -1,150 +1,125 @@
 apiVersion: rbac.authorization.k8s.io/v1
 kind: Role
 metadata:
+  creationTimestamp: null
   name: cluster-baremetal-operator
   namespace: openshift-machine-api
```
dhellmann, line 6: We're going to want the `include` annotation for the default cluster profile on everything resource we place in the manifests directory, otherwise the CVO will ignore it.

dhellmann:
> > Can we use this same approach with the other manifests, too? It would be nice to be able to be consistent.
> 
> @dhellmann should be done now. I have checked the generated roles against the original as best I can.. sorry the diff is
> not pretty as kustomize sorts everything.

Yeah, whew. Still, I think this is moving in the right direction.

The cluster profile annotation needs to be on every resource, which is why I asked about adding it via the same patching approach. I think what you have here would need to be amended to include the patch instruction for those manifests, right? Is there a way to have just 1 patch file to do that or do we have to reproduce the patch for different subdirectories under `config`?

We don't expect to have a lot of profiles in the future, but there will be more. Part of the point of the profiles is that we might have different deployment configurations. In our case, that's likely to mean passing different arguments to the operator or something. That seems like something kustomize would be able to help with, so while we shouldn't necessarily do anything about it now we should be thinking ahead about how we could.

asalkeld:
> > > Can we use this same approach with the other manifests, too? It would be nice to be able to be consistent.
> > 
> > 
> > @dhellmann should be done now. I have checked the generated roles against the original as best I can.. sorry the diff is
> > not pretty as kustomize sorts everything.
> 
> Yeah, whew. Still, I think this is moving in the right direction.
> 
> The cluster profile annotation needs to be on every resource, which is why I asked about adding it via the same patching approach. I think what you have here would need to be amended to include the patch instruction for those manifests, right? Is there a way to have just 1 patch file to do that or do we have to reproduce the patch for different subdirectories under `config`?
> 
> We don't expect to have a lot of profiles in the future, but there will be more. Part of the point of the profiles is that we might have different deployment configurations. In our case, that's likely to mean passing different arguments to the operator or something. That seems like something kustomize would be able to help with, so while we shouldn't necessarily do anything about it now we should be thinking ahead about how we could.

done, enjoy ;-)

stbenjam:
Needs rebase

Makefile
```diff
@@ -33,43 +35,58 @@ run: generate lint manifests
...
+	manifests/rbac.authorization.k8s.io_v1_clusterrole_cluster-baremetal-operator.yaml \
+	manifests/rbac.authorization.k8s.io_v1_rolebinding_cluster-baremetal-operator.yaml \
+  manifests/rbac.authorization.k8s.io_v1_clusterrolebinding_cluster-baremetal-operator.yaml
 
 # Generate manifests e.g. CRD, RBAC etc.
 .PHONY: manifests
-manifests:
-	hack/gen-crd.sh
+manifests: generate
+	$(KUSTOMIZE) build config/profiles/$(MANIFEST_PROFILE) -o manifests/
```
andfasano, line 58: My concern here is about generating directly the kustomize output into the `/manifests` folder. Whatever ends up here will be loaded by CVO during installation/upgrade. What about storing the intermediate generation into a temp dir, and then moving just the selected rendered artifacts into `/manifests`?

asalkeld: ok, i can do that.

PR #54, opened 2020-10-28
Use vendorized golangci-lint

honza:
* Add golangci-lint to tools.go so it shows up in go.mod
* Replace existing curl install with a call to `go build`
* go mod vendor

This patch is necessary because the downstream build system doesn't have
internet access, and thus can't install golangci-lint via curl.


andfasano:
lint/govet tests passed, I think they're enough for this change?

PR #61, opened 2020-10-30
WIP: controllers: "Unsupported", etc. condition messaging

wking:
"Operator is non-functional" sounds a bit frightening.  This commit pivots to say "Operator has no role on platform {platform-name}" in the cases where we determine we are not needed.

I'm also removing ReasonEmpty, because it's always nice to say *something* about why we picked the state we picked.  In the `ReasonInvalidConfiguration` and `ReasonDeployTimedOut` cases, that makes me wonder if we really intend to be Available then.  Maybe we should just echo whatever we're setting for Degraded in that case?  FIXME/WIP until we decide.

### DIFF
--- a/controllers/clusteroperator.go
+++ b/controllers/clusteroperator.go
@@ -22,8 +22,8 @@ const (
 	// OperatorDisabled represents a Disabled ClusterStatusConditionTypes
 	OperatorDisabled osconfigv1.ClusterStatusConditionType = "Disabled"
 
-	// ReasonEmpty is an empty StatusReason
-	ReasonEmpty StatusReason = ""
+	// ReasonAsExpected is a status reason for the happy state, when we have nothing else to say.
+	ReasonAsExpected StatusReason = "AsExpected"
 
 	// ReasonComplete the deployment of required resources is complete
 	ReasonComplete StatusReason = "DeployComplete"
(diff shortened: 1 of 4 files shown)


controllers/clusteroperator.go
```diff
@@ -179,7 +179,7 @@ func (r *ProvisioningReconciler) updateCOStatus(newReason StatusReason, msg, pro
 		v1helpers.SetStatusCondition(&conds, setStatusCondition(osconfigv1.OperatorProgressing, osconfigv1.ConditionFalse, string(newReason), progressMsg))
 	case ReasonInvalidConfiguration, ReasonDeployTimedOut:
 		v1helpers.SetStatusCondition(&conds, setStatusCondition(osconfigv1.OperatorDegraded, osconfigv1.ConditionTrue, string(newReason), msg))
-		v1helpers.SetStatusCondition(&conds, setStatusCondition(osconfigv1.OperatorAvailable, osconfigv1.ConditionTrue, string(ReasonEmpty), ""))
+		v1helpers.SetStatusCondition(&conds, setStatusCondition(osconfigv1.OperatorAvailable, osconfigv1.ConditionTrue, string(ReasonAsExpected), "FIXME: are we really available here?"))
```
stbenjam, line 182: I think Degraded makes sense

wking: I'm fine with Degraded=True here.  I'm not clear on Available=True.  What functionality is the operator available to fulfil when it has ReasonInvalidConfiguration or ReasonDeployTimedOut?

sadasu: We looked into several options detailed here ; https://github.com/openshift/enhancements/blob/master/enhancements/baremetal/an-slo-for-baremetal.md#not-in-use-slo-behaviors

There seems to be no defined behavior when a CO needs to convey that the Operator is running, is non-functional and that is expected.

wking: Neither `ReasonInvalidConfiguration` nor `ReasonDeployTimedOut` are "expected non-functional", are they?  That is `ReasonUnsupported`, which I also touch in this PR, but it is not what this thread/line is about.

sadasu: Agreed. Giving context for you question if Available=True is required.

wking: `Available` is definitely require.  In master, `ReasonInvalidConfiguration` and `ReasonDeployTimedOut` are `Available=True` `Degraded=False`.  My naive reaction to the reason strings has me suspecting the might actually call for `Available=False` `Degraded=False`.  Can you list and services available (but degraded) under these conditions?

sadasu: > I'm fine with Degraded=True here. I'm not clear on Available=True. What functionality is the operator available to fulfil when it has ReasonInvalidConfiguration or ReasonDeployTimedOut?

The Operator is Available because it is still watching its resources. If the configuration changes to a valid value for example, then the operator is able to get itself out of the Degraded state. That is how I view the Available=True in this scenario.

sadasu:
/close
I believe this is already fixed.

PR #63, opened 2020-11-04
Set Baremetal CO to Disabled before Reconciler runs

sadasu:
Current implementation sets the Baremetal CO to Disabled in Platforms other than BareMetal before starting the Reconciler.

Alternate approach: (re-worded based on @andfasano 's feedback below)
The CO manifest can be defined with the CO in Disabled state. So, when CVO installs all the manifests in the /manifest directory, the Baremetal CO starts in the Disabled state. On the BareMetal platform, the CO resource would be amended within the reconcile loop.

### DIFF
--- a/controllers/clusteroperator.go
+++ b/controllers/clusteroperator.go
@@ -245,6 +245,7 @@ func SetCOInDisabledState(osClient osclientset.Interface, version string) error 
 		v1helpers.SetStatusCondition(&co.Status.Conditions, c)
 	}
 	co.Status.Versions = operandVersions(version)
+	co.Status.RelatedObjects = relatedObjects()
 
 	_, err = osClient.ConfigV1().ClusterOperators().UpdateStatus(context.Background(), co, metav1.UpdateOptions{})
 	return err
--- a/controllers/clusteroperator.go
+++ b/controllers/clusteroperator.go
@@ -32,7 +32,7 @@ const (
 	// ReasonEmpty is an empty StatusReason
 	ReasonEmpty StatusReason = ""
 
-	// ReasonEmpty is an empty StatusReason
+	// ReasonExpected indicates that the operator is behaving as expected
 	ReasonExpected StatusReason = "AsExpected"
 
 	// ReasonComplete the deployment of required resources is complete
--- a/controllers/clusteroperator.go
+++ b/controllers/clusteroperator.go
@@ -32,6 +32,9 @@ const (
 	// ReasonEmpty is an empty StatusReason
 	ReasonEmpty StatusReason = ""
 
+	// ReasonEmpty is an empty StatusReason
+	ReasonExpected StatusReason = "AsExpected"
+
 	// ReasonComplete the deployment of required resources is complete
 	ReasonComplete StatusReason = "DeployComplete"
 
@@ -238,8 +241,8 @@ func SetCOInDisabledState(osClient osclientset.Interface, version string) error 
 	}
 
 	conds := defaultStatusConditions()
-	v1helpers.SetStatusCondition(&conds, setStatusCondition(OperatorDisabled, osconfigv1.ConditionTrue, "Operator on Standby", "Unsupported Platform"))
-	v1helpers.SetStatusCondition(&conds, setStatusCondition(osconfigv1.OperatorAvailable, osconfigv1.ConditionTrue, string(ReasonEmpty), ""))
+	v1helpers.SetStatusCondition(&conds, setStatusCondition(OperatorDisabled, osconfigv1.ConditionTrue, string(ReasonUnsupported), "Nothing to do on this Platform"))
+	v1helpers.SetStatusCondition(&conds, setStatusCondition(osconfigv1.OperatorAvailable, osconfigv1.ConditionTrue, string(ReasonExpected), "Operational"))
 
 	for _, c := range conds {
 		v1helpers.SetStatusCondition(&co.Status.Conditions, c)
--- a/controllers/clusteroperator.go
+++ b/controllers/clusteroperator.go
@@ -14,6 +14,7 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/controller/controllerutil"
 
 	osconfigv1 "github.com/openshift/api/config/v1"
+	osclientset "github.com/openshift/client-go/config/clientset/versioned"
 	metal3iov1alpha1 "github.com/openshift/cluster-baremetal-operator/api/v1alpha1"
 	"github.com/openshift/library-go/pkg/config/clusteroperator/v1helpers"
 )
@@ -82,13 +83,13 @@ func relatedObjects() []osconfigv1.ObjectReference {
 
 // operandVersions returns the current list of OperandVersions for the
 // ClusterOperator objects's status.
-func (r *ProvisioningReconciler) operandVersions() []osconfigv1.OperandVersion {
+func operandVersions(version string) []osconfigv1.OperandVersion {
 	operandVersions := []osconfigv1.OperandVersion{}
 
-	if r.ReleaseVersion != "" {
+	if version != "" {
 		operandVersions = append(operandVersions, osconfigv1.OperandVersion{
 			Name:    "operator",
-			Version: r.ReleaseVersion,
+			Version: version,
 		})
 	}
 
@@ -144,9 +145,9 @@ func (r *ProvisioningReconciler) ensureClusterOperator(baremetalConfig *metal3io
 		needsUpadate = true
 		co.Status.RelatedObjects = relatedObjects()
 	}
-	if !equality.Semantic.DeepEqual(co.Status.Versions, r.operandVersions()) {
+	if !equality.Semantic.DeepEqual(co.Status.Versions, operandVersions(r.ReleaseVersion)) {
 		needsUpadate = true
-		co.Status.Versions = r.operandVersions()
+		co.Status.Versions = operandVersions(r.ReleaseVersion)
 	}
 	if len(co.Status.Conditions) == 0 {
 		needsUpadate = true
@@ -190,7 +191,7 @@ func (r *ProvisioningReconciler) syncStatus(co *osconfigv1.ClusterOperator, cond
 
 	if len(co.Status.Versions) < 1 {
 		r.Log.Info("updating ClusterOperator Status Versions field")
-		co.Status.Versions = r.operandVersions()
+		co.Status.Versions = operandVersions(r.ReleaseVersion)
 	}
 
 	_, err := r.OSClient.ConfigV1().ClusterOperators().UpdateStatus(context.Background(), co, metav1.UpdateOptions{})
@@ -208,7 +209,7 @@ func (r *ProvisioningReconciler) updateCOStatus(newReason StatusReason, msg, pro
 	switch newReason {
 	case ReasonUnsupported:
 		v1helpers.SetStatusCondition(&conds, setStatusCondition(OperatorDisabled, osconfigv1.ConditionTrue, string(newReason), msg))
-		v1helpers.SetStatusCondition(&conds, setStatusCondition(osconfigv1.OperatorAvailable, osconfigv1.ConditionTrue, string(ReasonEmpty), ""))
+		v1helpers.SetStatusCondition(&conds, setStatusCondition(osconfigv1.OperatorAvailable, osconfigv1.ConditionTrue, string(ReasonExpected), "Operational"))
 	case ReasonSyncing:
 		v1helpers.SetStatusCondition(&conds, setStatusCondition(osconfigv1.OperatorAvailable, osconfigv1.ConditionTrue, string(newReason), msg))
 		v1helpers.SetStatusCondition(&conds, setStatusCondition(osconfigv1.OperatorProgressing, osconfigv1.ConditionTrue, string(newReason), progressMsg))
@@ -226,4 +227,25 @@ func (r *ProvisioningReconciler) updateCOStatus(newReason StatusReason, msg, pro
 	}
 
 	return r.syncStatus(co, conds)
+}
+
+func SetCOInDisabledState(osClient osclientset.Interface, version string) error {
+	//The CVO should have created the CO
+	co, err := osClient.ConfigV1().ClusterOperators().Get(context.Background(), clusterOperatorName, metav1.GetOptions{})
+
+	if err != nil {
+		return fmt.Errorf("failed to get clusterOperator %q: %v", clusterOperatorName, err)
+	}
+
+	conds := defaultStatusConditions()
+	v1helpers.SetStatusCondition(&conds, setStatusCondition(OperatorDisabled, osconfigv1.ConditionTrue, "Operator on Standby", "Unsupported Platform"))
+	v1helpers.SetStatusCondition(&conds, setStatusCondition(osconfigv1.OperatorAvailable, osconfigv1.ConditionTrue, string(ReasonEmpty), ""))
+
+	for _, c := range conds {
+		v1helpers.SetStatusCondition(&co.Status.Conditions, c)
+	}
+	co.Status.Versions = operandVersions(version)
+
+	_, err = osClient.ConfigV1().ClusterOperators().UpdateStatus(context.Background(), co, metav1.UpdateOptions{})
+	return err
 }
--- a/controllers/provisioning_controller.go
+++ b/controllers/provisioning_controller.go
@@ -149,6 +149,34 @@ func (r *ProvisioningReconciler) Reconcile(req ctrl.Request) (ctrl.Result, error
 		return ctrl.Result{}, nil
 	}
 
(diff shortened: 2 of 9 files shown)


stbenjam (changes requested):
Can you update the dockerfile to uncomment the release label, so we see if e2e-agnostic passes with the CBO as part of the release payload?

main.go
```diff
@@ -93,6 +93,21 @@ func main() {
...
+	enabled, err := controllers.IsEnabled(mgr.GetClient())
+	if err != nil {
+		setupLog.Error(err, "unable to determine Infrastructure Platform type")
+		os.Exit(1)
+	}
+	if !enabled {
+		// set ClusterOperator status to disabled=true, available=true
+		err = controllers.SetCOInDisabledState(osClient, releaseVersion)
+		if err != nil {
+			setupLog.Error(err, "unable to Baremetal CO to disabled")
```
stbenjam, line 106: is a verb missing here?

andfasano: Also, is it ok to exit in such case?

sadasu: Fixed the verb.

Regarding whether to continue with setup of the operator or exit was not a clear choice for me.
If the CO is not updated correctly, the CVO considers the Operator to be in a bad state anyways. And in the case of non-Baremetal platforms, this code will run only once so there is no path where this will correct itself (which can happen when we update the CO in the reconcile methond).

Please check the "Althernate approach" that I listed above and let me know what you think.

andfasano: My feeling is that the alternate approach could be a little bit more clear. The resource will be amended within a reconcile loop, which could be a better place than the main entry point of the operator.

controllers/clusteroperator.go
```diff
@@ -189,3 +193,25 @@ func (r *ProvisioningReconciler) updateCOStatus(newReason StatusReason, msg, pro
...
+func SetCOInDisabledState(osClient osclientset.Interface, version string) error {
+	//The CVO should have created the CO
+	co, err := osClient.ConfigV1().ClusterOperators().Get(context.Background(), clusterOperatorName, metav1.GetOptions{})
+
+	if err != nil {
+		return fmt.Errorf("failed to get clusterOperator %q: %v", clusterOperatorName, err)
+	}
+
+	conds := defaultStatusConditions()
+	v1helpers.SetStatusCondition(&conds, setStatusCondition(OperatorDisabled, osconfigv1.ConditionTrue, string(ReasonUnsupported), "Nothing to do on this Platform"))
```
sadasu, line 206: @wking I tried to incorporate some of your suggestions from https://github.com/openshift/cluster-baremetal-operator/pull/61.

stbenjam:
This looks good to me, and e2e-agnostic is passing.

@wking Would you mind taking a look at this?

/assign @wking

controllers/provisioning_controller.go
```diff
@@ -112,21 +108,6 @@ func (r *ProvisioningReconciler) readProvisioningCR(req ctrl.Request) (*metal3io
 func (r *ProvisioningReconciler) Reconcile(req ctrl.Request) (ctrl.Result, error) {
 	//log := r.Log.WithValues("provisioning", req.NamespacedName)
 
-	enabled, err := r.isEnabled()
```
andfasano, line 115: On a second thought, what about keeping just a safety check to prevent the reconcile loop running, in case the Provisioning resource will be deployed by mistake?

sadasu: I am not sure if we can claim to be "Available=True" without the reconcile loop running. Let me look into this some more.

dhellmann:
I don't think setting the ClusterOperator status in main() is going to be sufficient to address the problem.

According to https://github.com/openshift/cluster-version-operator/blob/master/docs/dev/clusteroperator.md#how-clusterversionoperator-handles-clusteroperator-in-release-image, the cluster-version-operator creates the CO resource initially but each operator is responsible for re-creating its own ClusterOperator resource if it is deleted. So, if we only set the status to Disabled in main() then there will be no code to handle that delete case.

Since we can't ensure our Provisioning API resource is always present, I think we're going to actually need another controller added to this process to manage the disabled case. If we create a controller that reconciles on the ClusterOperator resource then it can detect the disabled state and set the CO values accordingly. It could also detect when the CO is deleted, and then schedule a reconcile *after* that to recreate the resource (we will be able to tell that case because we're in the reconcile loop looking for a resource that does not exist). The cluster-authentication-operator has something like this, using the controller defined in library-go (see https://github.com/openshift/cluster-authentication-operator/blob/master/vendor/github.com/openshift/library-go/pkg/operator/status/status_controller.go). I don't know if we will be able to use that implementation directly.

The existing controller that is reconciling on the Provisioning resource should continue to check if the operator is enabled and do nothing in that case (by restoring the isEnabled() call that Andrea commented on).

controllers/provisioning_controller.go
sadasu: I misread your comment earlier. I think we need to add this check back not because the controller now watches the CO resource too. (https://github.com/openshift/cluster-baremetal-operator/blob/master/controllers/provisioning_controller.go#L239)

sadasu:
@dhellmann these were my concerns also. This PR just takes care of the successful install of CVO in non-Baremetal environments and doesn't really take care of the steady state scenarios.

I think https://github.com/openshift/cluster-baremetal-operator/pull/40 brings in functionality that can be used maintaining the CO in the correct state after initial bringup.

Specifically, https://github.com/openshift/cluster-baremetal-operator/blob/master/controllers/provisioning_controller.go#L239. With that change, the CO is another resource that is being watched. So, the reconcile loop would be called for any changes to the "baremetal" CO.

With this change in place, it makes sense to add back the isEnabled() check within the Reconciler and the updateCOState() already takes care of creating the baremetal CO if it doesn't exist.

dhellmann:
> @dhellmann these were my concerns also. This PR just takes care of the successful install of CVO in non-Baremetal environments and doesn't really take care of the steady state scenarios.
> 
> I think #40 brings in functionality that can be used maintaining the CO in the correct state after initial bringup.

OK, good, I hadn't made the connection between that work and re-creating the ClusterOperator.

controllers/clusteroperator.go
```diff
@@ -31,6 +32,9 @@ const (
 	// ReasonEmpty is an empty StatusReason
 	ReasonEmpty StatusReason = ""
 
+	// ReasonEmpty is an empty StatusReason
+	ReasonExpected StatusReason = "AsExpected"
```
dhellmann, line 36: The comment does not match the variable here.

main.go
```diff
@@ -96,6 +96,21 @@ func main() {
 	}
 
 	osClient := osclientset.NewForConfigOrDie(rest.AddUserAgent(config, controllers.ComponentName))
+	// Check the Platform Type to determine the state of the CO
+	enabled, err := controllers.IsEnabled(osClient)
+	if err != nil {
+		setupLog.Error(err, "unable to determine Infrastructure Platform type")
```
dhellmann, line 102: This error message assumes knowledge of what controllers.IsEnabled() is doing. Here we only know we are checking if we are enabled, and we can't figure that out but don't actually know why. I don't think it's worth holding up this PR over it, but we should plan a clean up patch in another PR.

main.go
```diff
@@ -96,6 +96,21 @@ func main() {
...
+		os.Exit(1)
+	}
+	if !enabled {
+		//Set ClusterOperator status to disabled=true, available=true
+		err = controllers.SetCOInDisabledState(osClient, releaseVersion)
+		if err != nil {
+			setupLog.Error(err, "unable to set Baremetal CO to disabled")
```
dhellmann, line 109: Someone reading the log may not know what a "CO" is. We should spell it out. The name is also "baremetal" not "Baremetal", right? Again, not worth holding this up, so we can fix it in the same PR that fixes the other message.

dhellmann (commented):
/lgtm

This looks good. We can iterate on the messages (see comments inline) in another PR.

/hold

Let's wait to merge it until our morning in case something breaks and we have to revert quickly again.

stbenjam:
e2e-metal-ipi failing with: 

```
 level=fatal msg=failed to initialize the cluster: Cluster operator baremetal is still updating 
```

main.go
sadasu: Done

main.go
sadasu: Done

controllers/clusteroperator.go
sadasu: Fixed.

controllers/clusteroperator.go
```diff
@@ -227,3 +231,25 @@ func (r *ProvisioningReconciler) updateCOStatus(newReason StatusReason, msg, pro
...
+	v1helpers.SetStatusCondition(&conds, setStatusCondition(OperatorDisabled, osconfigv1.ConditionTrue, string(ReasonUnsupported), "Nothing to do on this Platform"))
+	v1helpers.SetStatusCondition(&conds, setStatusCondition(osconfigv1.OperatorAvailable, osconfigv1.ConditionTrue, string(ReasonExpected), "Operational"))
+
+	for _, c := range conds {
+		v1helpers.SetStatusCondition(&co.Status.Conditions, c)
+	}
+	co.Status.Versions = operandVersions(version)
+	co.Status.RelatedObjects = relatedObjects()
+
+	_, err = osClient.ConfigV1().ClusterOperators().UpdateStatus(context.Background(), co, metav1.UpdateOptions{})
```
dhellmann, line 253: I'm still worried about the error handling here. Do we receive an error if the ClusterOperator isn't going to be changed because it is already set with the same conditions?

wking: I don't see why a no-op update would error.  But maybe patch, instead of update, if you're worried about generation noise?

sadasu: SetCOInDisabledState() is called once before the start of the reconcile loop. At that time, the baremetal ClusterOperator is created by CVO using whatever is provided in /manifests. That is a bare CO with no StatusConditions, Version or RelatedObjects. So, this will be the 1st time that these values would be set.
For all other times we are updating the CO (including putting it in Disabled state) from within the reconcile loop, we use  updateCOStatus() to do that.

dhellmann: I've seen errors updating other resources without changes, but maybe those were different cases. If we think this is OK, then I think this PR is probably ready to merge.

dhellmann:
This looks ready to me. I'll leave my lgtm and we'll see what CI says.

/lgtm

sadasu:
Not related to CBO or the baremetal CO.

/test e2e-agnostic

controllers/provisioning_controller.go
```diff
@@ -142,6 +138,34 @@ func (r *ProvisioningReconciler) Reconcile(req ctrl.Request) (ctrl.Result, error
...
+		}
+	} else {
+		// The Provisioning object is being deleted
+		if containsString(baremetalConfig.ObjectMeta.Finalizers, metal3iov1alpha1.ProvisioningFinalizer) {
+			// Add any specific deletion logic here
+
+			// Remove our finalizer from the list and update it.
+			baremetalConfig.ObjectMeta.Finalizers = removeString(baremetalConfig.ObjectMeta.Finalizers,
+				metal3iov1alpha1.ProvisioningFinalizer)
+			if err := r.Client.Update(context.Background(), baremetalConfig); err != nil {
```
dhellmann, line 163: As soon as this update is applied the API is going to delete the resource. Is that what we want here? Or is there some other logic we want to apply first?

sadasu: If we had additional logic to apply, this is the place to do it as you suggest. We may want to delete the metal3 deployment when the provisioning CR is deleted but it is not necessary to specific that now. 

Yes, removing the finalizer annotation lets OpenShift know that this resource and some internal data-structures it maintains for this resource can all be garbage collected.

dhellmann: OK. Normally the reason for adding a finalizer is to give the controller a chance to do something before the resource is deleted. The way this is implemented, we will be notified, but we won't actually make the API server wait to delete the resource until we've done what we wanted. In the other controllers I've seen, there is at least a check to decide if the controller is ready to have the finalizer removed. What sort of similar check could we apply here?

sadasu:
@asalkeld I needed to update 0000_31_cluster-baremetal-operator_05_rbac.yaml manually for this PR and generate-check doesn't like it. How can I fix that?

manifests/0000_31_cluster-baremetal-operator_05_rbac.yaml
```diff
@@ -71,13 +71,24 @@ metadata:
   creationTimestamp: null
   name: cluster-baremetal-operator
 rules:
-- resources:
-  - events
+- apiGroups:
```
dhellmann, line 74: I expect all of these changes need to be made in files under `config/rbac` instead of directly to this generated file.

sadasu: I am not sure if that is the complete picture because "deployments" are not specified in any yaml files in config/rbac but is present in manifests/0000_31_cluster-baremetal-operator_05_rbac.yaml.

dhellmann: Isn't that in config/rbac/role.yaml?

sadasu: I hadn't realized https://github.com/openshift/cluster-baremetal-operator/pull/52 had merged overnight. That explains all that I was seeing.

controllers/clusteroperator.go
asalkeld: I don't see anything unique in this function, can you just use updateCOStatus(). Note that updateCOStatus() will also need to handle the cases you talk about (no StatusConditions, Version or RelatedObjects)

controllers/provisioning_controller.go
```diff
@@ -250,3 +277,24 @@ func (r *ProvisioningReconciler) SetupWithManager(mgr ctrl.Manager) error {
 		Owns(&osconfigv1.ClusterOperator{}).
 		Complete(r)
 }
+
```
asalkeld, line 280: there are libraries that do this try https://github.com/thoas/go-funk
```
funk.Contains()
```

controllers/provisioning_controller.go
```diff
@@ -250,3 +277,24 @@ func (r *ProvisioningReconciler) SetupWithManager(mgr ctrl.Manager) error {
...
+func containsString(slice []string, s string) bool {
+	for _, item := range slice {
+		if item == s {
+			return true
+		}
+	}
+	return false
+}
+
+// Helper function to remove string from a slice of strings
```
asalkeld, line 291: see Prune() in the lib above

main.go
```diff
@@ -100,6 +100,21 @@ func main() {
...
 	osClient := osclientset.NewForConfigOrDie(rest.AddUserAgent(config, controllers.ComponentName))
+	// Check the Platform Type to determine the state of the CO
+	enabled, err := controllers.IsEnabled(osClient)
+	if err != nil {
+		setupLog.Error(err, "could not determine whether to run")
+		os.Exit(1)
+	}
+	if !enabled {
+		//Set ClusterOperator status to disabled=true, available=true
+		err = controllers.SetCOInDisabledState(osClient, releaseVersion)
```
asalkeld, line 111: I think you should move the content of SetCOInDisabledState() into updateCOStatus() - if there is anything missing
```suggestion
		err = controllers.updateCOStatus(...)
```

asalkeld: also, you could do this within SetupWithManager so that you have the reconcile object

controllers/provisioning_controller.go
asalkeld: this looks prime to be moved into a separate function

sadasu: @dhellmann agreed. And, we should be looking into adding new logic to probably delete a metal3 deployment when the provisioning resource is deleted. Deliberately not adding a lot of new functionality in here as part of this PR. Just making sure that the functionality that is already part of the CBO works in both metal3 and agnostic CIs.

controllers/clusteroperator.go
sadasu: We don't have the ProvisioningReconciler object at this time so added this new method. Went down the path of making  updateCOStatus() an exported function and ended up making a lot of changes in clusteroperator.go.

asalkeld: you don't have to do it this early. what i am suggesting is to do it within the SetupWithManager() function.

dhellmann:
The implementation looks OK as a first pass. I agree we can probably simplify, but since this works and we do have other work blocked on having *some* version of this, I think we should move ahead.

The commit history in the PR looks like it's going to be hard to follow later. There's a merge from master into the working branch and there are a bunch of separate RBAC commits. Could you rebase to get rid of the merge commit and squash the RBAC changes so they are 1 commit per type with the output of `make generate` included at each step?

PR #64, opened 2020-11-04
Bug 1893832: add missing ErrorCount field

andfasano:
The BareMetalHost CRD is missing the ErrorCount field, currently required by the BareMetal Operator


kirankt (approved):
/lgtm
/hold

Just adding a hold in case you want e2e-agnostic to pass, which is failing for unrelated reasons, I think.

kirankt:
/hold cancel
Just noticed that e2e-agnostic is a required job, so hold is not necessary. Sorry.

PR #65, opened 2020-11-05
Bug 1906105: Annotate existing metal3 deployment

sadasu:
- Create a new metal3 Deployment object and merge it with an existing one if available. The "merge" is done by ApplyDeployment(). As result of this, if the existing Deployment object was created by MAO (on upgrade from 4.6), the "cboOwnedAnnotation" would be added to the existing Deployment. If the Provisioning Config changed, then the DeploymentSpec would be updated with the DeploymentSpec of the new Deployment object that was created using the new config.

-  Use the state of the Deployment rollout to set the state in the "baremetal" ClusterOperator.

### DIFF
--- a/controllers/provisioning_controller.go
+++ b/controllers/provisioning_controller.go
@@ -259,6 +259,10 @@ func (r *ProvisioningReconciler) Reconcile(req ctrl.Request) (ctrl.Result, error
 	// Determine the status of the deployment
 	deploymentState, err := provisioning.GetDeploymentState(r.KubeClient.AppsV1(), ComponentNamespace, baremetalConfig)
 	if err != nil {
+		err = r.updateCOStatus(ReasonNotFound, "metal3 deployment inaccessible", "")
+		if err != nil {
+			return ctrl.Result{}, fmt.Errorf("unable to put %q ClusterOperator in Degraded state: %v", clusterOperatorName, err)
+		}
 		return ctrl.Result{}, errors.Wrap(err, "failed to determine state of metal3 deployment")
 	}
 	if deploymentState == appsv1.DeploymentReplicaFailure {
@@ -266,8 +270,7 @@ func (r *ProvisioningReconciler) Reconcile(req ctrl.Request) (ctrl.Result, error
 		if err != nil {
 			return ctrl.Result{}, fmt.Errorf("unable to put %q ClusterOperator in Degraded state: %v", clusterOperatorName, err)
 		}
-	}
-	if deploymentState == appsv1.DeploymentAvailable {
+	} else if deploymentState == appsv1.DeploymentAvailable {
 		err = r.updateCOStatus(ReasonSyncing, "metal3 pod running; starting other services", "")
 		if err != nil {
 			return ctrl.Result{}, fmt.Errorf("unable to put %q ClusterOperator in Progressing state: %v", clusterOperatorName, err)
@@ -299,6 +302,11 @@ func (r *ProvisioningReconciler) Reconcile(req ctrl.Request) (ctrl.Result, error
 	// Determine the status of the DaemonSet
 	daemonSetState, err := provisioning.GetDaemonSetState(r.KubeClient.AppsV1(), ComponentNamespace, baremetalConfig)
 	if err != nil {
+		err = r.updateCOStatus(ReasonNotFound, "metal3 image cache daemonset inaccessible", "")
+		if err != nil {
+			return ctrl.Result{}, fmt.Errorf("unable to put %q ClusterOperator in Degraded state: %v", clusterOperatorName, err)
+		}
+
 		return ctrl.Result{}, errors.Wrap(err, "failed to determine state of metal3 image cache daemonset")
 	}
 	if daemonSetState == provisioning.DaemonSetReplicaFailure {
@@ -306,9 +314,8 @@ func (r *ProvisioningReconciler) Reconcile(req ctrl.Request) (ctrl.Result, error
 		if err != nil {
 			return ctrl.Result{}, fmt.Errorf("unable to put %q ClusterOperator in Degraded state: %v", clusterOperatorName, err)
 		}
-	}
-	if daemonSetState == provisioning.DaemonSetAvailable {
-		err = r.updateCOStatus(ReasonSyncing, "metal3 pod and image cache are running", "")
+	} else if daemonSetState == provisioning.DaemonSetAvailable {
+		err = r.updateCOStatus(ReasonComplete, "metal3 pod and image cache are running", "")
 		if err != nil {
 			return ctrl.Result{}, fmt.Errorf("unable to put %q ClusterOperator in Progressing state: %v", clusterOperatorName, err)
 		}
--- a/controllers/provisioning_controller.go
+++ b/controllers/provisioning_controller.go
@@ -271,7 +271,7 @@ func (r *ProvisioningReconciler) Reconcile(req ctrl.Request) (ctrl.Result, error
 			return ctrl.Result{}, fmt.Errorf("unable to put %q ClusterOperator in Degraded state: %v", clusterOperatorName, err)
 		}
 	} else if deploymentState == appsv1.DeploymentAvailable {
-		err = r.updateCOStatus(ReasonSyncing, "metal3 pod running; starting other services", "")
+		err = r.updateCOStatus(ReasonSyncing, "metal3 pod running", "starting other metal3 services")
 		if err != nil {
 			return ctrl.Result{}, fmt.Errorf("unable to put %q ClusterOperator in Progressing state: %v", clusterOperatorName, err)
 		}
--- a/controllers/provisioning_controller.go
+++ b/controllers/provisioning_controller.go
@@ -256,6 +256,24 @@ func (r *ProvisioningReconciler) Reconcile(req ctrl.Request) (ctrl.Result, error
 		return ctrl.Result{Requeue: true}, err
 	}
 
+	// Determine the status of the deployment
+	deploymentState, err := provisioning.GetDeploymentState(r.KubeClient.AppsV1(), ComponentNamespace, baremetalConfig)
+	if err != nil {
+		return ctrl.Result{}, errors.Wrap(err, "failed to determine state of metal3 deployment")
+	}
+	if deploymentState == appsv1.DeploymentReplicaFailure {
+		err = r.updateCOStatus(ReasonDeployTimedOut, "metal3 deployment rollout taking too long", "")
+		if err != nil {
+			return ctrl.Result{}, fmt.Errorf("unable to put %q ClusterOperator in Degraded state: %v", clusterOperatorName, err)
+		}
+	}
+	if deploymentState == appsv1.DeploymentAvailable {
+		err = r.updateCOStatus(ReasonSyncing, "metal3 pod running; starting other services", "")
+		if err != nil {
+			return ctrl.Result{}, fmt.Errorf("unable to put %q ClusterOperator in Progressing state: %v", clusterOperatorName, err)
+		}
+	}
+
 	for _, ensureResource := range []ensureFunc{
 		provisioning.EnsureMetal3StateService,
 		provisioning.EnsureImageCache,
@@ -278,19 +296,19 @@ func (r *ProvisioningReconciler) Reconcile(req ctrl.Request) (ctrl.Result, error
 		}
 	}
 
-	// Determine the status of the deployment
-	deploymentState, err := provisioning.GetDeploymentState(r.KubeClient.AppsV1(), ComponentNamespace, baremetalConfig)
+	// Determine the status of the DaemonSet
+	daemonSetState, err := provisioning.GetDaemonSetState(r.KubeClient.AppsV1(), ComponentNamespace, baremetalConfig)
 	if err != nil {
-		return ctrl.Result{}, errors.Wrap(err, "failed to determine state of metal3 deployment")
+		return ctrl.Result{}, errors.Wrap(err, "failed to determine state of metal3 image cache daemonset")
 	}
-	if deploymentState == appsv1.DeploymentReplicaFailure {
-		err = r.updateCOStatus(ReasonDeployTimedOut, "metal3 deployment rollout taking too long", "")
+	if daemonSetState == provisioning.DaemonSetReplicaFailure {
+		err = r.updateCOStatus(ReasonDeployTimedOut, "metal3 image cache rollout taking too long", "")
 		if err != nil {
 			return ctrl.Result{}, fmt.Errorf("unable to put %q ClusterOperator in Degraded state: %v", clusterOperatorName, err)
 		}
 	}
-	if deploymentState == appsv1.DeploymentAvailable {
-		err = r.updateCOStatus(ReasonSyncing, "metal3 pod running; starting other services", "")
+	if daemonSetState == provisioning.DaemonSetAvailable {
+		err = r.updateCOStatus(ReasonSyncing, "metal3 pod and image cache are running", "")
 		if err != nil {
 			return ctrl.Result{}, fmt.Errorf("unable to put %q ClusterOperator in Progressing state: %v", clusterOperatorName, err)
 		}
--- a/controllers/provisioning_controller.go
+++ b/controllers/provisioning_controller.go
@@ -31,15 +31,12 @@ import (
 	"k8s.io/client-go/tools/record"
 	ctrl "sigs.k8s.io/controller-runtime"
 	"sigs.k8s.io/controller-runtime/pkg/client"
-	"sigs.k8s.io/controller-runtime/pkg/controller/controllerutil"
 
 	osconfigv1 "github.com/openshift/api/config/v1"
 	osclientset "github.com/openshift/client-go/config/clientset/versioned"
 	metal3iov1alpha1 "github.com/openshift/cluster-baremetal-operator/api/v1alpha1"
 	provisioning "github.com/openshift/cluster-baremetal-operator/provisioning"
 	"github.com/openshift/library-go/pkg/operator/events"
-	"github.com/openshift/library-go/pkg/operator/resource/resourceapply"
-	"github.com/openshift/library-go/pkg/operator/resource/resourcemerge"
 )
 
 const (
@@ -189,6 +186,14 @@ func (r *ProvisioningReconciler) Reconcile(req ctrl.Request) (ctrl.Result, error
 		return ctrl.Result{}, err
 	}
 
+	specChanged := baremetalConfig.Generation != baremetalConfig.Status.ObservedGeneration
+	if specChanged {
+		err = r.updateCOStatus(ReasonSyncing, "", "Applying metal3 resources")
+		if err != nil {
+			return ctrl.Result{}, fmt.Errorf("unable to put %q ClusterOperator in Syncing state: %v", clusterOperatorName, err)
+		}
+	}
+
 	if err := provisioning.ValidateBaremetalProvisioningConfig(baremetalConfig); err != nil {
 		// Provisioning configuration is not valid.
 		// Requeue request.
@@ -215,7 +220,7 @@ func (r *ProvisioningReconciler) Reconcile(req ctrl.Request) (ctrl.Result, error
 		return ctrl.Result{}, err
 	}
 
-	//Create Secrets needed for Metal3 deployment
+	// Create Secrets needed for Metal3 deployment
 	if err := provisioning.CreateMariadbPasswordSecret(r.KubeClient.CoreV1(), ComponentNamespace, baremetalConfig, r.Scheme); err != nil {
 		return ctrl.Result{}, errors.Wrap(err, "failed to create Mariadb password")
 	}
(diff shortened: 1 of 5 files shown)


controllers/provisioning_controller.go
```diff
@@ -223,48 +219,32 @@ func (r *ProvisioningReconciler) Reconcile(req ctrl.Request) (ctrl.Result, error
...
+	existing, err := provisioning.GetExistingMetal3Deployment(r.KubeClient.AppsV1(), ComponentNamespace)
 	if err != nil && !apierrors.IsNotFound(err) {
 		return ctrl.Result{}, errors.Wrap(err, "failed to find existing Metal3 Deployment")
 	}
 
-	if exists {
+	if existing != nil {
 		r.Log.V(1).Info("metal3 deployment already exists")
-		err = r.updateCOStatus(ReasonComplete, "found existing Metal3 deployment", "")
+		provisioning.AnnotateExistingMetal3Deployment(existing)
```
dhellmann, line 229: For right now we are assuming the deployment we would have created won't be any different from the one the machine-api-operator creates. That assumption will not hold for very long, though. I think we want to unify all of this logic to not treat the 2 cases quite so differently. Even if we find an existing deployment, we should make sure it matches the one we would have created, including the annotation. That will also let us reset the definition of the deployment if someone tries to edit it later, for example.

We should also say that we are Progressing until the Deployment reports that it is Available. That way we don't indicate success too soon, and leave an issue with the deployment undiscovered.

dhellmann (changes requested):

controllers/provisioning_controller.go
sadasu: Agree with point #1. I hadn't considered that while during implementation.

Regarding putting the CO in progressing state, not sure how I missed that!

controllers/provisioning_controller.go
```diff
@@ -222,49 +218,22 @@ func (r *ProvisioningReconciler) Reconcile(req ctrl.Request) (ctrl.Result, error
...
-	expectedGeneration := resourcemerge.ExpectedDeploymentGeneration(metal3Deployment, r.Generations)
 
-	err = controllerutil.SetControllerReference(baremetalConfig, metal3Deployment, r.Scheme)
-	if err != nil {
-		return ctrl.Result{}, fmt.Errorf("unable to set controllerReference on deployment: %v", err)
-	}
-
-	deployment, updated, err := resourceapply.ApplyDeployment(r.KubeClient.AppsV1(),
-		events.NewLoggingEventRecorder(ComponentName), metal3Deployment, expectedGeneration)
+	err = provisioning.ApplyNewOrUpdatedDeployment(r.KubeClient.AppsV1(), ComponentNamespace, metal3Deployment, r.Generations, ComponentName)
```
asalkeld, line 229:
```suggestion
	err = provisioning.ApplyDeployment(r.KubeClient.AppsV1(), ComponentNamespace, metal3Deployment, r.Generations, ComponentName)
```

sadasu: I was hoping that calling it ApplyNewOrUpdatedDeployment() will make it clear that sometimes we are just updating the deployment object with only portions that have changed since the previous reconcile run.

provisioning/baremetal_pod.go
```diff
@@ -468,10 +475,53 @@ func NewMetal3Deployment(targetNamespace string, images *Images, config *metal3i
...
+	existing.SetAnnotations(cboAnnotations)
+	return
+}
+
+func ApplyNewOrUpdatedDeployment(client appsclientv1.DeploymentsGetter, targetNamespace string, metal3Deployment *appsv1.Deployment, generations []osoperatorv1.GenerationStatus, componentName string) error {
+
+	// Check if there is an existing metal3 deployment
+	existing, err := client.Deployments(targetNamespace).Get(context.Background(), BaremetalDeploymentName, metav1.GetOptions{})
+
 	if existing != nil && err == nil {
```
asalkeld, line 497: can we make the function not have so much duplicated code?

controllers/provisioning_controller.go
asalkeld: :man_shrugging: I thought that was what apply ment - not a biggie, just looks overly wordy

sadasu: I like the short name too. I was trying to be unnecessarily "helpful" I guess.

provisioning/baremetal_pod.go
sadasu: Done.

controllers/provisioning_controller.go
sadasu: Both the above comments are addressed now.

controllers/provisioning_controller.go
```diff
@@ -298,3 +282,57 @@ func removeString(slice []string, s string) (result []string) {
 	}
 	return
 }
+
+func (r *ProvisioningReconciler) waitForDeploymentRollout(deployment *appsv1.Deployment, pollInterval, rolloutTimeout time.Duration) error {
+	var lastError error
+	err := wait.Poll(pollInterval, rolloutTimeout, func() (bool, error) {
```
andfasano, line 288: IIUC the `wait.Poll` can block, in the worst case, the reconcile loop up to 5 minutes. In such case I'd suggest to adopt a non-blocking design for the reconcile loop, such as tracking the state accordingly and rescheduling the request, so that the operator could continue to work on a regular basis.

dhellmann: It looks like this implementation is lifted from the machine-api-operator. @sadasu did you end up changing any of it?

I agree that normally we would want the operator to support more parallelism, but we expect our API resource to be a singleton in this case, right? So maybe blocking is OK in this case, especially if it simplifies the operator logic.

It would be nice if we could put this function in another package, but we can clean that up in a separate PR.

provisioning/baremetal_pod.go
```diff
@@ -468,10 +475,54 @@ func NewMetal3Deployment(targetNamespace string, images *Images, config *metal3i
...
+
+func ApplyNewOrUpdatedDeployment(client appsclientv1.DeploymentsGetter, targetNamespace string, metal3Deployment *appsv1.Deployment, generations []osoperatorv1.GenerationStatus, componentName string) error {
+	finalDeployment := metal3Deployment
+
+	// Check if there is an existing metal3 deployment
+	existing, err := client.Deployments(targetNamespace).Get(context.Background(), BaremetalDeploymentName, metav1.GetOptions{})
+
 	if existing != nil && err == nil {
-		return true, nil
+		if equality.Semantic.DeepEqual(metal3Deployment.Spec, existing.Spec) {
```
dhellmann, line 499: Do we need the annotation to be handled as a separate case? Can we simplify this by building the structure for what we want the deployment to look like, including the annotation, and then always updating it? If there is no change, nothing will happen. If there are any changes, including the annotation, the deployment would be updated all at once.

controllers/provisioning_controller.go
sadasu: Yes, re-using MAO logic here. That is our baseline and we can improve upon it based on our current discussion. Having the wait.Poll() allows us to put the CO is Available state right after the metal3 pod becomes available. It is also really useful to figure out when to put the CO in degraded state. (We don't want to do that after the first failure.)

Yes, I could move it to another package.

If we were to remove the wait, we could maintain a deploymentElapsedCount/Time value and keep decrementing it every time the metal3 deployment is not available in each reconcile loop. Once deploymentElapsedCount/Time hits zero we mark CO as Degraded and success at any point would put the CI in Available state. Is that an alternate approach that seems reasonable?

provisioning/baremetal_pod.go
sadasu: That is exactly the semantics ApplyNewOrUpdatedDeployment() is providing. 

But, within the function we need to look at different parts of the Deployment object and update them independently to prep the finalDeployment object. Then at https://github.com/openshift/cluster-baremetal-operator/pull/65/files#diff-1575ce96065be1a97bee923445ae60115c8ce02b4a2736788012df8162407100R508, this finalDeployment is applied.

controllers/provisioning_controller.go
sadasu: Not to be confused with this ApplyDeployment https://github.com/openshift/cluster-baremetal-operator/pull/65/files#diff-1575ce96065be1a97bee923445ae60115c8ce02b4a2736788012df8162407100R508.

controllers/provisioning_controller.go
asalkeld: personally i think you should ditch the Poll and just do a once off check. if it is not "up" just return
```
return ctrl.Result{RequeueAfter: time.Minute}, nil
```
Then just let it run the reconcile again

dhellmann: In order to tell when to mark the CO as failing, it does seem like we need to know when we started to try to launch/update the deployment. Does the deployment (or pod?) resource provide enough status information for us to compute that time just by looking at it, or do we really have to track it ourselves? It would be nice if we could avoid reverse-engineering failure detection that the deployment controller should be able to give us.

sadasu: Updated logic around how we determine the state of the deployment rollout and then set the CO state accordingly. Current approach might still need some tweaking but looking for feedback regarding general direction.

controllers/provisioning_controller.go
```diff
@@ -222,49 +219,42 @@ func (r *ProvisioningReconciler) Reconcile(req ctrl.Request) (ctrl.Result, error
...
 	}
 
-	// Proceed with creating a new Metal3 deployment
+	// Create metal3 deployment object based on current baremetal configuration
 	metal3Deployment := provisioning.NewMetal3Deployment(ComponentNamespace, &containerImages, &baremetalConfig.Spec)
-	expectedGeneration := resourcemerge.ExpectedDeploymentGeneration(metal3Deployment, r.Generations)
 
-	err = controllerutil.SetControllerReference(baremetalConfig, metal3Deployment, r.Scheme)
+	// Apply new deployment or compare against existing deployment and update
+	err = provisioning.ApplyNewOrUpdatedDeployment(r.KubeClient.AppsV1(), ComponentNamespace, metal3Deployment, r.Generations, ComponentName)
```
andfasano, line 232: Will `ApplyNewOrUpdatedDeployment` overwrite the specs of an already existing deployment? If so, I'd recommend to keep a simpler logic, ie if the deployment does not exist let's create it, otherwise just annotate it

sadasu: ApplyNewOrUpdatedDeployment() needs to do one other thing. As part of allowing changes to Provisioning CR on day 2, an existing metal3 deployment has to adjust to that. That is why I have that logic around comparing the Spec fields of existing metal3 deployment and the new metal3Deployment object just created using the latest Provisioning CR.

andfasano: My main concern is about supporting two scenarios:
1) Already existing metal3 deployment (and that's the most common one, during the upgrade)
2) Manual emergency change to metal3 deployment (ie an environment variable of one pod's container)
In such cases I don't think CBO should overwrite the existing metal3 specs with new values

dhellmann: The whole point of this operator is to manage the metal3 deployment. It should definitely overwrite whatever is there, entirely, if it is different from what it thinks it should be. That's it's #1 job.

If a user needs to change something in the deployment, we need to give them an API for that. If they want to make manual changes, they should turn the cluster-baremetal-operator off and recognize that they're outside of the supported configuration for a cluster.

provisioning/baremetal_pod.go
dhellmann: Why can't the Deployment be computed all at once, then applied via Create() if it does not exist or Update() if it does exist?

sadasu: I am looking into how to best use of some Create() vs Update() logic already built into resourceapply.applyDeployment().

zaneb: It's my understanding that ApplyDeployment() already implements what you're trying to do here.

provisioning/baremetal_pod.go
```diff
@@ -468,10 +476,73 @@ func NewMetal3Deployment(targetNamespace string, images *Images, config *metal3i
...
+		}
+	}
+	expectedGeneration := resourcemerge.ExpectedDeploymentGeneration(finalDeployment, generations)
+	deployment, updated, err := resourceapply.ApplyDeployment(client,
+		events.NewLoggingEventRecorder(componentName), finalDeployment, expectedGeneration)
+	if err != nil {
+		return errors.Wrap(err, "failed to apply new metal3 Deployment")
+	}
+	if updated {
+		resourcemerge.SetDeploymentGeneration(&generations, deployment)
```
zaneb, line 515: generations is not passed to this function as a pointer, so any update here will be lost when this function returns.

controllers/provisioning_controller.go
andfasano: Ok, that's fine, I didn't want to remove the ownership to CBO or end up in a scenario without it. I was looking to how to support a scenario like https://github.com/metal3-io/metal3-docs/blob/master/design/baremetal-operator/limit-hosts-provisioning.md#implementation-detailsnotesconstraints, where a configuration variable is present but not exposed in the complete chain (ie from the installer up). Could make sense to have some custom fields in the Provisioning CR for that?

provisioning/baremetal_pod.go
```diff
@@ -468,10 +476,73 @@ func NewMetal3Deployment(targetNamespace string, images *Images, config *metal3i
...
+	existing.SetAnnotations(cboAnnotations)
+	return
+}
+
+func ApplyNewOrUpdatedDeployment(client appsclientv1.DeploymentsGetter, targetNamespace string, metal3Deployment *appsv1.Deployment, generations []osoperatorv1.GenerationStatus, componentName string) error {
+	finalDeployment := metal3Deployment
+
+	// Check if there is an existing metal3 deployment
+	existing, err := client.Deployments(targetNamespace).Get(context.Background(), BaremetalDeploymentName, metav1.GetOptions{})
+
```
asalkeld, line 498: fyi Zane is adding support for whether or not mao or cbo owns the deployment https://github.com/openshift/cluster-baremetal-operator/pull/72/files#diff-1575ce96065be1a97bee923445ae60115c8ce02b4a2736788012df8162407100R471

provisioning/baremetal_pod.go
```diff
@@ -468,10 +476,73 @@ func NewMetal3Deployment(targetNamespace string, images *Images, config *metal3i
...
 	if existing != nil && err == nil {
-		return true, nil
+		if equality.Semantic.DeepEqual(metal3Deployment.Spec, existing.Spec) {
+			// the deployment spec of the existing deployment already matches what is required
+			// So, may need to just update annotation and generation
+
+			annotateExistingMetal3Deployment(existing)
```
asalkeld, line 504: why do we need to annotate it? isn't setting the controllerRef ownership enough?

provisioning/baremetal_pod.go
```diff
@@ -468,10 +476,73 @@ func NewMetal3Deployment(targetNamespace string, images *Images, config *metal3i
...
+	if updated {
+		resourcemerge.SetDeploymentGeneration(&generations, deployment)
+	}
+
+	return nil
+}
+
+func getDeploymentCondition(deployment *appsv1.Deployment) appsv1.DeploymentConditionType {
+	for _, cond := range deployment.Status.Conditions {
+		if cond.Status == corev1.ConditionTrue {
```
asalkeld, line 523: this seems very weird, we just return the first cond type that is true?

asalkeld: o, I see a deployment only has Available and Progressing.. so all good.

provisioning/baremetal_pod.go
sadasu: Fixed this.

provisioning/baremetal_pod.go
sadasu: MAO only looks at the annotation and not at the controllerRef. MAO adds the maoOwnedAnnotation to all the resources it creates and will look for cboOwnedAnnotation to know that it should no longer manage the metal3 pod.
But, CBO should use controllerRef and in some future release, we can remove this cboOwnedAnnotation altogether.

provisioning/baremetal_pod.go
sadasu: Yes, I'll rebase this PR after #72 merges.

controllers/provisioning_controller.go
zaneb: I think the way to handle that would be to pass those environment variables we want to make configurable from an optional ConfigMap that the user can supply.

provisioning/baremetal_pod.go
```diff
@@ -516,3 +517,32 @@ func ApplyNewOrUpdatedDeployment(client appsclientv1.DeploymentsGetter, targetNa
...
+	return appsv1.DeploymentProgressing
+}
+
+// Provide the current state of metal3 deployment
+func GetDeploymentState(client appsclientv1.DeploymentsGetter, targetNamespace string, config *metal3iov1alpha1.Provisioning) (appsv1.DeploymentConditionType, error) {
+	existing, err := client.Deployments(targetNamespace).Get(context.Background(), BaremetalDeploymentName, metav1.GetOptions{})
+	if apierrors.IsNotFound(err) {
+		// Deployment is not available and if Provisioning CR is not available,
+		// then everything is fine
+		if config == nil {
```
zaneb, line 507: This can't happen, so I wouldn't worry about it.

provisioning/baremetal_pod.go
```diff
@@ -516,3 +517,32 @@ func ApplyNewOrUpdatedDeployment(client appsclientv1.DeploymentsGetter, targetNa
...
+			return appsv1.DeploymentAvailable, nil
+		} else {
+			// Provisioning CR is present so the deployment rollout could be in early stages
+			return appsv1.DeploymentProgressing, err
```
zaneb, line 511: The error pretty much overrides the condition type that we are returning, so it would seem less misleading to just return the default value.

controllers/provisioning_controller.go
```diff
@@ -227,13 +229,32 @@ func (r *ProvisioningReconciler) Reconcile(req ctrl.Request) (ctrl.Result, error
...
 	if err != nil {
-		return ctrl.Result{}, fmt.Errorf("unable to apply Metal3 deployment: %v", err)
+		return ctrl.Result{}, fmt.Errorf("unable to apply metal3 deployment: %v", err)
+	}
+
+	// Determine the status of the deployment
+	deploymentState, err := provisioning.GetDeploymentState(r.KubeClient.AppsV1(), ComponentNamespace, baremetalConfig)
+	if apierrors.IsNotFound(err) {
+		// Deployment progressing
+		return ctrl.Result{RequeueAfter: time.Minute}, nil
```
zaneb, line 242: We're the ones creating the Deployment, so if it's not there then someone has deleted it; I don't think there's any reason not to requeue right away so we can create it ASAP. In fact, I'd consider just returning the NotFound error, since we should have *just* created it.

provisioning/baremetal_pod.go
sadasu: It is possible that the provisioning CR was deleted and and as a response to that we deleted the metal3 deployment.

controllers/provisioning_controller.go
sadasu: I have simplified the error handling in this path. I believe it takes care of all the above comments regarding error handling.

controllers/provisioning_controller.go
```diff
@@ -222,49 +219,42 @@ func (r *ProvisioningReconciler) Reconcile(req ctrl.Request) (ctrl.Result, error
...
+	// Update status of CO to indicate that it is progressing towards a metal3 deployment
+	err = r.updateCOStatus(ReasonSyncing, "syncing metal3 pod", "")
 	if err != nil {
-		return ctrl.Result{}, fmt.Errorf("unable to put %q ClusterOperator in Syncing state: %v", clusterOperatorName, err)
+		return ctrl.Result{}, fmt.Errorf("unable to put %q ClusterOperator in Progressing state: %v", clusterOperatorName, err)
 	}
 
-	// Proceed with creating a new Metal3 deployment
+	// Create metal3 deployment object based on current baremetal configuration
 	metal3Deployment := provisioning.NewMetal3Deployment(ComponentNamespace, &containerImages, &baremetalConfig.Spec)
```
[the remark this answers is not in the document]
sadasu: Method is called EnsureDeployment() and a new Deployment is created within that method. So, this has been addresses.

provisioning/baremetal_pod.go
sadasu: Not specifically adding just the annotation anymore. We create a new Deployment object with the cboOwnedAnnotation and representing the current state of Provisioning CR and if a Deployment is found without it, resourceapply.ApplyDeployment() will add to an existing Deployment.

provisioning/baremetal_pod.go
sadasu: Agreed. Fixed logic around this.

provisioning/baremetal_pod.go
sadasu: Again, updated logic around this. Still, continuing to return the DeploymentCondition and error though.

sadasu:
/retitle Bug 1906105: Annotate existing metal3 deployment

PR #66, opened 2020-11-06
Add provisioning configuration validations

imain:
Add configuration validations for each operational mode as well as unit tests to verify them.


provisioning/baremetal_config.go
```diff
@@ -75,56 +78,129 @@ func getProvisioningNetworkMode(prov *metal3iov1alpha1.Provisioning) metal3iov1a
...
-		{Name: "ProvisioningInterface", Value: prov.Spec.ProvisioningInterface},
-		{Name: "ProvisioningIP", Value: prov.Spec.ProvisioningIP},
-		{Name: "ProvisioningNetworkCIDR", Value: prov.Spec.ProvisioningNetworkCIDR},
-		{Name: "ProvisioningDHCPRange", Value: prov.Spec.ProvisioningDHCPRange},
-		{Name: "ProvisioningOSDownloadURL", Value: prov.Spec.ProvisioningOSDownloadURL},
-	} {
-		if toTest.Value == "" {
-			return fmt.Errorf("%s is required but is empty", toTest.Name)
-		}
+	provisionConfig := map[string]string{
```
dhellmann, line 81: What is the benefit of copying data out of well-defined types into maps like this?

provisioning/baremetal_config.go
```diff
@@ -75,56 +78,129 @@ func getProvisioningNetworkMode(prov *metal3iov1alpha1.Provisioning) metal3iov1a
...
+}
+
+func validateProvisioningOSDownloadURL(uri string) error {
+	parsedURL, err := url.ParseRequestURI(uri)
+	if err != nil {
+		return fmt.Errorf("the URI provided: %s is invalid", uri)
+	}
+	if parsedURL.Scheme == "http" || parsedURL.Scheme == "https" {
+		var sha256Checksum string
+		if sha256Checksums, ok := parsedURL.Query()["sha256"]; ok {
```
dhellmann, line 120: Is sha256 the only checksum type we support?

provisioning/baremetal_config.go
```diff
@@ -75,56 +78,129 @@ func getProvisioningNetworkMode(prov *metal3iov1alpha1.Provisioning) metal3iov1a
...
-		Value string
-	}{
-
-		{Name: "ProvisioningIP", Value: prov.Spec.ProvisioningIP},
-		{Name: "ProvisioningNetworkCIDR", Value: prov.Spec.ProvisioningNetworkCIDR},
-		{Name: "ProvisioningOSDownloadURL", Value: prov.Spec.ProvisioningOSDownloadURL},
-	} {
-		if toTest.Value == "" {
-			return fmt.Errorf("%s is required but is empty", toTest.Name)
+func validateProvisioningConfig(pc map[string]string) error {
```
dhellmann, line 138: Is there a way for us to build a list of all of the errors and return them all?

provisioning/baremetal_config.go
imain: Thanks for the quick review Doug!  Well there are a few options but I think this one produces the least amount of code.  Basically this just informs us what variables we are interested in testing.  You could use another method to do this but if you want to do it all in one function, as I have done here, you need a way to know what variables are being checked.

The reason this helps is because provisioning IP, DHCP range and network CIDR are all checked against each other.  So we need to know which are important and need to be checked and which can be ignored.  Otherwise we'd have to split them all up and somehow demarcate the involved variables.

This just seemed to me to be the best way to approach this.  I'm open to other ideas.

provisioning/baremetal_config.go
imain: This is what the installer expects and requests.  I do suppose we could support other hash types here since the user could request them.. thoughts?

provisioning/baremetal_config.go
imain: yeah that's a good idea.  I'd have to check if webhooks is cool with a list but otherwise it'd be fine. I guess we could also just return a longer error string if not.

provisioning/baremetal_config.go
imain: I should say the current version allows us to test all states with a single function just by changing what keys are set in the map.

dhellmann: Couldn't we also achieve that by looking directly at the Provisioning resource in the validation functions?

provisioning/baremetal_config.go
dhellmann: OK, if that's all we support for now then this is fine.

provisioning/baremetal_config.go
dhellmann: Yeah, if we return a list the caller can do what it wants with the list, including formatting a response payload as a single string if that's all it supports.

provisioning/baremetal_config.go
imain: I guess if we did some validation in each of validateManagedConfig(), validateUnmanagedConfig() and validateDisabledConfig to ensure that each only sets the variables we need then we could do it.  If a user wanted to move to Disabled config for example but left the DHCPRange set.. I'm not sure how we'd deal with that.  Error out?  Set it to "" for them?

provisioning/baremetal_config.go
imain: sounds good

provisioning/baremetal_config.go
dhellmann: Why do we care if they leave extra data defined? Why not just apply the rules based on the mode that is set?

imain: then we have to know the mode.. I guess that's another way to do it.  Pass in the mode to the validations.

dhellmann: Would it make sense to have a separate validation function for each mode, dispatched from a top level function? How different are the validations in each mode?

imain: There is a large amount of overlap.  The largest has 5 fields to validate while the smallest has a subset of 3 of those fields.  And again, they are interdependent which is why I took this approach.  It would be a mess to do them all separately.

imain: So far I see this approach, or passing in the mode as the best options.

dhellmann: Isn't the mode part of the data in the Provisioning resource, though? Or at least derivable from it? Why do we need to pass it as an argument?

dhellmann: Expanding the collapsed part of the diff, I see that we do have a single top level function dispatching to mode-specific functions at this level, which then put some of the configuration values into a map and passes that to another single function that is named generically but actually performs some mode-specific checks based on which values are found in the map. So, maybe passing the mode into validateProvisioningConfig makes sense.

Alternatively, creating a bunch of little functions for validating the parts of the config and then invoking each of those from these mode-specific functions would make sense and give clear, testable, and reusable code, without obfuscating the mode-specific behavior behind a map.

imain:
@dhellmann Well, that's it split apart based on the mode.  That's the best I could think of.  Let me know if you have other ideas.  I still like the map better :).  I didn't do the error list return yet as I'm not sure how best to do it.  Do you think we use the validation/field packages or just return a list of error strings?

provisioning/baremetal_config.go
```diff
@@ -48,13 +51,55 @@ var (
...
+	// They all use the provisioning network settings, except DHCPRange.  We'll
+	// Check that below.
+	if err := validateProvisioningNetworkSettings(prov.Spec.ProvisioningIP, prov.Spec.ProvisioningNetworkCIDR, prov.Spec.ProvisioningDHCPRange); err != nil {
+		return err
+	}
+
 	switch provisioningNetworkMode {
 	case metal3iov1alpha1.ProvisioningNetworkManaged:
-		return validateManagedConfig(prov)
+		if err := validateProvisioningInterface(prov.Spec.ProvisioningInterface); err != nil {
```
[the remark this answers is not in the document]
imain: ah yes good catch

provisioning/baremetal_config.go
```diff
@@ -48,13 +51,55 @@ var (
...
+
 	case metal3iov1alpha1.ProvisioningNetworkUnmanaged:
-		return validateUnmanagedConfig(prov)
+		if err := validateProvisioningInterface(prov.Spec.ProvisioningInterface); err != nil {
+			return err
+		}
+		if prov.Spec.ProvisioningDHCPRange == "" {
+			return fmt.Errorf("ProvisioningDHCPRange is required in this mode but is not set")
+		}
 	case metal3iov1alpha1.ProvisioningNetworkDisabled:
```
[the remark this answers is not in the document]
imain: nice

provisioning/baremetal_config.go
```diff
@@ -74,57 +119,93 @@ func getProvisioningNetworkMode(prov *metal3iov1alpha1.Provisioning) metal3iov1a
...
+		return fmt.Errorf("%q is not a valid provisioningNetworkCIDR: %s", cidr, err)
+	}
+
+	// Ensure provisioning IP is in the network CIDR
+	if !provisioningCIDR.Contains(provisioningIP) {
+		return fmt.Errorf("%q is not part of the provisioning network CIDR", ip)
+	}
+
+	// Verify DHCP Range if required
+	if dhcpRange != "" {
```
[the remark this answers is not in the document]
imain: Hmm this one is different in that dhcpRange is optional here.  If it's set to "" we don't test it.  Maybe we can still do something though, I'll think about it.

provisioning/baremetal_config.go
```diff
@@ -74,57 +119,93 @@ func getProvisioningNetworkMode(prov *metal3iov1alpha1.Provisioning) metal3iov1a
...
+			if !provisioningCIDR.Contains(dhcpIP) {
+				return fmt.Errorf("%q is not part of the provisioning network", dhcpIP)
+			}
+		}
+
+		// Ensure provisioning IP is not in the DHCP range
+		start := net.ParseIP(dhcpRangeSplit[0])
+		end := net.ParseIP(dhcpRangeSplit[1])
+
+		if start != nil && end != nil {
```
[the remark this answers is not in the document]
imain: In theory this should never happen because we've verified the translation to IPs above.  It's just done in a for loop there to perform 2 checks on each ip.

provisioning/baremetal_config.go
```diff
@@ -74,57 +119,93 @@ func getProvisioningNetworkMode(prov *metal3iov1alpha1.Provisioning) metal3iov1a
...
+		if sha256Checksums, ok := parsedURL.Query()["sha256"]; ok {
+			sha256Checksum = sha256Checksums[0]
+		}
+		if sha256Checksum == "" {
+			return fmt.Errorf("the sha256 parameter in the %s URI is missing", uri)
+		}
+		if len(sha256Checksum) != 64 {
+			return fmt.Errorf("the sha256 parameter in the %s URI is invalid", uri)
 		}
+	} else {
```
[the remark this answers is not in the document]
imain: oh yeah, good thinking.

provisioning/baremetal_config.go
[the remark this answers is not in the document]
imain: ya got it

provisioning/baremetal_config_test.go
```diff
@@ -44,7 +44,7 @@ func TestValidateManagedProvisioningConfig(t *testing.T) {
 		spec          metal3iov1alpha1.ProvisioningSpec
 		expectedError bool
 		expectedMode  metal3iov1alpha1.ProvisioningNetwork
-		expectedMsg   string
+		expectedMsg   []string
```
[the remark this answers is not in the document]
imain: And by that I mean using a list of expected return strings.  It lets us test more than one error condition at a time.. as well as testing the multiple error return itself.

controllers/provisioning_controller.go
```diff
@@ -185,11 +185,15 @@ func (r *ProvisioningReconciler) Reconcile(req ctrl.Request) (ctrl.Result, error
 		return ctrl.Result{}, err
 	}
 
-	if err := provisioning.ValidateBaremetalProvisioningConfig(baremetalConfig); err != nil {
+	if errs := provisioning.ValidateBaremetalProvisioningConfig(baremetalConfig); len(errs) > 0 {
 		// Provisioning configuration is not valid.
 		// Requeue request.
 		r.Log.Error(err, "invalid config in Provisioning CR")
```
asalkeld, line 191: r.Log.Error(errs, "invalid config in Provisioning CR")

imain: oh yeah, good catch

controllers/clusteroperator_test.go
```diff
@@ -281,12 +281,12 @@ func TestUpdateCOStatusDegraded(t *testing.T) {
...
 				ProvisioningIP:            "172.30.20.3",
 				ProvisioningNetworkCIDR:   "172.30.20.0/24",
-				ProvisioningDHCPRange:     "172.30.20.11, 172.30.20.101",
+				ProvisioningDHCPRange:     "172.30.20.11,172.30.20.101",
 				ProvisioningOSDownloadURL: "",
 				ProvisioningNetwork:       "Managed",
 			},
 			expectedConditions: []osconfigv1.ClusterOperatorStatusCondition{
-				setStatusCondition(osconfigv1.OperatorDegraded, osconfigv1.ConditionTrue, "InvalidConfiguration", "ProvisioningOSDownloadURL is required but is empty"),
+				setStatusCondition(osconfigv1.OperatorDegraded, osconfigv1.ConditionTrue, "InvalidConfiguration", "provisioningOSDownloadURL is required but is empty\n"),
```
dhellmann, line 289: Can we avoid newlines at the end of the error messages?

provisioning/baremetal_config.go
```diff
@@ -74,58 +121,121 @@ func getProvisioningNetworkMode(prov *metal3iov1alpha1.Provisioning) metal3iov1a
...
-	return nil
+
+	parsedURL, err := url.ParseRequestURI(uri)
+	if err != nil {
+		errs = append(errs, fmt.Errorf("the provisioningOSDownloadURL provided: %q is invalid", uri))
+		// If it's not a valid URI lets just return.
+		return errs
+	}
+	if parsedURL.Scheme != "http" && parsedURL.Scheme != "https" {
+		errs = append(errs, fmt.Errorf("the provisioningOSDownloadURL provided: %s must begin with http/https", uri))
```
dhellmann, line 139:
```suggestion
		errs = append(errs, fmt.Errorf("unsupported scheme %q in provisioningOSDownloadURL %s", parsedURL.Scheme, uri))
```

provisioning/baremetal_config.go
```diff
@@ -74,58 +121,121 @@ func getProvisioningNetworkMode(prov *metal3iov1alpha1.Provisioning) metal3iov1a
...
+
+	if uri == "" {
+		errs = append(errs, fmt.Errorf("provisioningOSDownloadURL is required but is empty"))
+		return errs
 	}
-	return nil
+
+	parsedURL, err := url.ParseRequestURI(uri)
+	if err != nil {
+		errs = append(errs, fmt.Errorf("the provisioningOSDownloadURL provided: %q is invalid", uri))
```
dhellmann, line 134:
```suggestion
		errs = append(errs, errors.Wrap(err, fmt.Sprintf("could not parse provisioningOSDownloadURL %q", uri)))
```

provisioning/baremetal_config.go
```diff
@@ -74,58 +121,121 @@ func getProvisioningNetworkMode(prov *metal3iov1alpha1.Provisioning) metal3iov1a
...
-			return fmt.Errorf("%s is required but is empty", toTest.Name)
+func validateProvisioningNetworkSettings(ip string, cidr string, dhcpRange string) []error {
+	// provisioningIP and networkCIDR are always set.  DHCP range is optional
+	// depending on mode.
+	var errs []error
+
+	// Verify provisioning ip and get it into net format for future tests.
+	provisioningIP := net.ParseIP(ip)
+	if provisioningIP == nil {
+		errs = append(errs, fmt.Errorf("The provisioningIP %q is not a valid IP", ip))
```
dhellmann, line 180:
```suggestion
		errs = append(errs, fmt.Errorf("could not parse provisioningIP %q", ip))
```

provisioning/baremetal_config.go
```diff
@@ -74,58 +121,121 @@ func getProvisioningNetworkMode(prov *metal3iov1alpha1.Provisioning) metal3iov1a
...
+		// I think we return here.  There are a few more tests below that require the IP be set.
```
dhellmann, line 181: Is this a comment from when the PR was WIP?

provisioning/baremetal_config.go
```diff
@@ -74,58 +121,121 @@ func getProvisioningNetworkMode(prov *metal3iov1alpha1.Provisioning) metal3iov1a
...
+		return errs
+	}
+
+	// Verify Network CIDR
+	_, provisioningCIDR, err := net.ParseCIDR(cidr)
+	if err != nil {
+		errs = append(errs, fmt.Errorf("%q is not a valid provisioningNetworkCIDR: %s", cidr, err))
```
dhellmann, line 188:
```suggestion
		errs = append(errs, errors.Wrap(err, fmt.Sprintf("could not parse provisioningNetworkCIDR %q", cidr)))
```

provisioning/baremetal_config.go
```diff
@@ -74,58 +121,121 @@ func getProvisioningNetworkMode(prov *metal3iov1alpha1.Provisioning) metal3iov1a
...
+		// Similar thing.. need this to be valid for further tests.
+		return errs
+	}
+
+	// Ensure provisioning IP is in the network CIDR
+	if !provisioningCIDR.Contains(provisioningIP) {
+		errs = append(errs, fmt.Errorf("The provisioningIP %q is not part of the provisioningNetworkCIDR %q", ip, cidr))
```
dhellmann, line 195:
```suggestion
		errs = append(errs, fmt.Errorf("provisioningIP %q is in the range defined by provisioningNetworkCIDR %q", ip, cidr))
```

provisioning/baremetal_config.go
```diff
@@ -74,58 +121,121 @@ func getProvisioningNetworkMode(prov *metal3iov1alpha1.Provisioning) metal3iov1a
...
+	}
+
+	// DHCP Range might not be set in which case we're done here.
+	if dhcpRange == "" {
+		return errs
+	}
+
+	// We want to allow a space after the ',' if the user likes it.
+	dhcpRange = strings.ReplaceAll(dhcpRange, " ", "")
```
dhellmann, line 204: This will remove spaces that appear in other places in the string, so that a value like `192.168  .20.1,192.168. 20.100` is treated as valid. Should it be?

provisioning/baremetal_config.go
```diff
@@ -74,58 +121,121 @@ func getProvisioningNetworkMode(prov *metal3iov1alpha1.Provisioning) metal3iov1a
...
+	if len(dhcpRangeSplit) != 2 {
+		errs = append(errs, fmt.Errorf("%q is not a valid provisioningDHCPRange.  DHCP range format: start_ip,end_ip", dhcpRange))
+		return errs
+	}
+
+	for _, ip := range dhcpRangeSplit {
+		// Ensure IP is valid
+		dhcpIP := net.ParseIP(ip)
+		if dhcpIP == nil {
+			errs = append(errs, fmt.Errorf("The entry %q in the provisioningDHCPRange is not a valid IP", ip))
```
dhellmann, line 217:
```suggestion
			errs = append(errs, fmt.Errorf("could not parse provisioningDHCPRange, %q is not a valid IP", ip))
```

provisioning/baremetal_config.go
```diff
@@ -74,58 +121,121 @@ func getProvisioningNetworkMode(prov *metal3iov1alpha1.Provisioning) metal3iov1a
...
+			// Can't really do further tests without valid IPs
+			return errs
+		}
+
+		// Validate IP is in the provisioning network
+		if !provisioningCIDR.Contains(dhcpIP) {
+			errs = append(errs, fmt.Errorf("The provisioningDHCPRange IP %q is not part of the provisioningNetworkCIDR", dhcpIP))
```
dhellmann, line 224:
```suggestion
			errs = append(errs, fmt.Errorf("invalid provisioningDHCPRange, IP %q is not part of the provisioningNetworkCIDR %q", dhcpIP, cidr))
```

provisioning/baremetal_config.go
```diff
@@ -74,58 +121,121 @@ func getProvisioningNetworkMode(prov *metal3iov1alpha1.Provisioning) metal3iov1a
...
 	}
-	return nil
+
+	// Ensure provisioning IP is not in the DHCP range
+	start := net.ParseIP(dhcpRangeSplit[0])
+	end := net.ParseIP(dhcpRangeSplit[1])
+
+	if start != nil && end != nil {
+		if bytes.Compare(provisioningIP, start) >= 0 && bytes.Compare(provisioningIP, end) <= 0 {
+			errs = append(errs, fmt.Errorf("The provisioningIP %q overlaps with the provisioningDHCPRange %q", provisioningIP, dhcpRange))
```
dhellmann, line 234:
```suggestion
			errs = append(errs, fmt.Errorf("invalid provisioningIP %q, value must be outside of the provisioningDHCPRange %q", provisioningIP, dhcpRange))
```

provisioning/baremetal_config_test.go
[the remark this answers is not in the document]
dhellmann: I think I would rather see each unit test spelled out with a single error message, unless there's a case where that is not possible.

dhellmann (commented):
I think the logic is OK, except for one question inline about the whitespace handling in the DHCP range value. I left a few suggestions for rewording the errors.

provisioning/baremetal_config_test.go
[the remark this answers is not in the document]
imain: OK, back to single string errors. :)

controllers/clusteroperator_test.go
imain: Yep.

provisioning/baremetal_config.go
imain: It's still valid but I'll remove it.

provisioning/baremetal_config.go
imain: True.  Fixed.

controllers/provisioning_controller.go
```diff
@@ -185,11 +186,17 @@ func (r *ProvisioningReconciler) Reconcile(req ctrl.Request) (ctrl.Result, error
...
 		// Requeue request.
-		r.Log.Error(err, "invalid config in Provisioning CR")
-		err = r.updateCOStatus(ReasonInvalidConfiguration, err.Error(), "Unable to apply Provisioning CR: invalid configuration")
+		var errorstring string
+		for _, err := range errs {
+			errorstring = err.Error() + "\n" + errorstring
+		}
+		errorstring = strings.TrimSuffix(errorstring, "\n")
+
+		r.Log.Error(errors.New(errorstring), "invalid config in Provisioning CR")
```
dhellmann, line 198: This is going to produce a multi-line log message. What do you think about joining the strings on `"; "` instead of newlines?

provisioning/baremetal_config.go
imain: I assume you want "not in the range defined by.."

controllers/provisioning_controller.go
imain: Yeah either way.  Maybe that makes more sense.

imain: I changed it to "; "

imain:
So Stephens changes added a test for an unset provisioningIP in the disabled case.. which makes sense to me.  However according to the docs I had read it was required in disabled.  I need to double check this tomorrow.

imain:
Going to close this in favor of https://github.com/openshift/cluster-baremetal-operator/pull/84

imain:
Reopened in case we want to merge this.

stbenjam:
I've looked through the validations and I think they are all correct. Could we add some test cases for IPv6 values?

@sadasu What do you think about taking this before FF today?

stbenjam:
CI is busted, they're working on it so we can hold off retesting for a bit.

imain:
> I've looked through the validations and I think they are all correct. Could we add some test cases for IPv6 values?
> 
> @sadasu What do you think about taking this before FF

> I've looked through the validations and I think they are all correct. Could we add some test cases for IPv6 values?
> 
> @sadasu What do you think about taking this before FF today?

Thanks Stephen.  Yes I can add some IPv6 tests.

stbenjam:
The govet failure looks real

imain:
> The govet failure looks real

Yeah gofmt was doing some really strange things with that comment.  It would indent it more every time I ran it.  Hopefully this latest push is right.

imain:
oh so close on the CI!

stbenjam:
Let's see if virtualmedia passes, not sure why the first one failed.

One other question, now that I'm looking coser, is there any reason we didn't use the k8s apimachinery validations here?

It's meant to be used for validations like this on fields. We use them in the installer: https://github.com/openshift/installer/blob/master/pkg/types/baremetal/validation/platform.go#L120

imain:
> Let's see if virtualmedia passes, not sure why the first one failed.
> 
> One other question, now that I'm looking coser, is there any reason we didn't use the k8s apimachinery validations here?
> 
> It's meant to be used for validations like this on fields. We use them in the installer: https://github.com/openshift/installer/blob/master/pkg/types/baremetal/validation/platform.go#L120

Yeah Doug and I discussed it briefly.  We went with just a list of errors.  I'm open to either.  @dhellmann ?

imain:
> Let's see if virtualmedia passes, not sure why the first one failed.
> 
> One other question, now that I'm looking coser, is there any reason we didn't use the k8s apimachinery validations here?
> 
> It's meant to be used for validations like this on fields. We use them in the installer: https://github.com/openshift/installer/blob/master/pkg/types/baremetal/validation/platform.go#L120

The other idea is to do what Angus did and use "return errors.NewAggregate(errs)" to return a single comma separated error.

imain:
Closing this as it is included with Angus's changes.

PR #69, opened 2020-11-12
move logic for managing disabled state back inside the reconciler

dhellmann:
Avoid duplicating the logic for handling the disabled state by moving
the call to initialize the ClusterOperator from main into
SetupWithManager. That also lets us reuse the existing reconciler
method to set the ClusterOperator status, so we can eliminate
SetCOInDisabledState entirely.

This is a follow-up to #63

/cc @sadasu @asalkeld

### DIFF
--- a/controllers/clusteroperator.go
+++ b/controllers/clusteroperator.go
@@ -14,7 +14,6 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/controller/controllerutil"
 
 	osconfigv1 "github.com/openshift/api/config/v1"
-	osclientset "github.com/openshift/client-go/config/clientset/versioned"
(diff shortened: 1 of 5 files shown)


sadasu:
+1 for eliminating 2 separate ways to set the CO in Disabled state. Will let @asalkeld take a look also.

sadasu:
e2e-agnostic failure not related to this change.

/test e2e-agnostic

sadasu (approved):
/lgtm

asalkeld (commented):
/lgtm

thanks! sorry for the slow review

sadasu:
/hold
Will remove hold after CIs start passing again.

stbenjam:
We are well past the point where CBO is relevant for e2e-agnostic. I'm comfortable overriding this for now, but let's make sure we're looking at the CI logs before we do that on any others.

PR #71, opened 2020-11-19
Bug 1906935: Delete resources when Provisioning CR is deleted

sadasu:
Delete all Secrets and the metal3 Deployment object created by the CBO when the Provisioning CR is deleted.

### DIFF
--- a/controllers/provisioning_controller.go
+++ b/controllers/provisioning_controller.go
@@ -213,18 +213,9 @@ func (r *ProvisioningReconciler) Reconcile(req ctrl.Request) (ctrl.Result, error
 		return ctrl.Result{}, nil
 	}
 
-	// Create Secrets needed for Metal3 deployment
-	if err := provisioning.CreateMariadbPasswordSecret(r.KubeClient.CoreV1(), ComponentNamespace, baremetalConfig, r.Scheme); err != nil {
-		return ctrl.Result{}, errors.Wrap(err, "failed to create Mariadb password")
-	}
-	if err := provisioning.CreateIronicPasswordSecret(r.KubeClient.CoreV1(), ComponentNamespace, baremetalConfig, r.Scheme); err != nil {
-		return ctrl.Result{}, errors.Wrap(err, "failed to create Ironic password")
-	}
-	if err := provisioning.CreateIronicRpcPasswordSecret(r.KubeClient.CoreV1(), ComponentNamespace, baremetalConfig, r.Scheme); err != nil {
-		return ctrl.Result{}, errors.Wrap(err, "failed to create Ironic rpc password")
-	}
-	if err := provisioning.CreateInspectorPasswordSecret(r.KubeClient.CoreV1(), ComponentNamespace, baremetalConfig, r.Scheme); err != nil {
-		return ctrl.Result{}, errors.Wrap(err, "failed to create Inspector password")
+	//Create Secrets needed for Metal3 deployment
+	if err := provisioning.CreateAllSecrets(r.KubeClient.CoreV1(), ComponentNamespace, baremetalConfig, r.Scheme); err != nil {
+		return ctrl.Result{}, err
 	}
 
 	// If Metal3 Deployment already exists and managed by MAO, do nothing.
@@ -365,7 +356,13 @@ func (r *ProvisioningReconciler) checkForCRDeletion(info *provisioning.Provision
 
 //Delete Secrets and the Metal3 Deployment objects
 func (r *ProvisioningReconciler) deleteMetal3Resources(info *provisioning.ProvisioningInfo) error {
-	provisioning.DeleteAllSecrets(info)
+	failedSecrets := provisioning.DeleteAllSecrets(info)
+	if len(failedSecrets) > 0 {
+		for _, secret := range failedSecrets {
+			r.Log.V(1).Info("failed to delete secret : ", "secret", secret)
+		}
+		r.Log.V(1).Info("proceeeding with deletion of other resources")
+	}
 
 	err := provisioning.DeleteMetal3Deployment(info)
 	if err != nil {
--- a/controllers/provisioning_controller.go
+++ b/controllers/provisioning_controller.go
@@ -126,6 +126,12 @@ func (r *ProvisioningReconciler) readProvisioningCR(namespacedName types.Namespa
 func (r *ProvisioningReconciler) Reconcile(req ctrl.Request) (ctrl.Result, error) {
 	//log := r.Log.WithValues("provisioning", req.NamespacedName)
 
+	// Make sure ClusterOperator exists
+	err := r.ensureClusterOperator(nil)
+	if err != nil {
+		return ctrl.Result{}, err
+	}
+
 	enabled, err := r.isEnabled()
 	if err != nil {
 		return ctrl.Result{}, errors.Wrap(err, "could not determine whether to run")
@@ -153,6 +159,13 @@ func (r *ProvisioningReconciler) Reconcile(req ctrl.Request) (ctrl.Result, error
 		return ctrl.Result{}, nil
 	}
 
+	// Provisioning CR is present
+	// Make sure ClusterOperator's ownership is updated
+	err = r.ensureClusterOperator(baremetalConfig)
+	if err != nil {
+		return ctrl.Result{}, err
+	}
+
 	// Read container images from Config Map
 	var containerImages provisioning.Images
 	if err := provisioning.GetContainerImages(&containerImages, r.ImagesFilename); err != nil {
@@ -168,12 +181,6 @@ func (r *ProvisioningReconciler) Reconcile(req ctrl.Request) (ctrl.Result, error
 	}
 
 	info := r.provisioningInfo(baremetalConfig, &containerImages)
-
-	// Make sure ClusterOperator exists and updated
-	err = r.ensureClusterOperator(baremetalConfig)
-	if err != nil {
-		return ctrl.Result{}, err
-	}
 
 	// Check if Provisioning Configuartion is being deleted
 	deleted, err := r.checkForCRDeletion(info)
--- a/controllers/provisioning_controller.go
+++ b/controllers/provisioning_controller.go
@@ -121,65 +121,6 @@ func (r *ProvisioningReconciler) readProvisioningCR(namespacedName types.Namespa
 	return instance, nil
 }
 
-//Ensure Finalizer is present on the Provisioning CR when not deleted and
-//delete resources and remove Finalizer when it is
-func (r *ProvisioningReconciler) checkForCRDeletion(baremetalConfig *metal3iov1alpha1.Provisioning) (bool, error) {
-	// examine DeletionTimestamp to determine if object is under deletion
-	if baremetalConfig.ObjectMeta.DeletionTimestamp.IsZero() {
-		// The object is not being deleted, so if it does not have our finalizer,
-		// then lets add the finalizer and update the object. This is equivalent
-		// to registering our finalizer.
-		if !containsString(baremetalConfig.ObjectMeta.Finalizers,
-			metal3iov1alpha1.ProvisioningFinalizer) {
-			// Add finalizer becasue it doesn't already exist
-			baremetalConfig.ObjectMeta.Finalizers = append(baremetalConfig.ObjectMeta.Finalizers,
-				metal3iov1alpha1.ProvisioningFinalizer)
-			if err := r.Client.Update(context.Background(), baremetalConfig); err != nil {
-				return false, errors.Wrap(err, "failed to update Provisioning CR with its finalizer")
-			}
-		}
-		return false, nil
-	} else {
-		// The Provisioning object is being deleted
-		if containsString(baremetalConfig.ObjectMeta.Finalizers, metal3iov1alpha1.ProvisioningFinalizer) {
-			err := r.deleteMetal3Resources()
-			if err != nil {
-				return false, errors.Wrap(err, "failed to delete metal3 pod")
-			}
-			// Remove our finalizer from the list and update it.
-			baremetalConfig.ObjectMeta.Finalizers = removeString(baremetalConfig.ObjectMeta.Finalizers,
-				metal3iov1alpha1.ProvisioningFinalizer)
-			if err = r.Client.Update(context.Background(), baremetalConfig); err != nil {
-				return true, errors.Wrap(err, "failed to remove finalizer from Provisioning CR")
-			}
-			return true, nil
-		}
-	}
-	return false, nil
-}
-
-//Delete Secrets and the Metal3 Deployment objects
-func (r *ProvisioningReconciler) deleteMetal3Resources() error {
-	err := provisioning.DeleteMariadbPasswordSecret(r.KubeClient.CoreV1(), ComponentNamespace)
-	if err != nil {
-		r.Log.V(1).Info("could not delete mariadb password Secret")
-	}
-	err = provisioning.DeleteIronicPasswordSecret(r.KubeClient.CoreV1(), ComponentNamespace)
-	if err != nil {
-		r.Log.V(1).Info("could not delete ironic password Secret")
-	}
-	err = provisioning.DeleteInspectorPasswordSecret(r.KubeClient.CoreV1(), ComponentNamespace)
-	if err != nil {
-		r.Log.V(1).Info("could not delete ironic inspector password Secret")
-	}
-	err = provisioning.DeleteIronicRpcPasswordSecret(r.KubeClient.CoreV1(), ComponentNamespace)
-	if err != nil {
-		r.Log.V(1).Info("could not delete ironic rpc password Secret")
-	}
-
-	return provisioning.DeleteMetal3Deployment(r.KubeClient.AppsV1(), ComponentNamespace)
-}
-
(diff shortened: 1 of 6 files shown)


sadasu:
@asalkeld I tried to incorporate comments provided as part of https://github.com/openshift/cluster-baremetal-operator/pull/63.

I tried using https://github.com/thoas/go-funk library based on comments https://github.com/openshift/cluster-baremetal-operator/pull/63#discussion_r521725979 and https://github.com/openshift/cluster-baremetal-operator/pull/63#discussion_r521726497.

I found funk.Contains() but not funk.Prune() although the goDoc does say that Prune() exists https://godoc.org/github.com/thoas/go-funk#Prune.

controllers/provisioning_controller.go
```diff
@@ -120,6 +120,63 @@ func (r *ProvisioningReconciler) readProvisioningCR(req ctrl.Request) (*metal3io
...
+	}
+	err = provisioning.DeleteIronicPasswordSecret(r.KubeClient.CoreV1(), ComponentNamespace)
+	if err != nil {
+		r.Log.V(1).Info("could not delete ironic password Secret")
+	}
+	err = provisioning.DeleteInspectorPasswordSecret(r.KubeClient.CoreV1(), ComponentNamespace)
+	if err != nil {
+		r.Log.V(1).Info("could not delete ironic inspector password Secret")
+	}
+	err = provisioning.DeleteMetal3Deployment(r.KubeClient.AppsV1(), ComponentNamespace)
```
asalkeld, line 172:
```suggestion
	return provisioning.DeleteMetal3Deployment(r.KubeClient.AppsV1(), ComponentNamespace)
```

controllers/provisioning_controller.go
```diff
@@ -120,6 +120,63 @@ func (r *ProvisioningReconciler) readProvisioningCR(req ctrl.Request) (*metal3io
...
+	if baremetalConfig.ObjectMeta.DeletionTimestamp.IsZero() {
+		// The object is not being deleted, so if it does not have our finalizer,
+		// then lets add the finalizer and update the object. This is equivalent
+		// registering our finalizer.
+		if !containsString(baremetalConfig.ObjectMeta.Finalizers,
+			metal3iov1alpha1.ProvisioningFinalizer) {
+			// Add finalizer becasue it doesn't already exist
+			baremetalConfig.ObjectMeta.Finalizers = append(baremetalConfig.ObjectMeta.Finalizers,
+				metal3iov1alpha1.ProvisioningFinalizer)
+			if err := r.Client.Update(context.Background(), baremetalConfig); err != nil {
```
asalkeld, line 136: you could just early return here

controllers/provisioning_controller.go
```diff
@@ -152,32 +209,8 @@ func (r *ProvisioningReconciler) Reconcile(req ctrl.Request) (ctrl.Result, error
...
-			// Add any specific deletion logic here
-
-			// Remove our finalizer from the list and update it.
-			baremetalConfig.ObjectMeta.Finalizers = removeString(baremetalConfig.ObjectMeta.Finalizers,
-				metal3iov1alpha1.ProvisioningFinalizer)
-			if err := r.Client.Update(context.Background(), baremetalConfig); err != nil {
-				return ctrl.Result{}, errors.Wrap(err, "failed to remove finalizer from Provisioning CR")
-			}
-		}
+	if err = r.checkForCRDeletion(baremetalConfig); err != nil {
```
asalkeld, line 212: if you have just deleted everything, is it worth continuing?
maybe
```
deleted, err = r.checkForCRDeletion()
if err != nil || deleted {
    return ctrl.Result{}, err
}
```

zaneb: I'd go so far as to say it's essential not to continue, otherwise it'll all get recreated again :)

controllers/provisioning_controller.go
```diff
@@ -120,6 +120,63 @@ func (r *ProvisioningReconciler) readProvisioningCR(req ctrl.Request) (*metal3io
...
+			if err := r.Client.Update(context.Background(), baremetalConfig); err != nil {
+				return errors.Wrap(err, "failed to update Provisioning CR with its finalizer")
+			}
+		}
+	} else {
+		// The Provisioning object is being deleted
+		if containsString(baremetalConfig.ObjectMeta.Finalizers, metal3iov1alpha1.ProvisioningFinalizer) {
+			err := r.deleteMetal3Resources()
+			if err != nil {
+				return errors.Wrap(err, "failed to delete metal3 pod")
```
zaneb, line 145: It could be any of the resources.

controllers/provisioning_controller.go
```diff
@@ -120,6 +120,63 @@ func (r *ProvisioningReconciler) readProvisioningCR(req ctrl.Request) (*metal3io
...
+	}
+	return nil
+}
+
+//Delete Secrets and the Metal3 Deployment objects
+func (r *ProvisioningReconciler) deleteMetal3Resources() error {
+	err := provisioning.DeleteMariadbPasswordSecret(r.KubeClient.CoreV1(), ComponentNamespace)
+	if err != nil {
+		r.Log.V(1).Info("could not delete mariadb password Secret")
+	}
```
zaneb, line 163: Need to return inside the if{} block here, otherwise `err` will simply get overwritten.

provisioning/baremetal_secrets.go
```diff
@@ -161,3 +161,33 @@ password = %s
 	_, err = client.Secrets(targetNamespace).Create(context.Background(), secret, metav1.CreateOptions{})
 	return err
 }
+
+func DeleteMariadbPasswordSecret(client coreclientv1.SecretsGetter, targetNamespace string) error {
```
asalkeld, line 165: why not make one function DeleteSecretIfExists(client, targetNamespace, targetName)

controllers/provisioning_controller.go
asalkeld: yeah, i was thinking "you are just trying to delete as much as possible and not stress about errors". except you are already taking notFound into account. what other reasons could we have? also if we do anywhere return an error and reenter the Reconcile, how are we going to detect that we are in the middle of a delete? is it just going to say "there is no CR"

controllers/provisioning_controller.go
sadasu: Done.

controllers/provisioning_controller.go
sadasu: Done.

controllers/provisioning_controller.go
sadasu: When we encounter an error while deleting a secret, we just log it and keep going. The return value of deleteMetal3Resources() is just the result of deleting the metal3 pod.

controllers/provisioning_controller.go
```diff
@@ -120,6 +120,63 @@ func (r *ProvisioningReconciler) readProvisioningCR(req ctrl.Request) (*metal3io
 	return instance, nil
 }
 
+//Ensure Finalizer is present on the Provisioning CR when not deleted and
+//delete resources and remove Finalizer when it is
+func (r *ProvisioningReconciler) checkForCRDeletion(baremetalConfig *metal3iov1alpha1.Provisioning) error {
```
asalkeld, line 125: i still think you need to return a bool here to say whether or not you have deleted the resources,
and if you have don't continue with the Reconcile .

controllers/provisioning_controller.go
asalkeld: if it has successfully deleted all resouces it will return nil here and then it will re-create the deployment below

controllers/provisioning_controller.go
```diff
@@ -156,39 +156,47 @@ func (r *ProvisioningReconciler) Reconcile(req ctrl.Request) (ctrl.Result, error
...
+	if err == nil && deleted {
+		err = r.updateCOStatus(ReasonComplete, "all Metal3 resources deleted", "")
+		if err != nil {
+			return ctrl.Result{}, fmt.Errorf("unable to put %q ClusterOperator in Available state: %v", clusterOperatorName, err)
 		}
+		return ctrl.Result{}, nil
 	}
 
+	// Make sure ClusterOperator exists and updated
 	err = r.ensureClusterOperator(baremetalConfig)
```
asalkeld, line 194: shouldn't this be above the first call to updateCOStatus()?

sadasu:
/retitle Bug 1906935: Delete resources when Provisioning CR is deleted

controllers/provisioning_controller.go
sadasu: We are now returning a bool & error so, can't return early.

controllers/provisioning_controller.go
sadasu: If we encounter an error while deleteing a resource we return without removing the finalizer. The fact that the finalizer still exists indicates that the cleanup operation is not complete and the Provisioning CR would not be deleted.

provisioning/baremetal_secrets.go
sadasu: Done.

controllers/provisioning_controller.go
sadasu: Done.

controllers/provisioning_controller.go
sadasu: Done.

controllers/provisioning_controller.go
sadasu: Reworded.

controllers/provisioning_controller.go
sadasu: > Need to return inside the if{} block here, otherwise `err` will simply get overwritten.

We are not returning an error when encountering an issue with deleting a secret. We would just log the fact that we could not delete a secret. We will return an error in the case of other resources and try to delete it next time.

sadasu:
/test e2e-agnostic

Failure seemed related to the kube version. Now that we have moved to 1.20, I expect this CI to pass too.

kirankt (approved):
/lgtm

controllers/provisioning_controller.go
```diff
@@ -153,39 +153,45 @@ func (r *ProvisioningReconciler) Reconcile(req ctrl.Request) (ctrl.Result, error
...
-				return ctrl.Result{}, errors.Wrap(err, "failed to remove finalizer from Provisioning CR")
-			}
+	// Read container images from Config Map
+	var containerImages provisioning.Images
+	if err := provisioning.GetContainerImages(&containerImages, r.ImagesFilename); err != nil {
+		// Images config map is not valid
+		// Provisioning configuration is not valid.
+		// Requeue request.
+		r.Log.Error(err, "invalid contents in images Config Map")
+		co_err := r.updateCOStatus(ReasonInvalidConfiguration, err.Error(), "invalid contents in images Config Map")
```
asalkeld, line 163: all these calls to updateCOStatus() can fail if the CO does not exist. Can you please move the ensureClusterOperator() to above the first possible call to updateCOStatus?

controllers/provisioning_controller.go
```diff
@@ -153,39 +153,45 @@ func (r *ProvisioningReconciler) Reconcile(req ctrl.Request) (ctrl.Result, error
...
+		err = r.updateCOStatus(ReasonDeployTimedOut, err.Error(), "Unable to delete a metal3 resource on Provisioning CR deletion")
+		if err != nil {
+			return ctrl.Result{}, fmt.Errorf("unable to put %q ClusterOperator in Degraded state: %v", clusterOperatorName, err)
+		}
+		return ctrl.Result{}, err
+	}
+	if err == nil && deleted {
+		err = r.updateCOStatus(ReasonComplete, "all Metal3 resources deleted", "")
+		if err != nil {
+			return ctrl.Result{}, fmt.Errorf("unable to put %q ClusterOperator in Available state: %v", clusterOperatorName, err)
```
asalkeld, line 190:
```suggestion
			return ctrl.Result{}, fmt.Errorf("unable to put %q ClusterOperator in Available state: %w", clusterOperatorName, err)
```

asalkeld: i believe that when wrapping errors we should be using %w as this allows the error to be unwrapped (please check in the rest of the PR too)

controllers/provisioning_controller.go
```diff
@@ -335,6 +317,68 @@ func (r *ProvisioningReconciler) provisioningInfo(provConfig *metal3iov1alpha1.P
...
+			}
+			return true, nil
+		}
+	}
+	return false, nil
+}
+
+//Delete Secrets and the Metal3 Deployment objects
+func (r *ProvisioningReconciler) deleteMetal3Resources(info *provisioning.ProvisioningInfo) error {
+	failedSecrets := provisioning.DeleteAllSecrets(info)
```
asalkeld, line 359: i think this should return an error

provisioning/baremetal_secrets.go
```diff
@@ -168,3 +154,41 @@ password = %s
...
+	}
+	// Create a Secret for the Ironic Inspector Password
+	if err := createIronicSecret(client, targetNamespace, inspectorSecretName, inspectorUsername, "inspector", baremetalConfig, scheme); err != nil {
+		return errors.Wrap(err, "failed to create Inspector password")
+	}
+	return nil
+}
+
+func DeleteAllSecrets(info *ProvisioningInfo) []string {
+	deleteFailures := []string{}
```
asalkeld, line 179: errs := []error

provisioning/baremetal_secrets.go
```diff
@@ -168,3 +154,41 @@ password = %s
...
+		deleteFailures = append(deleteFailures, ironicSecretName)
+	}
+	if err := info.Client.CoreV1().Secrets(info.Namespace).Delete(context.Background(), inspectorSecretName, metav1.DeleteOptions{}); err != nil {
+		deleteFailures = append(deleteFailures, inspectorSecretName)
+	}
+	if err := info.Client.CoreV1().Secrets(info.Namespace).Delete(context.Background(), ironicrpcSecretName, metav1.DeleteOptions{}); err != nil {
+		deleteFailures = append(deleteFailures, ironicrpcSecretName)
+	}
+
+	return deleteFailures
```
asalkeld, line 193: return errors.NewAggregate(errs)

controllers/provisioning_controller.go
```diff
@@ -335,6 +317,68 @@ func (r *ProvisioningReconciler) provisioningInfo(provConfig *metal3iov1alpha1.P
...
+
+	err := provisioning.DeleteMetal3Deployment(info)
+	if err != nil {
+		return errors.Wrap(err, "failed to delete metal3 deployment")
+	}
+	err = provisioning.DeleteMetal3StateService(info)
+	if err != nil {
+		return errors.Wrap(err, "failed to delete metal3 service")
+	}
+	err = provisioning.DeleteImageCache(info)
```
asalkeld, line 375: What do you think about this? We have a lot of repeated functions that just check for notFound
```golang
err = provisioning.IgnoreNotFound(info.Client.AppsV1().DaemonSets(info.Namespace).Delete(context.Background(), imageCacheService, metav1.DeleteOptions{}))
```

```
func IgnoreNotFound(err error) error {
	if apierrors.IsNotFound(err) {
		return nil
	}
	return err
}
```

controllers/provisioning_controller.go
sadasu: Updated. Please let me know if you are fine with the current amount of refactoring done within Reconcile() so that it doesn't get bloated.

controllers/provisioning_controller.go
```diff
@@ -153,6 +159,13 @@ func (r *ProvisioningReconciler) Reconcile(req ctrl.Request) (ctrl.Result, error
 		return ctrl.Result{}, nil
 	}
 
+	// Provisioning CR is present
```
asalkeld, line 162: do we need this a second time?

sadasu (commented):
I am assuming you are referring the 2nd call to ensureClusterOperator(). And, yes, this time it will add the baremetal config in the OwnerReference for the CO. 
The 1st time ensureClusterOperator(), we pass in nil because the baremetalConfig is not yet read and verified to be correct at that time. And the only time we update the CO before reading/verifying baremetalConfig is when the platform != BareMetal, so passing in nil and not setting the OwnerReference should be OK.

controllers/provisioning_controller.go
sadasu: Do you think deleteMetal3Resources() should return error when DeleteAllSecrets() returns error(s) or just log it and proceed with deleting other resources as it is doing now.

controllers/provisioning_controller.go
sadasu: Done.

provisioning/baremetal_secrets.go
sadasu: Didn't want to import a new errors package so achieved the same result differently.

provisioning/baremetal_secrets.go
sadasu: Fixed.

controllers/provisioning_controller.go
sadasu: Not really going to help the Reconcile method a whole lot. We currently call apierrors.IsNotFound() twice and the number of return values are different.
In other places in the provisioning package, we use the return value of apierrors.IsNotFound() to decide if we need to create something and not return nil.

sadasu:
/test e2e-agnostic

Errors unrelated to CBO.

PR #72, opened 2020-11-21
Enable Deployment updates

zaneb:
Unblock updating of the metal3 Deployment created by the CBO (but not those originally created by the MAO - this will be added in future by #65).

### DIFF
--- a/controllers/provisioning_controller.go
+++ b/controllers/provisioning_controller.go
@@ -33,7 +33,6 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/controller/controllerutil"
 
 	osconfigv1 "github.com/openshift/api/config/v1"
-	osoperatorv1 "github.com/openshift/api/operator/v1"
 	osclientset "github.com/openshift/client-go/config/clientset/versioned"
 	metal3iov1alpha1 "github.com/openshift/cluster-baremetal-operator/api/v1alpha1"
 	provisioning "github.com/openshift/cluster-baremetal-operator/provisioning"
@@ -63,8 +62,6 @@ type ProvisioningReconciler struct {
 	KubeClient     kubernetes.Interface
 	ReleaseVersion string
 	ImagesFilename string
-
-	Generations []osoperatorv1.GenerationStatus
 }
 
 // +kubebuilder:rbac:namespace=openshift-machine-api,groups="",resources=configmaps,verbs=get;list;watch;create;update;patch;delete
@@ -244,7 +241,7 @@ func (r *ProvisioningReconciler) Reconcile(req ctrl.Request) (ctrl.Result, error
 
 	// Proceed with creating a new Metal3 deployment
 	metal3Deployment := provisioning.NewMetal3Deployment(ComponentNamespace, &containerImages, &baremetalConfig.Spec)
-	expectedGeneration := resourcemerge.ExpectedDeploymentGeneration(metal3Deployment, r.Generations)
+	expectedGeneration := resourcemerge.ExpectedDeploymentGeneration(metal3Deployment, baremetalConfig.Status.Generations)
 
 	err = controllerutil.SetControllerReference(baremetalConfig, metal3Deployment, r.Scheme)
 	if err != nil {
@@ -258,7 +255,11 @@ func (r *ProvisioningReconciler) Reconcile(req ctrl.Request) (ctrl.Result, error
 	}
 
 	if updated {
-		resourcemerge.SetDeploymentGeneration(&r.Generations, deployment)
+		resourcemerge.SetDeploymentGeneration(&baremetalConfig.Status.Generations, deployment)
+		err = r.Client.Status().Update(context.Background(), baremetalConfig)
+		if err != nil {
+			return ctrl.Result{}, err
+		}
 	}
 	err = r.updateCOStatus(ReasonComplete, "new Metal3 deployment completed", "")
 	if err != nil {
--- a/controllers/provisioning_controller.go
+++ b/controllers/provisioning_controller.go
@@ -239,34 +239,45 @@ func (r *ProvisioningReconciler) Reconcile(req ctrl.Request) (ctrl.Result, error
 		return ctrl.Result{}, fmt.Errorf("unable to put %q ClusterOperator in Syncing state: %v", clusterOperatorName, err)
 	}
 
-	// Proceed with creating a new Metal3 deployment
-	metal3Deployment := provisioning.NewMetal3Deployment(ComponentNamespace, &containerImages, &baremetalConfig.Spec)
-	expectedGeneration := resourcemerge.ExpectedDeploymentGeneration(metal3Deployment, baremetalConfig.Status.Generations)
-
-	err = controllerutil.SetControllerReference(baremetalConfig, metal3Deployment, r.Scheme)
+	// Proceed with creating or updating the Metal3 deployment
+	updated, err := r.ensureMetal3Deployment(baremetalConfig, &containerImages)
 	if err != nil {
-		return ctrl.Result{}, fmt.Errorf("unable to set controllerReference on deployment: %v", err)
+		return ctrl.Result{}, err
 	}
-
-	deployment, updated, err := resourceapply.ApplyDeployment(r.KubeClient.AppsV1(),
-		events.NewLoggingEventRecorder(ComponentName), metal3Deployment, expectedGeneration)
-	if err != nil {
-		return ctrl.Result{}, fmt.Errorf("unable to apply Metal3 deployment: %v", err)
-	}
-
 	if updated {
-		resourcemerge.SetDeploymentGeneration(&baremetalConfig.Status.Generations, deployment)
-		err = r.Client.Status().Update(context.Background(), baremetalConfig)
-		if err != nil {
-			return ctrl.Result{}, err
-		}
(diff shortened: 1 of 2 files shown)


provisioning/baremetal_pod.go
```diff
@@ -471,7 +471,9 @@ func NewMetal3Deployment(targetNamespace string, images *Images, config *metal3i
 func GetExistingMetal3Deployment(client appsclientv1.DeploymentsGetter, targetNamespace string) (bool, error) {
 	existing, err := client.Deployments(targetNamespace).Get(context.Background(), baremetalDeploymentName, metav1.GetOptions{})
 	if existing != nil && err == nil {
-		return true, nil
+		if _, maoOwned := existing.Annotations["machine.openshift.io/owned"]; maoOwned {
+			return true, nil
+		}
```
asalkeld, line 476: i think this changes the entire meaning of the function from does it exist, to is the deployment owned by mao
maybe keep it out of the function?

controllers/provisioning_controller.go
```diff
@@ -222,7 +219,7 @@ func (r *ProvisioningReconciler) Reconcile(req ctrl.Request) (ctrl.Result, error
 		return ctrl.Result{}, errors.Wrap(err, "failed to create Inspector password")
 	}
 
-	// If Metal3 Deployment already exists, do nothing.
+	// If Metal3 Deployment already exists and managed by MAO, do nothing.
```
asalkeld, line 222: I think we need a function named ExistingDeploymentNotOwedByUS() - also get rid of the "Get"

controllers/provisioning_controller.go
```diff
@@ -237,35 +234,59 @@ func (r *ProvisioningReconciler) Reconcile(req ctrl.Request) (ctrl.Result, error
...
+	}
+
+	// Proceed with creating or updating the Metal3 deployment
+	updated, err := r.ensureMetal3Deployment(baremetalConfig, &containerImages)
 	if err != nil {
-		return ctrl.Result{}, fmt.Errorf("unable to put %q ClusterOperator in Syncing state: %v", clusterOperatorName, err)
+		return ctrl.Result{}, err
+	}
+	if updated {
+		return ctrl.Result{Requeue: true}, err
```
asalkeld, line 251: why is this here? this makes the function logically harder to parse.

controllers/provisioning_controller.go
```diff
@@ -237,35 +234,59 @@ func (r *ProvisioningReconciler) Reconcile(req ctrl.Request) (ctrl.Result, error
...
 	}
 
-	// Proceed with creating a new Metal3 deployment
-	metal3Deployment := provisioning.NewMetal3Deployment(ComponentNamespace, &containerImages, &baremetalConfig.Spec)
-	expectedGeneration := resourcemerge.ExpectedDeploymentGeneration(metal3Deployment, r.Generations)
+	if specChanged {
+		baremetalConfig.Status.ObservedGeneration = baremetalConfig.Generation
+		err = r.Client.Status().Update(context.Background(), baremetalConfig)
+		return ctrl.Result{Requeue: true}, err
```
asalkeld, line 257: also this, why don't we continue with the function

controllers/provisioning_controller.go
```diff
@@ -237,35 +234,59 @@ func (r *ProvisioningReconciler) Reconcile(req ctrl.Request) (ctrl.Result, error
...
+	err = r.updateCOStatus(ReasonComplete, "new Metal3 deployment completed", "")
 	if err != nil {
-		return ctrl.Result{}, fmt.Errorf("unable to set controllerReference on deployment: %v", err)
+		return ctrl.Result{}, fmt.Errorf("unable to put %q ClusterOperator in Available state: %v", clusterOperatorName, err)
+	}
+
+	return ctrl.Result{}, nil
+}
+
+func (r *ProvisioningReconciler) ensureMetal3Deployment(provConfig *metal3iov1alpha1.Provisioning, images *provisioning.Images) (updated bool, err error) {
```
asalkeld, line 268: nit, not a fan of named return variables and the associated empty returns

asalkeld (commented):
left some comments..
I thinks this Reconcile function needs breaking up, it's getting to big.. (not your fault, it's just growing)

controllers/provisioning_controller.go
zaneb: It's going to disappear soon anyway, but I agree this isn't very well named.

provisioning/baremetal_pod.go
zaneb: That does sounds less hacky, but I think we have to follow the plan because we need the MAO to do the right thing during downgrades, and the thing we taught it to do in 4.6 is look at the annotations. As far as I know the MAO doesn't update the Deployment at all once it exists, so I doubt that it rewrites the ControllerRef on downgrade, even if it does set it on create.
Again, this code is going away soon anyway. I _think_ we could actually just delete it now and it would work, since we call ApplyDeployment() and IIUC that should rewrite the annotations. But Sandhya has improved tests and stuff in #65, so I'm leaving it for that PR.

controllers/provisioning_controller.go
zaneb: We wrote to the status, so we should requeue otherwise the next write to the status will fail.

controllers/provisioning_controller.go
zaneb: We probably could in this case. I think I wrote this before I completely realised that the Generation only changes when the Spec is modified, so in fact there's no danger that writing the Status could cause us to bounce the CO state from Syncing->Complete->Syncing->Complete.

controllers/provisioning_controller.go
zaneb: Yeah, I can't decide how I feel about them. In this case it seems essential to name the parameter to document what it means, and then after that it feels weird to me to not use the name in the function.

controllers/provisioning_controller.go
asalkeld: seems silly we can't write to the status twice

controllers/provisioning_controller.go
asalkeld: you are allowed to write a comment :smiley:

controllers/provisioning_controller.go
sadasu: The logic around this is getting refined as part of https://github.com/openshift/cluster-baremetal-operator/pull/65. And, please feel free to propose better names.

provisioning/baremetal_pod.go
sadasu: MAO adds the maoOwnedAnnotations to all the resources it creates and so we just using it.

This documentation about ControllerRef says it is partially implemented https://stupefied-goodall-e282f7.netlify.app/contributors/design-proposals/api-machinery/controller-ref/. 

This is hand-off mechanism is required only during upgrade/downgrade between 4.6 and 4.7.

provisioning/baremetal_pod.go
```diff
@@ -468,10 +468,12 @@ func NewMetal3Deployment(targetNamespace string, images *Images, config *metal3i
 	}
 }
 
-func GetExistingMetal3Deployment(client appsclientv1.DeploymentsGetter, targetNamespace string) (bool, error) {
+func MAOMetal3DeploymentExists(client appsclientv1.DeploymentsGetter, targetNamespace string) (bool, error) {
 	existing, err := client.Deployments(targetNamespace).Get(context.Background(), baremetalDeploymentName, metav1.GetOptions{})
 	if existing != nil && err == nil {
-		return true, nil
+		if _, maoOwned := existing.Annotations["machine.openshift.io/owned"]; maoOwned {
```
sadasu, line 474: This check really is not necessary because even if the metal3 deployment was created by MAO, we are allowed to change it.

provisioning/baremetal_pod.go
sadasu: I do not think the name suggested here is correct. We want it to be a generic method to see if a metal3 deployment already exists after/during an upgrade. When we upgrade from 4.6 to 4.7, the older metal3 deployment would have been created by MAO. From 4.7 to 4.8, it would have been created by CBO. 
So, I am not for adding MAO in the name of the method.

sadasu:
I think commit "Allow metal3 Deployment to be updated" is not necessary and comments against code in that commit reflect the reasons why. All other commits look good.
Also, as previously discussed #65 refactors a lot of the code around metal3 deployment better.

controllers/provisioning_controller.go
sadasu: Can we move ensureMetal3Deployment() to baremetal_pod.go?

sadasu:
> I think commit "Allow metal3 Deployment to be updated" is not necessary and comments against code in that commit reflect the reasons why. All other commits look good.
> Also, as previously discussed #65 refactors a lot of the code around metal3 deployment better.

Reading the description again, I see that the "Allow metal3 Deployment to be updated" commit is just meant to get you unblocked for further work. I am Ok for this to merge as is. I can rebase #65 over these changes and rework the logic around annotations to its final form.

sadasu:
Nice changes to the expected generations code.

provisioning/baremetal_pod.go
zaneb: I am expecting that this method will either go away or be completely rewritten (with a new name) in #65.

PR #73, opened 2020-11-23
Use hostIP for ironic containers when provisionig network disabled 

stbenjam:
This change makes the ironic containers get the hostIP when the
provisioningIP is blank. This removes the requirement for a
provisioningIP in Ironic, but additional work is needed to fix the image
URL's we pass to baremetal-operator.

As we are using an existing IP on an interface, we no longer have a
requirement on the static IP manager when the provisioningIP is blank,
so we do not start the containers in this case.

This also changes baremetal operator to communicate with the ironic
containers on localhost, as they are in the same pod.

### DIFF
--- a/provisioning/baremetal_config.go
+++ b/provisioning/baremetal_config.go
@@ -116,15 +116,17 @@ func validateDisabledConfig(prov *metal3iov1alpha1.Provisioning) error {
 		Name  string
 		Value string
 	}{
-
-		{Name: "ProvisioningIP", Value: prov.Spec.ProvisioningIP},
-		{Name: "ProvisioningNetworkCIDR", Value: prov.Spec.ProvisioningNetworkCIDR},
 		{Name: "ProvisioningOSDownloadURL", Value: prov.Spec.ProvisioningOSDownloadURL},
 	} {
 		if toTest.Value == "" {
 			return fmt.Errorf("%s is required but is empty", toTest.Name)
 		}
 	}
+
+	if prov.Spec.ProvisioningNetworkCIDR == "" && prov.Spec.ProvisioningIP != "" {
+		return fmt.Errorf("provisioningNetworkCIDR is required when specifying a provisioningIP")
+	}
+
 	return nil
 }
 
@@ -156,20 +158,14 @@ func getDeployRamdiskUrl(config *metal3iov1alpha1.ProvisioningSpec) *string {
 	return nil
 }
 
-func getIronicEndpoint(config *metal3iov1alpha1.ProvisioningSpec) *string {
-	if config.ProvisioningIP != "" {
-		ironicEndpoint := fmt.Sprintf("http://%s/%s", net.JoinHostPort(config.ProvisioningIP, baremetalIronicPort), baremetalIronicEndpointSubpath)
-		return &ironicEndpoint
-	}
-	return nil
+func getIronicEndpoint() *string {
+	ironicEndpoint := fmt.Sprintf("http://localhost:%s/%s", baremetalIronicPort, baremetalIronicEndpointSubpath)
+	return &ironicEndpoint
 }
 
-func getIronicInspectorEndpoint(config *metal3iov1alpha1.ProvisioningSpec) *string {
-	if config.ProvisioningIP != "" {
-		inspectorEndpoint := fmt.Sprintf("http://%s/%s", net.JoinHostPort(config.ProvisioningIP, baremetalIronicInspectorPort), baremetalIronicEndpointSubpath)
-		return &inspectorEndpoint
-	}
-	return nil
+func getIronicInspectorEndpoint() *string {
+	ironicInspectorEndpoint := fmt.Sprintf("http://localhost:%s/%s", baremetalIronicInspectorPort, baremetalIronicEndpointSubpath)
+	return &ironicInspectorEndpoint
 }
 
 func getProvisioningOSDownloadURL(config *metal3iov1alpha1.ProvisioningSpec) *string {
@@ -190,9 +186,9 @@ func getMetal3DeploymentConfig(name string, baremetalConfig *metal3iov1alpha1.Pr
 	case deployRamdiskUrl:
 		return getDeployRamdiskUrl(baremetalConfig)
 	case ironicEndpoint:
-		return getIronicEndpoint(baremetalConfig)
+		return getIronicEndpoint()
 	case ironicInspectorEndpoint:
-		return getIronicInspectorEndpoint(baremetalConfig)
+		return getIronicInspectorEndpoint()
 	case httpPort:
 		return pointer.StringPtr(baremetalHttpPort)
 	case dhcpRange:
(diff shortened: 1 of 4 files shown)


provisioning/baremetal_config.go
```diff
@@ -140,36 +139,53 @@ func getProvisioningIPCIDR(config *metal3iov1alpha1.ProvisioningSpec) *string {
...
+func joinHostIPPort(config *metal3iov1alpha1.ProvisioningSpec, port string) string {
+	_, net, err := net.ParseCIDR(config.ProvisioningNetworkCIDR)
+	if err == nil {
+		if net.IP.To4() == nil {
+			return fmt.Sprintf("[$(PROVISIONING_IP)]:%s", port)
+		}
+	}
+
+	return fmt.Sprintf("$(PROVISIONING_IP):%s", port)
+}
```
stbenjam, line 155: My whole solution for this is a disaster.

Could we query the Kubernetes API somewhere else to get the node's host IP and pass that through to here? I didn't find a logical place to do it.

Still investigating but open to ideas.

provisioning/baremetal_config_test.go
```diff
@@ -41,63 +41,29 @@ func TestValidateManagedProvisioningConfig(t *testing.T) {
...
-				ProvisioningDHCPRange:     "172.30.20.11, 172.30.20.101",
-				ProvisioningOSDownloadURL: "http://172.22.0.1/images/rhcos-44.81.202001171431.0-openstack.x86_64.qcow2.gz?sha256=e98f83a2b9d4043719664a2be75fe8134dc6ca1fdbde807996622f8cc7ecd234",
-				ProvisioningDHCPExternal:  false,
-			},
-			expectedError: false,
-			expectedMode:  metal3iov1alpha1.ProvisioningNetworkManaged,
-		},
-		{
-			// Verifying default behavior where both ProvisioningNetwork and ProvisioningDHCPExternal are not specified.
-			name: "DefaultManaged",
```
stbenjam, line 79: I removed this test since it was a duplicate of the one below. There's no concept of "not specifying" a boolean field, it's zero-value is false (unless it's a pointer, which in this case it's not).

provisioning/baremetal_config.go
zaneb: If we were to run [ironic-ipa-downloader](https://github.com/openshift/ironic-ipa-downloader/blob/master/get-resource.sh) (with some modifications to ensure it doesn't download multiple times from the Internet) in the [second-level image cache](https://github.com/zaneb/cluster-baremetal-operator/commits/image-cache) then we could refer to this file in the second-level cache using the API VIP when there is no provisioning VIP. This is probably a good thing anyway as it would mean we never have to re-download IPA from the Internet once we have it.

Out of curiousity, can anybody explain why we need to both [set this path](https://github.com/metal3-io/ironic-image/blob/master/runhttpd.sh#L34-L37) in [inspector.pxe](https://github.com/metal3-io/ironic-image/blob/master/inspector.ipxe) *and* [pass it to ironic](https://github.com/metal3-io/baremetal-operator/blob/master/pkg/provisioner/ironic/ironic.go#L294-L297) from the BMO?

stbenjam: > f we were to run ironic-ipa-downloader (with some modifications to ensure it doesn't download multiple times from the Internet) in the second-level image cache then we could refer to this file in the second-level cache using the API VIP when there is no provisioning VIP. 

Indeed, this is simpler if we get the second-level stuff up and running. If we wanted this solution to work by itself  though we need a way to generate the URL.

I'm pretty sure I could query for pod for the IP somewhere in the early controller start-up code? I'll play around with that today.

> Out of curiousity, can anybody explain why we need to both set this path in inspector.pxe and pass it to ironic from the BMO?

Just a guess, but I think it uses inspector.pxe for PXE booting, and Ironic itself needs it for virtuamedia boots

stbenjam: > I'm pretty sure I could query for pod for the IP somewhere in the early controller start-up code? I'll play around with that today.

Angus already wrote it  :) 
https://github.com/openshift/cluster-baremetal-operator/commit/c2ab117ed3001e72389f031d7b4fde28f3c96ae1

stbenjam:
After a discussion on Slack, this PR is only handling the Ironic and static-ip-manager parts. The URL's we pass to baremetal-operator are going to need to use the DaemonSet we discussed before and the second-level caching layer using the API VIP.

That means, this PR is not testable by itself, other than making sure it doesn't break any current workflows.

We need to enable some more CI jobs to be able to test disabled network (virtualmedia). See https://github.com/openshift/release/pull/13850

stbenjam:
@zaneb I've broken the commits apart and updated as discussed, please take another look! Thanks

asalkeld:
lgtm

zaneb:
The last step to set the IPA URLs independently of the provisioning VIP will be [something like this](https://github.com/zaneb/cluster-baremetal-operator/commit/ipa-url-api-vip).

provisioning/baremetal_pod.go
```diff
@@ -109,11 +109,20 @@ func buildEnvVar(name string, baremetalProvisioningConfig *metal3iov1alpha1.Prov
 			Name:  name,
 			Value: *value,
 		}
-	} else {
+	} else if name == provisioningIP && baremetalProvisioningConfig.ProvisioningNetwork == metal3iov1alpha1.ProvisioningNetworkDisabled {
```
sadasu, line 112: This logic can be added to https://github.com/openshift/cluster-baremetal-operator/blob/master/provisioning/baremetal_config.go#L131. The get*() methods in baremetal_config.go are meant to take care of all the interdependencies between the config. Implementing it like this will spread the knowledge of config interdependencies in multiple places.

stbenjam: That function returns a string pointer. I need to construct a different kind of environment variable using a FieldRef.

sadasu: Is the status.hostIP going to be localhost?

stbenjam: I agree it would be nicer if this logic wasn't here, but I didn't find an easy way to do it. I'm happy to refactor it later if we can let this land. There's still a lot left to do for 3704.

sadasu: So, should this value be an actual IP address vs "localhost"?

sadasu: It is Ok to merge even with some outstanding comments. We can capture it here and fix it later.

zaneb: It's filled in by the runtime, but yes it will be an actual IP address

sadasu:
Nice improvements to the unit test code. Thanks for answering my questions.
Some minor comments inline that can be fixed in a later iteration.

provisioning/baremetal_pod.go
stbenjam: > So, should this value be an actual IP address vs "localhost"?

The ironic containers themselves do get the host IP.  The FieldRef in this code is telling kubernetes to wire that value in at run time. CBO can't know the value since it won't know where the pod will land.

stbenjam:

Not sure why it didn't stick

PR #75, opened 2020-11-24
provisioning controller: set correct status when platform=baremetal and no CR is found

stbenjam:
Assisted deploys with platform=baremetal, but without any provisioning
CR. I believe the correct fix for this is to set the operator to
Available=true, Disabled=true, Progressing=false, and requeue to wait to
see if a provisioning CR is created.

### DIFF
--- a/controllers/provisioning_controller.go
+++ b/controllers/provisioning_controller.go
@@ -147,9 +147,14 @@ func (r *ProvisioningReconciler) Reconcile(req ctrl.Request) (ctrl.Result, error
 	}
 	if baremetalConfig == nil {
 		// Provisioning configuration not available at this time.
-		// Cannot proceed wtih metal3 deployment.
-		r.Log.V(1).Info("Provisioning CR not found")
-		return ctrl.Result{}, nil
+		// Cannot proceed with metal3 deployment.
+		err = r.updateCOStatus(ReasonUnsupported, "Provisioning CR not found", "")
+		if err != nil {
+			return ctrl.Result{}, fmt.Errorf("unable to put %q ClusterOperator in Disabled state: %v", clusterOperatorName, err)
+		}
+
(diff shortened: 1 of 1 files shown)


controllers/provisioning_controller.go
```diff
@@ -147,9 +147,14 @@ func (r *ProvisioningReconciler) Reconcile(req ctrl.Request) (ctrl.Result, error
...
-		r.Log.V(1).Info("Provisioning CR not found")
-		return ctrl.Result{}, nil
+		// Cannot proceed with metal3 deployment.
+		err = r.updateCOStatus(ReasonUnsupported, "Provisioning CR not found", "")
+		if err != nil {
+			return ctrl.Result{}, fmt.Errorf("unable to put %q ClusterOperator in Disabled state: %v", clusterOperatorName, err)
+		}
+
+		// We're disabled, but requeue in case we get a provisioning CR
+		return ctrl.Result{}, fmt.Errorf("no provisioning CR found; will remain Disabled until CR available")
```
stbenjam, line 157: I'm not entirely clear if this is the right fix; will we hit the Reconcile loop if there's no CR at all? This might need to go to main.

asalkeld:
```suggestion
		return ctrl.Result{}, nil
```
We are watching the CR so we will get a Reconcile if one is created/updated. this will just cause a lot of unnecessary Reconcile calls

stbenjam: I guess we need to check in main; assisted is still failing and looking at the logs it looks like it does start but we're never reconciling which makes sense if there's no CR at all yet.

asalkeld: yeah, this will only hit if you delete the cr. We probably need to initialize the status to Unsupported in case the cr never gets created

stbenjam:
/close

Fixed by #81

PR #76, opened 2020-11-25
Add a second-level image cache

zaneb:
Add a cache for OS images accessible on port 6181 on every master node.
This cache downloads the images from the first-level cache in the metal3
Pod (port 6180). The images are kept in shared storage, so even when the
metal3 Pod moves, we'll never need to download them from the Internet
again.

The first few commits here are already proposed in #72.

### DIFF
--- a/provisioning/image_cache.go
+++ b/provisioning/image_cache.go
@@ -1,12 +1,34 @@
 package provisioning
 
 import (
+	"fmt"
+	"net"
+	"net/url"
+	"path"
+	"regexp"
+	"strconv"
+
+	appsv1 "k8s.io/api/apps/v1"
 	corev1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/util/intstr"
+	"k8s.io/utils/pointer"
+	"sigs.k8s.io/controller-runtime/pkg/controller/controllerutil"
+
+	"github.com/openshift/library-go/pkg/operator/resource/resourceapply"
+	"github.com/openshift/library-go/pkg/operator/resource/resourcemerge"
+
+	metal3iov1alpha1 "github.com/openshift/cluster-baremetal-operator/api/v1alpha1"
 )
 
 const (
 	imageCacheSharedVolume = "metal3-shared-image-cache"
+	imageCacheService      = "metal3-image-cache"
+	imageCachePort         = 6181
+	imageCachePortName     = "http"
 )
+
+var fileCompressionSuffix = regexp.MustCompile(`\.[gx]z$`)
 
 func imageVolume() corev1.Volume {
 	volType := corev1.HostPathDirectoryOrCreate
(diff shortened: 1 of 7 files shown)


stbenjam:
My local test shows an RBAC problem with services:

```
2020-11-25T18:38:09.167Z	ERROR	controller-runtime.controller	Reconciler error	{"controller": "provisioning", "request": "/provisioning-configuration", "error": "unable to apply Metal3-state service: services \"metal3-state\" is forbidden: User \"system:serviceaccount:openshift-machine-api:cluster-baremetal-operator\" cannot get resource \"services\" in API group \"\" in the namespace \"openshift-machine-api\""}
```

stbenjam:
It was the way I tested it: using the dev-scripts _LOCAL_IMAGE doesn't seem to work when you're updating manifests, since I think there's some magic in the CVO that collects them and bundles them together.

Anyway, testing a build made by cluster-bot, I get the daemonset:

```
NAME                 DESIRED   CURRENT   READY   UP-TO-DATE   AVAILABLE   NODE SELECTOR                     AGE
metal3-image-cache   3         3         3       3            3           node-role.kubernetes.io/master=   14m
```

And can fetch stuff from it:

```
$ curl -I http://metal3-state.apps.ostest.test.metalkube.org:6181/images/ironic-python-agent.initramfs 
HTTP/1.1 200 OK
Date: Wed, 25 Nov 2020 21:33:47 GMT
Server: Apache
Last-Modified: Wed, 25 Nov 2020 21:18:35 GMT
ETag: "1c4c1c61-5b4f4f86512f9"
Accept-Ranges: bytes
Content-Length: 474750049
```

asalkeld:
so Zane about these label selectors.. I assume that we not care as we are not touching mao owned resources and we haven't released cbo yet?
Also re: mao owned resources.. what happens when moa stops updating these and we ignore them.. or are we adding a patch to mao to remove the ownership label (or just delete the resources) once we formally transition to cbo?

asalkeld:
Really nice patch set btw :+1:

zaneb:
I have removed the changes to the metal3 Deployment's Pod selector from this PR. Fixing that will be a more involved process documented here:

* https://bugzilla.redhat.com/show_bug.cgi?id=1903717
* https://bugzilla.redhat.com/show_bug.cgi?id=1903700

In the meantime, we are now just adding labels to the Pod template so that the new Service can use a correct selector, but not updating the pod selector of the Deployment itself.

sadasu:
As the machineOSDownloader init container is run as part of the daemonset, does it still have to be part of the metal3 pod?

zaneb:
> As the machineOSDownloader init container is run as part of the daemonset, does it still have to be part of the metal3 pod?

It does, because the one in the daemonset downloads it from the one in the metal3 pod.

provisioning/state_service.go
```diff
@@ -0,0 +1,58 @@
...
+	stateService = "metal3-state"
+)
+
+func newMetal3StateService(targetNamespace string) *corev1.Service {
+	port, _ := strconv.Atoi(baremetalHttpPort) // #nosec
+	return &corev1.Service{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      stateService,
+			Namespace: targetNamespace,
+			Annotations: map[string]string{
```
dhellmann, line 24: Do we need this annotation for things we create dynamically? I thought this was a signal to the thing that reads the manifests?

provisioning/image_cache.go
```diff
@@ -1,13 +1,33 @@
...
+	"github.com/openshift/library-go/pkg/operator/resource/resourceapply"
+	"github.com/openshift/library-go/pkg/operator/resource/resourcemerge"
+
+	metal3iov1alpha1 "github.com/openshift/cluster-baremetal-operator/api/v1alpha1"
 )
 
 const (
 	imageCacheSharedVolume = "metal3-shared-image-cache"
+	imageCacheService      = "metal3-image-cache"
+	imageCachePort         = "6181"
```
dhellmann, line 26: Is this a host port? Should it be registered in https://github.com/openshift/enhancements/blob/master/enhancements/network/host-port-registry.md?

provisioning/state_service.go
zaneb: I have no idea, very much cargo-culting it. @stbenjam?

provisioning/image_cache.go
zaneb: Yes; Probably; TIL; and why is 6180 not already in that list?

provisioning/state_service.go
sadasu: I think CVO is the only one reading this annotation for the resources that it is creating using the files in /manifests. Although, I do see this as the only place it is being added dynamically https://github.com/openshift/machine-api-operator/blob/master/pkg/operator/baremetal_pod.go#L89.

Probably harmless here but not necessary.

provisioning/image_cache.go
zaneb: openshift/enhancements#556

stbenjam:
Can we update this code here to point the images to the API VIP when ProvisioningIP is blank? Or should that be a separate PR?

https://github.com/openshift/cluster-baremetal-operator/blob/55db2ce0b66e9517f4308b2860b05fd296647cc3/provisioning/baremetal_config.go#L145-L160

zaneb:
> Can we update this code here to point the images to the API VIP when ProvisioningIP is blank? Or should that be a separate PR?

I'll do it as a separate PR once this merges. I [have a patch](https://github.com/zaneb/cluster-baremetal-operator/commit/ipa-url-api-vip) basically ready to go.

stbenjam:
Ok, I'm happy with this and I think everyone else is too, so let's land it.

/lgtm

stbenjam:
e2e-agnostic failures are the on-going gcp issues. Install completed, so we're past the point where CBO is relevant.

```
2020/12/03 04:26:02 Executing pod "e2e-agnostic-ipi-install-install"
2020/12/03 05:08:28 Container test in pod e2e-agnostic-ipi-install-install completed successfully
```

stbenjam:
Looks like maybe we broke something:

```
2020-12-03T17:17:13.485448990Z E1203 17:17:13.485288       1 reflector.go:127] sigs.k8s.io/controller-runtime/pkg/cache/internal/informers_map.go:224: Failed to watch *v1.Service: failed to list *v1.Service: services is forbidden: User "system:serviceaccount:openshift-machine-api:cluster-baremetal-operator" cannot list resource "services" in API group "" at the cluster scope
```

sadasu:
This is the from the logs : error from server (NotFound): secrets "metal3-ironic-inspector-password" not found

sadasu:
> Looks like maybe we broke something:
> 
> ```
> 2020-12-03T17:17:13.485448990Z E1203 17:17:13.485288       1 reflector.go:127] sigs.k8s.io/controller-runtime/pkg/cache/internal/informers_map.go:224: Failed to watch *v1.Service: failed to list *v1.Service: services is forbidden: User "system:serviceaccount:openshift-machine-api:cluster-baremetal-operator" cannot list resource "services" in API group "" at the cluster scope
> ```

This needs an rbac update.

sadasu:
We need to allow "services" to be watched at cluster scope by adding rbac here: https://github.com/openshift/cluster-baremetal-operator/pull/76/files#diff-7ad72d08a06fe28b03a15c84289b815f693443562d0c67e6f461c7e3f5ee0707R88

We might hit a similar issue with Daemonset rbacs too.

stbenjam:
@zaneb generate-check is failing

PR #77, opened 2020-11-25
Turn on auth for ironic json-rpc

honza:
The companion to https://github.com/openshift/machine-api-operator/pull/727

### DIFF
--- a/provisioning/baremetal_pod.go
+++ b/provisioning/baremetal_pod.go
@@ -32,6 +32,7 @@ const (
 	baremetalSharedVolume      = "metal3-shared"
 	metal3AuthRootDir          = "/auth"
 	ironicCredentialsVolume    = "metal3-ironic-basic-auth"
+	ironicrpcCredentialsVolume = "metal3-ironic-rpc-basic-auth"
 	inspectorCredentialsVolume = "metal3-inspector-basic-auth"
 	htpasswdEnvVar             = "HTTP_BASIC_HTPASSWD" // #nosec
 	mariadbPwdEnvVar           = "MARIADB_PASSWORD"    // #nosec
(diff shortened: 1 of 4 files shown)


hardys:
Do we need to go ahead with https://github.com/openshift/machine-api-operator/pull/727 now that CBO is enabled by default, or just make this change?

asalkeld:
this looks good to me..

dtantsur (approved):

provisioning/baremetal_pod.go
```diff
@@ -218,6 +238,7 @@ func createContainerMetal3BaremetalOperator(images *Images, config *metal3iov1al
 		ImagePullPolicy: "IfNotPresent",
 		VolumeMounts: []corev1.VolumeMount{
 			ironicCredentialsMount,
+			rpcCredentialsMount,
```
zaneb, line 241: baremetal-operator doesn't need RPC creds. It's an interface that is purely internal to Ironic.

sadasu:
Can we add any unit tests here?

provisioning/baremetal_pod.go
honza: Derek mentioned in the original PR that Ironic needs the creds to talk to itself?

dtantsur: Ironic Conductor, but not BMO.

honza: Done

sadasu:
I am ready to approve. Could you please add a unit test to baremetal_secrets_test.go here https://github.com/openshift/cluster-baremetal-operator/blob/master/provisioning/baremetal_secrets_test.go#L65, like we do for the other secrets?

honza:
> Could you please add a unit test to baremetal_secrets_test.go here https://github.com/openshift/cluster-baremetal-operator/blob/master/provisioning/baremetal_secrets_test.go#L65, like we do for the other secrets?

Done

zaneb:
It's so weird to me that every repo managed by Prow has different rules for e.g. whether /lgtm from an approver counts as approval. Anyway...
/approve

sadasu:

the same 5 e2e-agnostic tests failing. So, OK to override.

PR #80, opened 2020-11-30
Support TLS for ironic and inspector

dtantsur:
Note that host validation is disabled via environment variables
since we don't yet have a reliable way to learn the provisioning
IP(s) before starting ironic. This is unfortunate, but still
better than plain HTTP.


dtantsur:
Won't pass the CI until https://github.com/openshift/ironic-image/pull/127 and https://github.com/openshift/ironic-inspector-image/pull/51 merge.

zaneb:
There is a [service in OpenShift](https://github.com/openshift/service-ca-operator) that will automatically generate (and rotate!) a k8s Secret for a given Service resource. It _appears_ (but I have not confirmed) that the IP address will be added as a Subject Alternative Name, so it may be possible to create a [headless Service](https://kubernetes.io/docs/concepts/services-networking/service/#headless-services) that will allow us to use this mechanism for free. I'm not sure if we'd have most luck using a selector or going without a selector and creating an Endpoint explicitly, though I suspect the latter.

Failing that, at the very least we should try to use the [same library](https://github.com/openshift/library-go/blob/master/pkg/crypto/crypto.go#L762) methods to generate the certificates. We don't want to be responsible for custom crypto or FIPS compliance or any of that other fun stuff in the CBO.

dtantsur:
> We don't want to be responsible for custom crypto

To be clear, it's not custom crypto, I'm using only existing and well known primitives. Most of the code here is converting between formats. I'll look into your links, although at first glance I have no idea how to apply them in our situation.

dhellmann:
> > We don't want to be responsible for custom crypto
> 
> To be clear, it's not custom crypto, I'm using only existing and well known primitives. Most of the code here is converting between formats. I'll look into your links, although at first glance I have no idea how to apply them in our situation.

Yes, please take advantage of the platform features for managing certificates. We can find someone on the OpenShift team to help, if needed.

dtantsur:
> We can find someone on the OpenShift team to help, if needed.

Realistically, if you want to go down the service path, the amount of explanation I'll need is larger than the amount of work for somebody who understands k8s to actually implement this.

dtantsur:
/test e2e-metal-ipi

Trying to see if the whole thing at least works together

dtantsur:
@zaneb quoting the docs you linked:

> The generated certificate is only valid for the internal service DNS name <service.name>.<service.namespace>.svc, and are only valid for internal communications.

It looks like if we go down the service path, we'll have to disable host validation forever. Not sure how bad it is, but it does sound limiting. Or am I misreading it?

dtantsur:
The current version uses the OpenShift crypto modules, takes ProvisioningIP into account and refreshes expired certificates. I'm not sure it will pass the CI (too late here to wait), but any feedback is very welcome.

zaneb:
> @zaneb quoting the docs you linked:
> 
> > The generated certificate is only valid for the internal service DNS name <service.name>.<service.namespace>.svc, and are only valid for internal communications.
> 
> It looks like if we go down the service path, we'll have to disable host validation forever. Not sure how bad it is, but it does sound limiting. Or am I misreading it?

I'm assuming that note is correct in general but hides a lot of detail. What it's really saying is that you can't use this mechanism to when exposing public services with external DNS names.

If the IP addresses are added as Subject Alternative Names (and I haven't verified that, but I certainly see [some code that appears to do it](https://github.com/openshift/library-go/blob/master/pkg/crypto/crypto.go#L1000-L1005)) then if should work with verification using the IP and not just the internal DNS name. And while that only means the IP of the Service, if we use a headless Service and create the Endpoints ourselves then _maybe_ we can convince it to issue one for the provisioning VIP.

dtantsur:
@zaneb 

> If the IP addresses are added as Subject Alternative Names

The code you're linking to has nothing to do with that, it's a workaround for the case where IP addresses are passed, but we don't know if they are. The [certificate generation code](https://github.com/openshift/service-ca-operator/blob/e4e0b6b80f9bca5fc1bbd0914fee481a92df50ad/pkg/controller/servingcert/controller/secret_creating_controller.go#L333) seems to only include the service FQDN, nothing from endpoints.

I'd rather move forward with this patch, since the service approach is way beyond my expertise in k8s, and we risk getting blocked without any HTTPS at all while trying to find a perfect solution. If I understand it right, we can always swap the implementations later while keeping the same secret.

dhellmann:
> I'd rather move forward with this patch, since the service approach is way beyond my expertise in k8s, and we risk getting blocked without any HTTPS at all while trying to find a perfect solution. If I understand it right, we can always swap the implementations later while keeping the same secret.

I really really don't think we want to manage certificates ourself in any way. Definitely not in a way that's different from what the rest of the platform does.

zaneb:
Let's take a pause until we can figure out if there's a path toward a more automated way to handle this.
/hold

dtantsur:
@dhellmann 

> I really really don't think we want to manage certificates ourself in any way. Definitely not in a way that's different from what the rest of the platform does.

Are we fine with compromising security because of that? Unfortunately, the world of TLS is very browser-oriented, one cannot just disable host validation, you disable all validations. The traffic will be secure against passive sniffing, but not against active MitM.

provisioning/baremetal_secrets.go
```diff
@@ -168,3 +152,72 @@ password = %s
...
+
+		if changed {
+			_, err = client.Secrets(targetNamespace).Update(context.Background(), existing, metav1.UpdateOptions{})
+			return err
+		} else {
+			return nil
+		}
+	}
+
+	if !apierrors.IsNotFound(err) {
```
dhellmann, line 194: In go it is more common to exit early when there is an error. Moving this block up to line 159 would be consistent with that style advice.

provisioning/baremetal_secrets.go
```diff
@@ -168,3 +152,72 @@ password = %s
...
+			}
+			existing.StringData[tlsCertificateKey] = cert.certificate
+			existing.StringData[tlsPrivateKeyKey] = cert.privateKey
+			changed = true
+		}
+
+		if changed {
+			_, err = client.Secrets(targetNamespace).Update(context.Background(), existing, metav1.UpdateOptions{})
+			return err
+		} else {
```
dhellmann, line 189: This else is not needed.

provisioning/baremetal_secrets.go
```diff
@@ -168,3 +152,72 @@ password = %s
...
+// CreateTlsSecret creates a Secret for the Ironic and Inspector TLS
+func CreateTlsSecret(client coreclientv1.SecretsGetter, targetNamespace string, baremetalConfig *metal3iov1alpha1.Provisioning, scheme *runtime.Scheme) error {
+	existing, err := client.Secrets(targetNamespace).Get(context.Background(), tlsSecretName, metav1.GetOptions{})
+	if err == nil {
+		changed := false
+
+		if len(existing.ObjectMeta.OwnerReferences) == 0 {
+			err = controllerutil.SetControllerReference(baremetalConfig, existing, scheme)
+			if err != nil {
+				return err
```
dhellmann, line 165:
```suggestion
				return errors.Wrap(err, fmt.Sprintf("failed to set controller reference of Secret %s", tlsSecretName))
```

provisioning/baremetal_secrets.go
```diff
@@ -168,3 +152,72 @@ password = %s
...
+			}
+			changed = true
+		}
+
+		existingCert := existing.StringData[tlsCertificateKey]
+		expired, err := isTlsCertificateExpired(existingCert)
+		if err != nil {
+			return err
```
dhellmann, line 173:
```suggestion
			return errors.Wrap(err, "failed to determine expiration date of TLS certificate")
```

provisioning/baremetal_secrets.go
```diff
@@ -168,3 +152,72 @@ password = %s
...
+		}
+
+		if expired {
+			cert, err := generateTlsCertificate(baremetalConfig.Spec.ProvisioningIP)
+			if err != nil {
+				return err
```
dhellmann, line 179:
```suggestion
				return errors.Wrap(err, "failed to generate new TLS certificate")
```

provisioning/baremetal_secrets.go
```diff
@@ -168,3 +152,72 @@ password = %s
 	_, err = client.Secrets(targetNamespace).Create(context.Background(), secret, metav1.CreateOptions{})
 	return err
 }
+
+// CreateTlsSecret creates a Secret for the Ironic and Inspector TLS
+func CreateTlsSecret(client coreclientv1.SecretsGetter, targetNamespace string, baremetalConfig *metal3iov1alpha1.Provisioning, scheme *runtime.Scheme) error {
```
dhellmann, line 157: This function also rotates the certificate, which isn't obvious from the name.

provisioning/baremetal_secrets.go
```diff
@@ -168,3 +152,72 @@ password = %s
...
+	existing, err := client.Secrets(targetNamespace).Get(context.Background(), tlsSecretName, metav1.GetOptions{})
+	if err == nil {
```
dhellmann, line 159: The tests do not appear to cover this part of the logic.

provisioning/baremetal_crypto.go
```diff
@@ -0,0 +1,104 @@
...
+		if err != nil {
+			return "", err
+		}
+		buf[i] = chars[c.Uint64()]
+	}
+	return string(buf), nil
+}
+
+func generateTlsCertificate(provisioningIP string) (TlsCertificate, error) {
+	caConfig, err := crypto.MakeSelfSignedCAConfig("metal3-ironic", tlsExpirationDays)
```
dhellmann, line 57: The cluster has a CA. Should we use that to sign the certificate?

dtantsur: Interesting, how do I access that?

dhellmann: It doesn't look like there's a direct way, and the API that is exposed via k8s appears focused on client access certificates rather than general server certificates. That API requires an "approver" to work as part of the process, which we may not have when we need our certificate. Let's stick with self-signed for now and switch if we learn that we can.

dhellmann:
/hold clear

After doing more research today, I'm leaning towards accepting that we have to manage this certificate ourselves. The service-ca-operator doesn't give us certificates with the IP we need. A CertificateSigningRequest is really more for client authentication certificates than server certs like we need.

Let's remove the hold and work the code into shape to land.

dtantsur:
Added your suggestions, will work on other comments soon.

dtantsur:
/retest

These results are from Monday, wtf?

dtantsur:
/retest

old version of dev-scripts cached?

asalkeld:
have a look here: https://github.com/openshift/cluster-baremetal-operator/pull/87/files#diff-3029d4e11509bb8f6b821ac0254eb0bb71a7148a44426f111f9f97ada97717d2R7 
https://github.com/openshift/cluster-baremetal-operator/pull/87/files#diff-97451f7e9872660ef7c689e92171cf764d18dc86fe1958794897c5847d7cd3d1R7
we are using the cert injection for the webhook PR

dtantsur:
@asalkeld please see the discussion above: using this will rule out a future possibility to use secure TLS (with host validation).

dtantsur:
/retest

Packet playing up now

elfosardo (approved):

dtantsur:
@zaneb @stbenjam could you please take another look?

dtantsur:
/retest

all important jobs have passed

stbenjam:
Just want to test disconnected still works.

/test e2e-metal-ipi-virtualmedia

dtantsur:
/test e2e-metal-ipi-virtualmedia

> failed to fetch service account builder: the server has received too many requests

dtantsur:
@stbenjam what's the status of this job? seems quite broken to me.

stbenjam:
/test e2e-metal-ipi-virtualmedia

The job itself is fine, "pod pending timeout" indicates CI cluster issues

stbenjam:
The reason I want to test virtualmedia is Ironic builds the URL to tell the BMC where to download virtual media itself, and during setup of Ironic, it picks the first IP it finds. CBO and the cluster components only ever talk to `localhost` since 4.7, so you're pretty much only ever generating the certificate only for `localhost`. 

Is the certificate validation off in IPA too?

dtantsur:
> Is the certificate validation off in IPA too?

Yep. We'll need to look into certificate validation later.

stbenjam:
/lgtm

/assign @sadasu

Could you have a look?

sadasu (approved):

PR #82, opened 2020-12-01
baremetal_pod: add log watcher containers

stbenjam:
This adds the Ironic and Inspector log watching containers to the Metal3
deployment.  These are used to make the logs that baremetal hosts upload
to Ironic visible and collectable by OCP tools like `oc` and
`must-gather`.

### DIFF
--- a/provisioning/baremetal_pod.go
+++ b/provisioning/baremetal_pod.go
@@ -210,7 +210,9 @@ func newMetal3Containers(images *Images, config *metal3iov1alpha1.ProvisioningSp
 		createContainerMetal3Mariadb(images),
 		createContainerMetal3Httpd(images, config),
 		createContainerMetal3IronicConductor(images, config),
+		createContainerIronicInspectorRamdiskLogs(images),
 		createContainerMetal3IronicApi(images, config),
+		createContainerIronicDeployRamdiskLogs(images),
 		createContainerMetal3IronicInspector(images, config),
 	}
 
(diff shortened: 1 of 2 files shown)


kirankt (approved):

kirankt:
One nit is that lets not link intranet links in the description/comments. Any links posted should be visible for everyone using github.

stbenjam:
> One nit is that lets not link intranet links in the description/comments. Any links posted should be visible for everyone using github.

Yea sorry, I was just experimenting with the plug-in that does the automatic linking.

The jira instance is now public although our project is not yet. Once it is, we should be putting jira references in GitHub PR's so it makes the link properly like we do with BZ's.

FWIw, I didn't make the link. The bot did it!

stbenjam:
https://github.com/openshift/installer/pull/4437 does the bootstrap, we'll have to reconcile the names so they're consistent across CBO and the bootstrap.

asalkeld (commented):
needs rebase..

hardys:
> [openshift/installer#4437](https://github.com/openshift/installer/pull/4437) does the bootstrap, we'll have to reconcile the names so they're consistent across CBO and the bootstrap.

Did we reach a consensus on the names?  I'm happy to update the installer PR if needed but I wasn't sure if we agreed on what the names should be :)

I went with ironic-deploy-ramdisk-logs ironic-inspector-ramdisk-logs - it's a bit verbose but describes the purpose/content of the container - I'm happy to go with whatever folks prefer though :)

stbenjam:
Updated to match the installer names. Also as discussed on Slack, I didn't use the metal3 prefix on these containers as it's already in the pod name. Perhaps we can update the others in the future.

dtantsur (approved):

PR #83, opened 2020-12-01
The bmh relatedobject name should be an empty string.

kirankt:
The BMH relatedobject name should be an empty string. The relatedobject will be referenced like so:

Name.Resource.Group 

which can result in referencing an incorrect object.


zaneb:
/lgtm
Thanks Kiran!

PR #85, opened 2020-12-02
Bug 1903717: Change default Deployment selector and handle downgrades

zaneb:
When the metal3 Deployment was managed by the MAO, it used the Pod
labels `api=clusterapi` and `k8s-app=controller` to identify Metal³ as part
of the cluster API, and the `k8s-app=controller` label as the selector to
associate Pods with the Deployment. This was not only confusing, but
actually erroneous as the MAO also uses the same label as a selector for
its own Deployments of the Machine API controllers - this can
theoretically cause the separate Deployments to each try to manage both
sets of Pods matching the selector.

However, we cannot update the selector for an existing Deployment,
because that field is immutable. We also cannot (yet) delete the
existing Deployment because if, during an upgrade from OpenShift 4.6,
the MAO sees that it is missing then it will also try to recreate it.
The labels in the Pod template need not be identical to the selector,
but the selector must match the labels in the Pod template.

The `api=clusterapi` label is removed from the Pod template to reduce
confusion, since it is not targeted by the selector anyway.

The Deployment gets the additional label
`baremetal.openshift.io/cluster-baremetal-operator=metal3-state` to match
the Pods.

This change alters the labels and selectors for new Deployments only, as
follows:

* The Deployment's selector changes to `k8s-app=metal3` plus
  `baremetal.openshift.io/cluster-baremetal-operator=metal3-state`
* In the Pod template, the `k8s-app` label changes to `k8s-app=metal3`
* The Deployment's label changes to `k8s-app=metal3` to match the other
  resources

Once the Deployment exists, we keep the existing selector (since
attempting to change the selector would result in an error). The k8s-app
label in the Pod template always takes the same value as in the
selector, so that the Pods are always selectable (again, violating this
condition would cause an error).

Since this works equally well with both old an new selectors, it will
continue to work after a downgrade from some future release where the
selectors are changed and/or existing Deployments with legacy removed.

### DIFF
--- a/controllers/provisioning_controller.go
+++ b/controllers/provisioning_controller.go
@@ -230,12 +230,12 @@ func (r *ProvisioningReconciler) Reconcile(req ctrl.Request) (ctrl.Result, error
 	}
 
 	// If Metal3 Deployment already exists and managed by MAO, do nothing.
-	exists, err := provisioning.MAOMetal3DeploymentExists(r.KubeClient.AppsV1(), ComponentNamespace)
+	metal3DeploymentSelector, maoOwned, err := provisioning.CheckExistingMetal3Deployment(r.KubeClient.AppsV1(), ComponentNamespace)
 	if err != nil && !apierrors.IsNotFound(err) {
-		return ctrl.Result{}, errors.Wrap(err, "failed to find existing Metal3 Deployment")
+		return ctrl.Result{}, errors.Wrap(err, "failed to check for existing Metal3 Deployment")
 	}
 
-	if exists {
+	if maoOwned {
 		r.Log.V(1).Info("metal3 deployment already exists")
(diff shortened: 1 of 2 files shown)


zaneb:
Once additional pods with `k8s-app=metal3` are added in #76, we will need to add additional labels to the selector for new Deployments, so leaving as draft until then.

sadasu:
/lgtm
Solution looks neat. I had to go through it a couple of times to understand it fully.
So, as long as we do not change the LabelSelector of an existing Deployment, we would not have to delete the existing deployment. Thanks for taking the time to understand the Label nuances.

sadasu:
/lgtm cancel 
removing lgtm in case someone else wants to take a look.

sadasu (commented):
We still need some changes from https://github.com/openshift/cluster-baremetal-operator/pull/65 for monitoring the state of the deployment to set the CO.

PR #86, opened 2020-12-03
Set IPA URLs using API VIP

zaneb:
Store the IPA kernel and initrd in the image cache, pass the URLs for
them to the baremetal-operator that resolve to the API VIP. This ensures
that ironic can fetch these files whether or not there is a provisioning
VIP.

This is a follow-up to #73.


provisioning/baremetal_config.go
```diff
@@ -142,20 +143,14 @@ func getProvisioningIPCIDR(config *metal3iov1alpha1.ProvisioningSpec) *string {
...
 }
 
-func getDeployKernelUrl(config *metal3iov1alpha1.ProvisioningSpec) *string {
-	if config.ProvisioningIP != "" {
-		deployKernelUrl := fmt.Sprintf("http://%s/%s", net.JoinHostPort(config.ProvisioningIP, baremetalHttpPort), baremetalKernelUrlSubPath)
-		return &deployKernelUrl
-	}
-	return nil
+func getDeployKernelUrl() *string {
+	deployKernelUrl := fmt.Sprintf("http://%s/%s", net.JoinHostPort("api", strconv.Itoa(imageCachePort)), baremetalKernelUrlSubPath)
```
stbenjam, line 147: Should we always access these over the API VIP? I think users would expect that if there's a provisioning network, the traffic would happen over it.

stbenjam: It could still hit the image cache port. I'm just imagining that these IPA images are a few hundred megs, and a rapid scale out of a few dozen machines is going to use a lot of bandwidth on the machine networks.

stbenjam:
It works (tested with https://github.com/openshift-metal3/dev-scripts/pull/1169)! Just one comment above. But, this is beautiful. 

**Provisioning CR has no IP or CIDR:**

```yaml
spec:
  provisioningNetwork: Disabled
  provisioningOSDownloadURL: http://192.168.111.1/images/rhcos-47.82.202010211043-0-openstack.x86_64.qcow2.gz?sha256=3072921dfd2d7d405ba92466a09a7b2772a0b4aa684beaaf03f45d3173d879cc
```

**And everybody's up:**

```
$ oc get bmh -A
NAMESPACE               NAME              STATUS   PROVISIONING STATUS      CONSUMER                      BMC                                                                                                      HARDWARE PROFILE   ONLINE   ERROR
openshift-machine-api   ostest-master-0   OK       externally provisioned   ostest-vpp98-master-0         redfish-virtualmedia+http://192.168.111.1:8000/redfish/v1/Systems/e93751dc-2b5c-45ba-8a48-25b9b0ece9a4                      true     
openshift-machine-api   ostest-master-1   OK       externally provisioned   ostest-vpp98-master-1         redfish-virtualmedia+http://192.168.111.1:8000/redfish/v1/Systems/a5a0d806-ce4c-4f2a-8f88-c56536e23ea3                      true     
openshift-machine-api   ostest-master-2   OK       externally provisioned   ostest-vpp98-master-2         redfish-virtualmedia+http://192.168.111.1:8000/redfish/v1/Systems/ea77b1d4-1425-4997-9840-214f51a775ae                      true     
openshift-machine-api   ostest-worker-0   OK       provisioned              ostest-vpp98-worker-0-52spm   redfish-virtualmedia+http://192.168.111.1:8000/redfish/v1/Systems/430c50de-d9e0-4bb5-8def-bc000c655462   unknown            true     
openshift-machine-api   ostest-worker-1   OK       provisioned              ostest-vpp98-worker-0-s89nl   redfish-virtualmedia+http://192.168.111.1:8000/redfish/v1/Systems/159858df-2cc0-444c-badf-a9d5f881a6e2   unknown            true   
```

```
$ oc get nodes
NAME       STATUS   ROLES    AGE    VERSION
master-0   Ready    master   25m    v1.19.2+ad738ba
master-1   Ready    master   25m    v1.19.2+ad738ba
master-2   Ready    master   25m    v1.19.2+ad738ba
worker-0   Ready    worker   89s    v1.19.2+ad738ba
worker-1   Ready    worker   113s   v1.19.2+ad738ba
```

stbenjam:
From the e2e-metal-ipi [must-gather](https://prow.ci.openshift.org/view/gs/origin-ci-test/pr-logs/pull/openshift_cluster-baremetal-operator/86/pull-ci-openshift-cluster-baremetal-operator-master-e2e-metal-ipi/1334662341442146304):

```
Validation of image href http://api:6181/images/ironic-python-agent.kernel failed,
reason: HTTPConnectionPool(host='api', port=6181): Max retries exceeded with url: /images/ironic-python-agent.kernel (Caused by
NewConnectionError('<urllib3.connection.HTTPConnection object at 0x7f42f0f7d668>: Failed to establish a new connection: [Errno 111] ECONNREFUSED',))
```

provisioning/baremetal_config.go
zaneb: Hmm, I was working on the theory that this URL was used for traffic only between Ironic and httpd, so all internal to the box anyway. If Ironic is passing this URLs to BMCs then it definitely needs to use the provisioning IP when it is available.

stbenjam: The API VIP isn't guaranteed to be on the same box as Ironic, though.

zaneb: Actually if that's the case then this won't work at all, and we'll need to put an IP address always, since the DNS is only resolvable by ironic.
And not even that, judging by the failure in the gate.

stbenjam: So could we use the ProvisioningIP when we have one, and if not use the API VIP?

zaneb:
The test failure appears to be due to a startup ordering problem. On the node with the metal3 Pod, the image cache Pod starts up quickly, but on the other 2 it takes some time because it must download (and recompresses) the OS image from the metal3 Pod (and can only start this once the metal3 Pod is up). Until that point httpd in the image cache pods can't start. (It starts about 3 minutes later in the tests.)
The BMO doesn't wait for the image cache, so it goes ahead and starts inspecting, resulting in the Ironic error @stbenjam pasted above.
Finally, the BMO does not notice this error and fail inspection (which would hopefully be retried), it just waits forever for the inspection results to show up.

zaneb:
This should be fixed by metal3-io/baremetal-operator#737, but it might make sense to not declare the operator ready until the URLs are reachable also.

provisioning/baremetal_config.go
zaneb: The DNS wasn't the problem, but actually we could just always use localhost:6181. Then it would always hit the local image cache even when there's no provisioning VIP, so no E-W traffic in any configuration. And as a bonus, it will likely work without waiting for BMO fixes because the image cache comes up more or less instantly on the host with the metal3 Pod.

stbenjam: Ok I think this makes sense then. PXE clients get the URL from the ram disk from the TFTP server. 

Localhost should probably be ok here

zaneb:
This version just uses localhost:6181 (i.e. the image-cache Pod on whichever machine the metal3 Pod happens to be running on). This minimises the timing issue, since the image cache on this machine shares storage with the first-level cache in the metal3 Pod and thus comes up very quickly. The bug in the bmo will be nevertheless fixed by openshift/baremetal-operator#114.

Using localhost as the address also means that we'll never need traffic to cross the network to download the IPA initrd+kernel.

dtantsur:
Late to the party, but since the images are on a shared volume, you could also use `file://` URLs for them, excluding HTTP completely.

PR #84, opened 2020-12-02
Use the webhook for CR validation

asalkeld:
This uses the webhook for more CR validation (see @imain's commits that add the validation).

PR #87, opened 2020-12-04
[minimal] Create a webhook for CR validation

asalkeld:
this is a simplified version of https://github.com/openshift/cluster-baremetal-operator/pull/84 at @sadasu 's request.
/cc @sadasu @imain 

The point is to get the mechanism merged, then follow up with the remaining validations in #84 

So this uses service.beta.openshift.io for certificate injection. How ever the "service-ca" operator that implements this checks
the platform-type and if it is not ready, doesn't work properly.. So a chicken/egg problem. 
I solved this by initially not registering the webhook until the "service-ca" operator is up. 
This is done by just exiting and getting restarted.
I have also updated the contorller-runtime as there was a webhook serialization bug.

for reviewers: this is easier to review by commit


config/profiles/default/kustomization.yaml
```diff
@@ -1,3 +1,5 @@
+namespace: openshift-machine-api
+namePrefix: cluster-baremetal-
```
asalkeld, line 2: @sadasu this adds 'cluster-baremetal-' to all resource names

api/v1alpha1/provisioning_webhook.go
```diff
@@ -0,0 +1,66 @@
...
+// ValidateUpdate implements webhook.Validator so a webhook will be registered for the type
+func (r *Provisioning) ValidateUpdate(old runtime.Object) error {
+	provisioninglog.Info("validate update", "name", r.Name)
+
+	// TODO(imain): fill in your validation logic upon object creation.
+	return nil
+}
+
+// ValidateDelete implements webhook.Validator so a webhook will be registered for the type
+func (r *Provisioning) ValidateDelete() error {
```
sadasu, line 63: Do we need to add the ProvisioningFinalizer here or in the ValidateCreate() method?

asalkeld: this is purely a validating webhook, not a mutating webhook. I have had a look at mao, and it has both a validating webhook and the finalizer within the controller and it somehow works, so I think there is just something small I am missing.

asalkeld:
It looks like some early race, as there is no provisioning CR.
```
Provisioning CR not found on BareMetal Platform; marking operator as available
```
It could be that the validating webhook is not available when the CR is created, and it errors resulting in no CR.
cc @imain

asalkeld:
next problem
```
level=info msg=Cluster operator baremetal Progressing is True with DeployTimedOut: Unable to delete a metal3 resource on Provisioning CR deletion
level=error msg=Cluster operator baremetal Degraded is True with DeployTimedOut: failed to update Provisioning CR with its finalizer: Internal error occurred: admission plugin "ValidatingAdmissionWebhook" failed to complete validation in 13s 
```
This is cbo it's self updating the CR and that seems to cause the webhook to timeout

sadasu:
> This is cbo it's self updating the CR and that seems to cause the webhook to timeout

Is this "self-updating" to do with CBO removing the finalizer on the CR?

asalkeld:
> > This is cbo it's self updating the CR and that seems to cause the webhook to timeout
> 
> Is this "self-updating" to do with CBO removing the finalizer on the CR?

@sadasu yes, I am guessing that the webhook is not ready when controller runs through the first time.

config/rbac/role.yaml
```diff
@@ -39,6 +39,17 @@ rules:
   - patch
   - update
   - watch
+- apiGroups:
+  - admissionregistration.k8s.io
+  resources:
+  - ValidatingWebhookConfiguration
+  verbs:
+  - create
```
asalkeld, line 47: this is so we can create the webhookConfig dynamically

controllers/provisioning_controller.go
```diff
@@ -186,11 +187,19 @@ func (r *ProvisioningReconciler) Reconcile(req ctrl.Request) (ctrl.Result, error
...
-		err = r.updateCOStatus(ReasonDeployTimedOut, err.Error(), "Unable to delete a metal3 resource on Provisioning CR deletion")
-		if err != nil {
-			return ctrl.Result{}, fmt.Errorf("unable to put %q ClusterOperator in Degraded state: %w", clusterOperatorName, err)
+		if deleted {
+			coErr := r.updateCOStatus(ReasonDeployTimedOut, err.Error(), "Unable to delete a metal3 resource on Provisioning CR deletion")
+			if coErr != nil {
+				return ctrl.Result{}, fmt.Errorf("unable to put %q ClusterOperator in Degraded state: %w", clusterOperatorName, coErr)
+			}
+			return ctrl.Result{}, err
+		} else {
```
asalkeld, line 196: this is so we don't have a really confusing error if the finalizer add false

config/cluster-baremetal-operator/cluster-baremetal-operator.yaml
```diff
@@ -1,7 +1,7 @@
 apiVersion: apps/v1
 kind: Deployment
 metadata:
-  name: cluster-baremetal-operator
+  name: operator
```
sadasu, line 4: Is this change correct?

asalkeld: yes, see https://github.com/openshift/cluster-baremetal-operator/pull/87/files#diff-f557e464031f4695b338c76660478a714639a766096c14b2fc4ebb385afe11e2R2

asalkeld:
more flaky tests
/test e2e-metal-ipi-upgrade

asalkeld:
more flaky tests
/test e2e-metal-ipi-upgrade

PR #88, opened 2020-12-07
Bug 1905307: cluster-baremetal-operator: add provisioning CR as related object

stbenjam:
Without having the provisioning CRD set as a related resource, we aren't
collecting it in a must-gather.

After applying this PR and running a must-gather, we get the
provisioning CR in the must-gather:

```
./cluster-scoped-resources/metal3.io/provisionings/provisioning-configuration.yaml
```


stbenjam:
/retitle Bug 1905307: cluster-baremetal-operator: add provisioning CR as related object

asalkeld (commented):
seems good to me

PR #89, opened 2020-12-09
Bug 1866901: Don't do rolling updates of metal3 Deployment

zaneb:
Up to now we have relied on the ironic-static-ip-manager containers
managing the Provisioning IP to ensure that only one metal3 Pod is
started at a time. However, now that we support configurations without
such an IP, we should take other measures to ensure that we don't create
multiple inconsistent copies of the state (e.g. ironic database).

Changing the Deployment Strategy from the default rolling update to
recreate means that upon any changes to the Pod template, the existing
ReplicaSet will be deleted before an updated one is created.

### DIFF
--- a/provisioning/baremetal_pod.go
+++ b/provisioning/baremetal_pod.go
@@ -601,6 +601,9 @@ func NewMetal3Deployment(targetNamespace string, images *Images, config *metal3i
 			Replicas: pointer.Int32Ptr(1),
 			Selector: selector,
 			Template: *template,
+			Strategy: appsv1.DeploymentStrategy{
+				Type: appsv1.RecreateDeploymentStrategyType,
+			},
 		},
 	}
 }


dhellmann:
/retitle Bug 1866901: Don't do rolling updates of metal3 Deployment

sadasu:
Thanks for taking care of this!

sadasu:
> > Is this applicable to the DaemonSet also?
> 
> At the time I wrote the image-cache I thought that we would want to upgrade as quickly as possible and didn't care about uniqueness, so I set it to rolling update with 100% unavailability allowed (in theory I think this should mean we delete the old pods and create the new ones at the same time).
> 
> OTOH we have to bind to a fixed host port, which will fail if the old one is still running. But it should sort itself out eventually, and in the case where we have to download a new image we want it to start as quickly as possible.

I agree with this reasoning.

PR #91, opened 2020-12-18
Bug 1913716: Use existing libraries instead of rolling our own

asalkeld:
@sadasu I hope you don't mind me tweaking code you have just written
- This only adds one dependency (stretchr/stew/slice - note we already use stretchr/testify)
- the other functions we already have vendored

/cc @sadasu


sadasu:
Thanks @asalkeld ! Really great set of updates!

sadasu:
/retitle Bug 1913716: Use existing libraries instead of rolling our own

sadasu (approved):

PR #93, opened 2021-01-07
Bug 1909527: Update Labels on the Pod TemplateSpec of the metal3 Deployment

sadasu:
Add "api" label as one of the Labels in metal3 Deployment Pod
Template Spec because the metal3 Deployment created by MAO
uses LabelSelectors that are looking for this label on the Pod.

### DIFF
--- a/provisioning/baremetal_pod.go
+++ b/provisioning/baremetal_pod.go
@@ -526,7 +526,7 @@ func createContainerMetal3StaticIpManager(images *Images, config *metal3iov1alph
 	return container
 }
 
-func newMetal3PodTemplateSpec(images *Images, config *metal3iov1alpha1.ProvisioningSpec, k8sAppLabel string) *corev1.PodTemplateSpec {
+func newMetal3PodTemplateSpec(images *Images, config *metal3iov1alpha1.ProvisioningSpec, labels *map[string]string) *corev1.PodTemplateSpec {
 	initContainers := newMetal3InitContainers(images, config)
 	containers := newMetal3Containers(images, config)
 	tolerations := []corev1.Toleration{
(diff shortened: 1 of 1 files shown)


provisioning/baremetal_pod.go
```diff
@@ -586,13 +584,21 @@ func newMetal3Deployment(targetNamespace string, images *Images, config *metal3i
...
 			k8sAppLabel = v
-			break
 		}
+		if k == apiLabelKey {
+			apiLabelValue = v
+		}
+	}
+	podSpecLabels := map[string]string{
+		"k8s-app":    k8sAppLabel,
+		apiLabelKey:  apiLabelValue,
```
zaneb, line 598: This means that we will always have an `api` label (with an empty value when we are creating a new Deployment). I think it would be better if we just didn't add that label at all unless it is necessary for the selector.

provisioning/baremetal_pod.go
```diff
@@ -46,6 +46,7 @@ const (
 	mariadbPwdEnvVar           = "MARIADB_PASSWORD"    // #nosec
 	cboOwnedAnnotation         = "baremetal.openshift.io/owned"
 	cboLabelName               = "baremetal.openshift.io/cluster-baremetal-operator"
+	apiLabelKey                = "api"
```
zaneb, line 49: I'm normally a fan of named constants, but in this case it's only used in one place and is for handling legacy support (i.e. not something we ever want to use again). On balance, I think this serves only to make it more difficult to tell what is going on when reading the code.

provisioning/baremetal_pod.go
sadasu: Ok. Cool!

provisioning/baremetal_pod.go
sadasu: Honestly, it is hard to predict which way a reviewer would go. I can remove this constant.

PR #96, opened 2021-01-15
Bug 1900322: Add Toleration OP Exists to metal3 pod

zerodayz:
The metal3 pod's toleration currently matches on
exact value matches only.

This can be confusing for some people as the vast
majority of our pods we use operator Exists, mean
any value can be specified. This

  tolerations:
  - effect: NoSchedule
    key: node-role.kubernetes.io/master

This patch propose solution to add operator: Exists

  - effect: NoSchedule
    key: node-role.kubernetes.io/master
    operator: Exists

This will make it easier for everyone to use this operator.

Closes BZ: #1900322


asalkeld (approved):
/lgtm
/approve

PR #106, opened 2021-02-12
Allow BMO to watch all Namespaces

sadasu:
When the metal3 deployment is created, the BMO is not provided with
the WATCH_NAMESPACE env var which allows it to watch all Namespaces.
Also, change the rbacs to reflect the fact that BMH objects now can
exist in multiple namespaces and not just the openshift-machine-api
namespace.

### DIFF
--- a/api/v1alpha1/provisioning_types.go
+++ b/api/v1alpha1/provisioning_types.go
@@ -99,6 +99,14 @@ type ProvisioningSpec struct {
 	// accessible from the machine networks. User should provide two IPs on
 	// the external network that would be used for provisioning services.
 	ProvisioningNetwork ProvisioningNetwork `json:"provisioningNetwork,omitempty"`
+
+	// WatchAllNamespaces provides a way to explicitly allow use of this
+	// Provisioning configuration across all Namespaces. It is an
+	// optional configuration which defaults to false and in that state
+	// will be used to provision baremetal hosts in only the
+	// openshift-machine-api namespace. When set to true, this provisioning
+	// configuration would be used for baremetal hosts across all namespaces.
+	WatchAllNamespaces bool `json:"watchAllNamespaces,omitempty"`
 }
 
 // ProvisioningStatus defines the observed state of Provisioning
--- a/provisioning/baremetal_pod.go
+++ b/provisioning/baremetal_pod.go
@@ -262,6 +262,21 @@ func newMetal3Containers(images *Images, config *metal3iov1alpha1.ProvisioningSp
 	return containers
 }
 
+func getWatchNamespace(config *metal3iov1alpha1.ProvisioningSpec) corev1.EnvVar {
+	if config.WatchAllNamespaces {
+		return corev1.EnvVar{}
+	} else {
+		return corev1.EnvVar{
+			Name: "WATCH_NAMESPACE",
+			ValueFrom: &corev1.EnvVarSource{
+				FieldRef: &corev1.ObjectFieldSelector{
+					FieldPath: "metadata.namespace",
+				},
+			},
+		}
+	}
+}
+
 func createContainerMetal3BaremetalOperator(images *Images, config *metal3iov1alpha1.ProvisioningSpec) corev1.Container {
 	container := corev1.Container{
 		Name:  "metal3-baremetal-operator",
@@ -280,14 +295,7 @@ func createContainerMetal3BaremetalOperator(images *Images, config *metal3iov1al
 			inspectorCredentialsMount,
 		},
 		Env: []corev1.EnvVar{
-			{
-				Name: "WATCH_NAMESPACE",
-				ValueFrom: &corev1.EnvVarSource{
-					FieldRef: &corev1.ObjectFieldSelector{
-						FieldPath: "metadata.namespace",
-					},
-				},
-			},
+			getWatchNamespace(config),
 			{
 				Name: "POD_NAMESPACE",
 				ValueFrom: &corev1.EnvVarSource{
--- a/config/crd/bases/metal3.io_provisionings.yaml
+++ b/config/crd/bases/metal3.io_provisionings.yaml
@@ -56,6 +56,9 @@ spec:
               provisioningOSDownloadURL:
                 description: ProvisioningOSDownloadURL is the location from which the OS Image used to boot baremetal host machines can be downloaded by the metal3 cluster.
                 type: string
+              watchAllNamespaces:
+                description: WatchAllNamespaces provides a way to explicitly allow use of this Provisioning configuration across all Namespaces. It is an optional configuration which defaults to false and in that state will be used to provision baremetal hosts in only the openshift-machine-api namespace. When set to true, this provisioning configuration would be used for baremetal hosts across all namespaces.
+                type: boolean
             type: object
           status:
             description: ProvisioningStatus defines the observed state of Provisioning
--- a/config/rbac/role.yaml
+++ b/config/rbac/role.yaml
@@ -94,6 +94,23 @@ rules:
 - apiGroups:
   - metal3.io
   resources:
+  - baremetalhosts
+  verbs:
+  - get
+  - list
+  - patch
+  - update
+  - watch
+- apiGroups:
+  - metal3.io
+  resources:
+  - baremetalhosts/finalizers
+  - baremetalhosts/status
+  verbs:
+  - update
+- apiGroups:
+  - metal3.io
+  resources:
   - provisionings
   verbs:
   - create
@@ -198,23 +215,6 @@ rules:
   - patch
   - update
   - watch
-- apiGroups:
-  - metal3.io
-  resources:
-  - baremetalhosts
-  verbs:
-  - get
-  - list
-  - patch
-  - update
-  - watch
-- apiGroups:
-  - metal3.io
-  resources:
-  - baremetalhosts/finalizers
-  - baremetalhosts/status
-  verbs:
-  - update
 - apiGroups:
   - monitoring.coreos.com
   resources:
--- a/controllers/provisioning_controller.go
+++ b/controllers/provisioning_controller.go
@@ -67,8 +67,6 @@ type ensureFunc func(*provisioning.ProvisioningInfo) (bool, error)
 
 // +kubebuilder:rbac:namespace=openshift-machine-api,groups="",resources=configmaps,verbs=get;list;watch;create;update;patch;delete
 // +kubebuilder:rbac:namespace=openshift-machine-api,groups="",resources=secrets,verbs=get;list;watch;create;update;patch;delete
-// +kubebuilder:rbac:namespace=openshift-machine-api,groups=metal3.io,resources=baremetalhosts,verbs=get;list;watch;update;patch
-// +kubebuilder:rbac:namespace=openshift-machine-api,groups=metal3.io,resources=baremetalhosts/status;baremetalhosts/finalizers,verbs=update
 // +kubebuilder:rbac:namespace=openshift-machine-api,groups=security.openshift.io,resources=securitycontextconstraints,verbs=use
 // +kubebuilder:rbac:namespace=openshift-machine-api,groups=apps,resources=deployments,verbs=get;list;watch;create;update;patch;delete
 // +kubebuilder:rbac:namespace=openshift-machine-api,groups=apps,resources=daemonsets,verbs=get;list;watch;create;update;patch;delete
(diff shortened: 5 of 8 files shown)


asalkeld:
/retest
seems ok to me...

provisioning/baremetal_pod.go
```diff
@@ -280,14 +280,6 @@ func createContainerMetal3BaremetalOperator(images *Images, config *metal3iov1al
 			inspectorCredentialsMount,
 		},
 		Env: []corev1.EnvVar{
-			{
-				Name: "WATCH_NAMESPACE",
```
hardys, line 284: Are we Ok with unconditionally watching all namespaces, or should we have a provisioning CR option to enable it just for the hub-cluster use-case?

@dhellmann I recall we discussed both options before - in the regular IPI case where BMH resources only exist in the openshift-machine-api namespace, will there be any overhead to this being unconditional?

dhellmann: I don't think there would be significant overhead in always watching all namespaces in a normal cluster. There's a kubernetes API to list all resources of a type, regardless of the namespace, so I would expect the client in the controller-runtime to use that instead of the API for listing resources within a namespace.

There could be some user confusion if a user creates a host outside of the `openshift-machine-api`  namespace and it is inspected but cannot be used for provisioning because CAPBM won't see it.

hardys: > There could be some user confusion if a user creates a host outside of the openshift-machine-api namespace and it is inspected but cannot be used for provisioning because CAPBM won't see it.

I think that will be the next step, e.g look at the work being done in the machine-api team to enable machines in a multi-namespace environment, IIUC the current plan is a machine-controller (and thus CAPBM) per-namespace, so we'll need to sync up with folks working on that and figure out how to make it work on baremetal (current testing is only on AWS I believe)

@JoelSpeed is ^^ accurate, and can you point to any docs/scripts etc that we can refer to re running multi-namespace machine controllers (I'm thinking it's probably simplest to start without hive and just test Machine+BaremetalHost resources?)

JoelSpeed: In a normal OpenShift cluster we restrict the controllers to only watch a single namespace (`openshift-machine-api`), so if you were to create a machine outside of this namespace, nothing would happen, and that is expected.

> I don't think there would be significant overhead in always watching all namespaces in a normal cluster. 

This entirely depends on what kinds of resources you are watching. For example, in the machine controller we need to load secrets for userdata and credentials. Normally you would use the controller-runtime cached client for this so that you get notified of changes to the secrets to trigger a reconcile. If you don't restrict this cache to a certain namespace, it will cache all secrets from all namespaces, which not only could impact memory usage, crosses a security boundary (eg multi-tenant clusters must not list secrets from multiple namespaces).

sadasu: > Are we Ok with unconditionally watching all namespaces, or should we have a provisioning CR option to enable it just for the hub-cluster use-case?

In some earlier discussions around this, I recall that we said we cannot assume Provisioning CR is going to be present in the multi-namespace scenario.

@JoelSpeed the issues you mentioned above are valid. I think our team is viewing this approach only for 4.8 and aligning with MAO's approach for 4.9 which might include multiple instances of BMO or even CBO (still under review). The current assumption is that we can live with these limitations/concerns listed above for 1 release.

dhellmann: > In a normal OpenShift cluster we restrict the controllers to only watch a single namespace (`openshift-machine-api`), so if you were to create a machine outside of this namespace, nothing would happen, and that is expected.
> 
> > I don't think there would be significant overhead in always watching all namespaces in a normal cluster.
> 
> This entirely depends on what kinds of resources you are watching. For example, in the machine controller we need to load secrets for userdata and credentials. Normally you would use the controller-runtime cached client for this so that you get notified of changes to the secrets to trigger a reconcile. If you don't restrict this cache to a certain namespace, it will cache all secrets from all namespaces, which not only could impact memory usage, crosses a security boundary (eg multi-tenant clusters must not list secrets from multiple namespaces).

We do watch Secrets as well as the hosts, but only those associated with the hosts. Based on what you're saying, it sounds like we may want to change the deployment so that our single baremetal-operator doesn't have to watch all Secrets? I'm not sure we'll actually save any RAM that way, since we'll have N copies of the operator with smaller caches, but it does address the security boundary issue.

We really can't afford to run multiple copies of Ironic, though. I think that means splitting the metal3 pod apart, which we didn't feel we would have time to do this release.

provisioning/baremetal_pod.go
```diff
@@ -280,14 +280,6 @@ func createContainerMetal3BaremetalOperator(images *Images, config *metal3iov1al
...
-				ValueFrom: &corev1.EnvVarSource{
```
andfasano, line 285: Question: what will be the impact, if any, on `LeaderElectionNamespace` for [BMO](https://github.com/metal3-io/baremetal-operator/blob/1432bfeab49722de48eaa35eef8279899c57c0fd/main.go#L114) in case of empty namespace?

provisioning/baremetal_pod.go
JoelSpeed: I spoke with @hardys yesterday and I think given your current plans for 4.8, it makes sense to be able to watch all namespaces for the first iteration and look into later splitting into per namespace. Security boundaries are less important if all the hosts belong to the same person for example.

But I do still question removing the namespace limiting functionality altogether, is it not useful to keep that for regular OCP clusters and just have the ability to configure whether it should be namespaced or watch everything?

> We do watch Secrets

Depending on what you do with the secrets, it may be better to use a non-caching client and Get instead of list. Looks like you're creating the secrets though so that's probably not possible here, else you'd lose the event notification if they are modified out of band

provisioning/baremetal_pod.go
sadasu: It appears that leader election needs a Namespace to be passed in https://github.com/kubernetes-sigs/controller-runtime/blob/0b554ebb54901a88427d4a89014af0d2dba1bbcf/pkg/manager/manager.go#L319.

With this change, BMO will not have the Namespace passed in via WATCH_NAMESPACE. But, with https://github.com/openshift/cluster-baremetal-operator/pull/107 we are passing in the POD_NAMESPACE which is the correct Namespace to be used for leader election. /cc @asalkeld

provisioning/baremetal_pod.go
sadasu: The CBO is not responsible for creating the Secrets for BMC credentials and does not try to access them. I think the above comment is meant for those Secrets and we need to make sure within BMO that we follow the above guidelines i.e do a GET for a specific Namespace vs a List for all Namespaces etc. Also, make sure we are using the right client to access the Secrets.

hardys: > I recall that we said we cannot assume Provisioning CR is going to be present in the multi-namespace scenario. Has that changed?

@sadasu I don't recall saying that - I don't see any reason we couldn't put some configuration option in the provisioning CR, so that the current single-namespace behavior is maintained, but for the hub-cluster use-case watching all namespaces could be enabled?

The main disadvantage of that approach is probably that it expands the test matrix, and we potentially have to care about the situation where we switch from single->multi namespace mode as mentioned in https://github.com/metal3-io/baremetal-operator/pull/797#discussion_r579026389

sadasu: https://github.com/openshift/cluster-baremetal-operator/pull/106/commits/cb4787025dfcb058431674e92a4ba59ea41b643b is taking care of adding this config to Provisioning CR.

api/v1alpha1/provisioning_types.go
```diff
@@ -99,6 +99,10 @@ type ProvisioningSpec struct {
 	// accessible from the machine networks. User should provide two IPs on
 	// the external network that would be used for provisioning services.
 	ProvisioningNetwork ProvisioningNetwork `json:"provisioningNetwork,omitempty"`
+
+	// EnableMultiNamespaces provides a way to explicitly allow use of this
+	// Provisioning configuration across multiple namespaces.
+	EnableMultiNamespaces bool `json:"enableMultiNamespaces,omitempty"`
```
hardys, line 105: I wonder if we can use some existing API for this, e.g for BMH resources we already set metadata/namespace:

https://github.com/openshift/installer/blob/master/pkg/asset/machines/baremetal/hosts.go#L75

However, we unfortunately didn't add that to the provisioning CR here:

https://github.com/openshift/installer/blob/master/data/data/manifests/openshift/baremetal-provisioning-config.yaml.template#L4

But I think such an interface might fit more cleanly, e.g similar to the OLM operator namespace behavior described here:

https://docs.openshift.com/container-platform/4.6/operators/understanding/olm/olm-understanding-operatorgroups.html#olm-operatorgroups-target-namespace_olm-understanding-operatorgroups

Is there any way we could add that missing metadata on upgrade, e.g by looking at the status/generations which does show the openshift-machine-api namespace?

Alternatively if we think modifying that existing behavior will be too difficult or ambiguous, I'd suggest renaming this interface to something terser e.g:

```
   spec:
        allNamespaces: true
```

That would align pretty well with the OLM docs e.g `targetNamespaces` ?

sadasu: Currently, Provisioning CR is cluster scoped. If we change it to Namespace scope, we would get this metadata change. This would be a breaking API change during an upgrade. The solution that is proposed here only tells us if multi-namespace mode is enabled. It will not allow up to specify an actual namespace. When we move to model where there would be a Provisioning CR per namespace, we could make the change to make Provisioning CRD with its scope set to Namespace.

sadasu: After going through the links mentioned above, I see that the semantics of the OLM operator namespace behavior different from what we are looking for here. If we re-use the name, would it cause some confusion?

hardys: @sadasu Ok sounds like we should go with the existing boolean, but perhaps rename the flag to `allNamespaces` (not blocking on that, but it seems to communicate the same as the proposed flag in a slightly terser and more accurate way, since "multi" could mean all, or some subset of all?)

stbenjam: - Would we never want to watch some but not all namespaces?

- I'm also not a fan of more knobs.  If it's always going to be turned on, why do we need it? And if it can be conditionally on/off how does one do that from the installer? Or do you change it on day 2?
-

sadasu: When in this mode we ask BMO to watch all Namespaces (by not setting the WATCH_NAMESPACE env var), although in practice BMH objects would likely be created in some Namespaces.

Point taken about additional knob. Best solution would be to make Provisioning CR Namespaced but that is coming later. This config would be false by default and also from the Installer and can be set to true on Day2.

Also, @kirankt 's suggestion to call the knob, "WatchAllNamespaces" is a better choice that what I have now, imo. I am not sure if it is terse enough but it conveys its purpose clearly.

hardys: >   Would we never want to watch some but not all namespaces?

Possibly, but I don't think there's any requirement for that right now.

>  I'm also not a fan of more knobs. If it's always going to be turned on, why do we need it? And if it can be conditionally on/off how does one do that from the installer? Or do you change it on day 2?

IMO this shouldn't always be turned on, particularly given that we want to align with the machine-api e.g "In a normal OpenShift cluster we restrict the controllers to only watch a single namespace (openshift-machine-api)" comment from @JoelSpeed 

The future direction for multi-namespace at the machine-api layer seems unclear at this point, so it seems safest to leave the existing single-namespace behavior alone, and add a knob to enable the specific "hub cluster" use-case which isn't a normal OpenShift cluster.

In terms of how to configure it, I was thinking either modify the installer manifest or ensure it can be turned on after initial deployment (at least until the story around multi-namespace machine management is worked out for other controllers, when perhaps there will be some kind of installer flag).

kirankt (approved):
/lgtm

api/v1alpha1/provisioning_types.go
```diff
@@ -99,6 +99,14 @@ type ProvisioningSpec struct {
...
 	// the external network that would be used for provisioning services.
 	ProvisioningNetwork ProvisioningNetwork `json:"provisioningNetwork,omitempty"`
+
+	// WatchAllNamespaces provides a way to explicitly allow use of this
+	// Provisioning configuration across all Namespaces. It is an
+	// optional configuration which defaults to false and in that state
+	// will be used to provision baremetal hosts in only the
+	// openshift-machine-api namespace. When set to true, this provisioning
+	// configuration would be used for baremetal hosts across all namespaces.
+	WatchAllNamespaces bool `json:"watchAllNamespaces,omitempty"`
```
dhellmann, line 109: We generally try to avoid using `bool` fields in the API, because it makes it difficult to expand or redefine them later. Here I would suggest an enum with 2 values `Single` and `All`, using `All` as the default.

PR #107, opened 2021-02-18
Add POD_NAMESPACE env variable to bmo container

asalkeld:
This is to support https://github.com/metal3-io/baremetal-operator/pull/797


kirankt (approved):
/lgtm

PR #109, opened 2021-02-22
Bug 1885349: Inject proxy environment variables everywhere

stbenjam:
This mounts the trusted CA bundle, and reads and injects the proxy
configuration into all of our containers. This is somewhat based off
what machine-api-operator does. This is because the "cluster-wide" proxy
configuration is not truly clsuter-wide, and we have to inject this
variables ourselves. We also inject the trusted CA bundle, as it's likely
many users have custom CA's in use for their TLS certificates.

### DIFF
--- a/controllers/provisioning_controller.go
+++ b/controllers/provisioning_controller.go
@@ -73,6 +73,7 @@ type ensureFunc func(*provisioning.ProvisioningInfo) (bool, error)
 // +kubebuilder:rbac:namespace=openshift-machine-api,groups="",resources=services,verbs=get;list;watch;create;update;patch;delete
 // +kubebuilder:rbac:namespace=openshift-machine-api,groups=monitoring.coreos.com,resources=servicemonitors,verbs=create;watch;get;list;patch
 
+// +kubebuilder:rbac:groups=config.openshift.io,resources=proxies,verbs=get;list;watch
 // +kubebuilder:rbac:groups=config.openshift.io,resources=infrastructures,verbs=get;list;watch
 // +kubebuilder:rbac:groups=security.openshift.io,resources=securitycontextconstraints,verbs=use
 // +kubebuilder:rbac:groups=config.openshift.io,resources=clusteroperators;clusteroperators/status,verbs=get;list;watch;create;update;patch;delete
@@ -178,7 +179,13 @@ func (r *ProvisioningReconciler) Reconcile(req ctrl.Request) (ctrl.Result, error
 		return ctrl.Result{}, err
 	}
 
-	info := r.provisioningInfo(baremetalConfig, &containerImages)
+	// Get cluster-wide proxy information
+	clusterWideProxy, err := r.OSClient.ConfigV1().Proxies().Get(context.Background(), "cluster", metav1.GetOptions{})
+	if err != nil {
+		return ctrl.Result{}, err
+	}
+
+	info := r.provisioningInfo(baremetalConfig, &containerImages, clusterWideProxy)
 
 	// Check if Provisioning Configuartion is being deleted
 	deleted, err := r.checkForCRDeletion(info)
@@ -301,7 +308,7 @@ func (r *ProvisioningReconciler) Reconcile(req ctrl.Request) (ctrl.Result, error
 	return ctrl.Result{}, nil
 }
 
-func (r *ProvisioningReconciler) provisioningInfo(provConfig *metal3iov1alpha1.Provisioning, images *provisioning.Images) *provisioning.ProvisioningInfo {
+func (r *ProvisioningReconciler) provisioningInfo(provConfig *metal3iov1alpha1.Provisioning, images *provisioning.Images, proxy *osconfigv1.Proxy) *provisioning.ProvisioningInfo {
 	return &provisioning.ProvisioningInfo{
 		Client:        r.KubeClient,
 		EventRecorder: events.NewLoggingEventRecorder(ComponentName),
@@ -309,6 +316,7 @@ func (r *ProvisioningReconciler) provisioningInfo(provConfig *metal3iov1alpha1.P
 		Scheme:        r.Scheme,
 		Namespace:     ComponentNamespace,
 		Images:        images,
+		Proxy:         proxy,
 	}
 }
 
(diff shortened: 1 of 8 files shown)


provisioning/baremetal_pod.go
```diff
@@ -29,23 +29,26 @@ import (
...
+	baremetalSharedVolume            = "metal3-shared"
+	metal3AuthRootDir                = "/auth"
+	ironicCredentialsVolume          = "metal3-ironic-basic-auth"
+	ironicrpcCredentialsVolume       = "metal3-ironic-rpc-basic-auth"
+	inspectorCredentialsVolume       = "metal3-inspector-basic-auth"
+	htpasswdEnvVar                   = "HTTP_BASIC_HTPASSWD" // #nosec
+	mariadbPwdEnvVar                 = "MARIADB_PASSWORD"    // #nosec
+	cboOwnedAnnotation               = "baremetal.openshift.io/owned"
+	cboLabelName                     = "baremetal.openshift.io/cluster-baremetal-operator"
+	externalTrustBundleConfigMapName = "cbo-trusted-ca"
```
kirankt, line 51: What is the reason for duplication? Why can't we use the mao-trusted-ca configmap here? AFAIK, CBO and MAO need to live in the same namespace.

stbenjam: Cross-repo dependency seems like a bad idea, I don't want to rely on MAO's config map. What if they remove it or change the name?

stbenjam:
This is still somewhat broken I think, workers don't come up even though all the services start. Still investigating.

provisioning/baremetal_pod.go
```diff
@@ -29,23 +29,26 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/client"
 	"sigs.k8s.io/controller-runtime/pkg/controller/controllerutil"
 
+	configv1 "github.com/openshift/api/config/v1"
+
```
asalkeld, line 33: extra new line?

provisioning/image_cache.go
```diff
@@ -19,6 +19,8 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/client"
 	"sigs.k8s.io/controller-runtime/pkg/controller/controllerutil"
 
+	osconfigv1 "github.com/openshift/api/config/v1"
+
```
asalkeld, line 23: same

asalkeld (commented):
seems ok to me..

provisioning/baremetal_pod.go
```diff
@@ -651,7 +716,7 @@ func CheckExistingMetal3Deployment(client appsclientv1.DeploymentsGetter, target
 func EnsureMetal3Deployment(info *ProvisioningInfo) (updated bool, err error) {
 	// Create metal3 deployment object based on current baremetal configuration
 	// It will be created with the cboOwnedAnnotation
-	metal3Deployment := newMetal3Deployment(info.Namespace, info.Images, &info.ProvConfig.Spec, info.PodLabelSelector)
+	metal3Deployment := newMetal3Deployment(info.Namespace, info.Images, &info.ProvConfig.Spec, info.PodLabelSelector, info.Proxy)
```
sadasu, line 719: Since we are are passing in several items already stored in info and now we are adding the proxy config to this list, maybe we should just pass ProvisioningInfo as a parameter. This PR can continue to merge without this change but would be nice to see this change in this or a subsequent PR.

controllers/provisioning_controller.go
```diff
@@ -178,7 +179,13 @@ func (r *ProvisioningReconciler) Reconcile(req ctrl.Request) (ctrl.Result, error
 		return ctrl.Result{}, err
 	}
 
-	info := r.provisioningInfo(baremetalConfig, &containerImages)
+	// Get cluster-wide proxy information
+	clusterWideProxy, err := r.OSClient.ConfigV1().Proxies().Get(context.Background(), "cluster", metav1.GetOptions{})
```
sadasu, line 183: Should the operator watch for the Proxy config and react if it changes day -2 ? (it seems unlikely that a customer might change their proxy config on day-2) If so, do we need to add it to the list of watched resources here https://github.com/openshift/cluster-baremetal-operator/blob/master/controllers/provisioning_controller.go#L401 ?

manifests/0000_31_cluster-baremetal-operator_01_trusted-ca-configmap.yaml
```diff
@@ -0,0 +1,12 @@
+apiVersion: v1
```
sadasu, line 1: @asalkeld should this configmap start off in the config directory? Some of the initial manifests have been added directly to the manifests/ like the Images configmap. But, since then we have always started with yaml in config/ that kubebuilder generates the manifest for and we later move it manifests/. I feel like we could go either ways with this configmap, so trying to get your opinion.

controllers/provisioning_controller.go
stbenjam: Yes that makes sense, I was trying to figure out how to do this actually, thanks!

stbenjam:
/test e2e-metal-ipi-ovn-ipv6

Breakage should be resolved.

controllers/provisioning_controller.go
stbenjam: @sadasu How do I add this as being watched? It seems like I have a few options. I could use `For()`, but is that enough? Or do I need a `Watches()` with some specific handler that knows what to do?

sadasu: We could probably use Watches() from the controller-runtim https://github.com/openshift/cluster-baremetal-operator/blob/aab97dfe194c9f527f54209bb530ae249cedead3/vendor/sigs.k8s.io/controller-runtime/pkg/builder/controller.go#L115 to start watching the Proxies{} resource.

stbenjam: Do you know what the eventhandler would be?

stbenjam: I tested with `Owns` instead and my metal3 pod did eventually get restarted after modifying the Proxy object.

provisioning/baremetal_pod.go
stbenjam: That makes sense, but would prefer to change it separately to keep this PR from getting too big.

PR #115, opened 2021-03-05
Bug 1961226: Configure an IPA sshkey in the metal3 pod

derekhiggins:
Pass the SSHKey from install-config to the metal3
pod to allow ssh access to running IPA instances.

### DIFF
--- a/controllers/provisioning_controller.go
+++ b/controllers/provisioning_controller.go
@@ -24,6 +24,7 @@ import (
 	"strings"
 	"time"
 
+	"github.com/ghodss/yaml"
 	"github.com/pkg/errors"
 	"github.com/stretchr/stew/slice"
 	appsv1 "k8s.io/api/apps/v1"
@@ -51,6 +52,10 @@ const (
 	ComponentNamespace = "openshift-machine-api"
 	// ComponentName is the full name of CBO
 	ComponentName = "cluster-baremetal-operator"
+	// install-config access details
+	clusterConfigName      = "cluster-config-v1"
+	clusterConfigKey       = "install-config"
+	clusterConfigNamespace = "kube-system"
 )
 
 // ProvisioningReconciler reconciles a Provisioning object
@@ -81,7 +86,7 @@ type ensureFunc func(*provisioning.ProvisioningInfo) (bool, error)
 // +kubebuilder:rbac:groups=config.openshift.io,resources=clusteroperators;clusteroperators/status,verbs=get;list;watch;create;update;patch;delete
 // +kubebuilder:rbac:groups=config.openshift.io,resources=infrastructures;infrastructures/status,verbs=get
 // +kubebuilder:rbac:groups="",resources=events,verbs=create;watch;list;patch
-// +kubebuilder:rbac:groups="",resources=secrets;services,verbs=get;list;watch;create;update;patch;delete
+// +kubebuilder:rbac:groups="",resources=configmaps;secrets;services,verbs=get;list;watch;create;update;patch;delete
 // +kubebuilder:rbac:groups=apps,resources=deployments;daemonsets,verbs=get;list;watch;create;update;patch;delete
 // +kubebuilder:rbac:groups=metal3.io,resources=provisionings;provisionings/finalizers,verbs=get;list;watch;create;update;patch;delete
 // +kubebuilder:rbac:groups=metal3.io,resources=provisionings/status,verbs=get;update;patch
@@ -116,6 +121,25 @@ func (r *ProvisioningReconciler) readProvisioningCR(ctx context.Context) (*metal
 		return nil, errors.Wrap(err, "unable to read Provisioning CR")
 	}
 	return instance, nil
+}
+
+type InstallConfigData struct {
+	SSHKey string
+}
+
+func (r *ProvisioningReconciler) readSSHKey() string {
+	installConfigData := InstallConfigData{}
+	clusterConfig, err := r.KubeClient.CoreV1().ConfigMaps(clusterConfigNamespace).Get(context.Background(), clusterConfigName, metav1.GetOptions{})
+	if err != nil {
+		klog.Warningf("Error: %v", err)
+		return ""
+	}
+	err = yaml.Unmarshal([]byte(clusterConfig.Data[clusterConfigKey]), &installConfigData)
+	if err != nil {
+		klog.Warningf("Error: %v", err)
+		return ""
+	}
+	return strings.TrimSpace(installConfigData.SSHKey)
 }
 
 // Reconcile updates the cluster settings when the Provisioning
@@ -197,7 +221,7 @@ func (r *ProvisioningReconciler) Reconcile(ctx context.Context, req ctrl.Request
 		}
 	}
 
-	info, err := r.provisioningInfo(baremetalConfig, &containerImages)
+	info, err := r.provisioningInfo(baremetalConfig, &containerImages, r.readSSHKey())
 	if err != nil {
 		return ctrl.Result{}, err
 	}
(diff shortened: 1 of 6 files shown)


manifests/0000_31_cluster-baremetal-operator_05_rbac.yaml
```diff
@@ -236,21 +236,19 @@ rules:
...
     include.release.openshift.io/self-managed-high-availability: "true"
     include.release.openshift.io/single-node-developer: "true"
-  name: cluster-baremetal-operator
-  namespace: openshift-machine-api
 roleRef:
-  apiGroup: rbac.authorization.k8s.io
-  kind: Role
-  name: cluster-baremetal-operator
+  kind: ClusterRole
+  name: cluster-admin
```
derekhiggins, line 247: I needed to set this rolebinding to allow access to the configmap in another namespace,
is there a more appropriate way to allow access?
`error: configmaps "cluster-config-v1" is forbidden: User "system:serviceaccount:openshift-machine-api:cluster-baremetal-operator" cannot get resource "configmaps" in API group "" in the namespace "kube-system"`

config/rbac/role_binding.yaml
```diff
@@ -19,7 +19,7 @@ metadata:
 roleRef:
   apiGroup: rbac.authorization.k8s.io
   kind: ClusterRole
-  name: cluster-baremetal-operator
+  name: cluster-admin
```
sadasu, line 22: Do we need 2 cluster roles cluster-baremetal-operator and cluster-admin here?

provisioning/baremetal_pod_test.go
```diff
@@ -212,7 +212,7 @@ func TestProxyAndCAInjection(t *testing.T) {
 		},
 		{
 			name:       "metal3 containers have proxy and CA information",
-			containers: newMetal3Containers(&images, managedProvisioning().build(), &proxy),
+			containers: newMetal3Containers(&images, managedProvisioning().build(), &proxy, ""),
 		},
 	}
 	for _, tc := range tCases {
```
sadasu, line 218: Can we add a new test for when the sshKey is not ""?

manifests/0000_31_cluster-baremetal-operator_05_rbac.yaml
sadasu: You'll need to add a new rbac rule in this section https://github.com/openshift/cluster-baremetal-operator/blob/master/controllers/provisioning_controller.go#L69.

config/rbac/role_binding.yaml
derekhiggins: I don't know to be honest, what I've don't appears to work but mightn't be correct

provisioning/baremetal_pod_test.go
derekhiggins: will upload it now

config/rbac/role_binding.yaml
asalkeld: i think it would be better if possible to keep the existing name

manifests/0000_31_cluster-baremetal-operator_05_rbac.yaml
derekhiggins: I've added configmaps to the cluster-baremetal-operator ClusterRole which I think is what I should have been doing in the first place.

config/rbac/role_binding.yaml
derekhiggins: I've added configmaps to the cluster-baremetal-operator ClusterRole which I think is what I should have been doing in the first place.

provisioning/baremetal_pod.go
```diff
@@ -719,10 +726,9 @@ func createContainerMetal3StaticIpManager(images *Images, config *metal3iov1alph
 	return container
 }
 
-func newMetal3PodTemplateSpec(info *ProvisioningInfo, labels *map[string]string) *corev1.PodTemplateSpec {
+func newMetal3PodTemplateSpec(info *ProvisioningInfo, labels *map[string]string, sshKey string) *corev1.PodTemplateSpec {
 	initContainers := newMetal3InitContainers(info)
-	containers := newMetal3Containers(info)
-
+	containers := newMetal3Containers(info, sshKey)
```
sadasu, line 731: Isn't sshKey part of ProvisioningInfo already?

provisioning/baremetal_pod.go
```diff
@@ -818,7 +824,7 @@ func envWithProxy(proxy *configv1.Proxy, envVars []corev1.EnvVar) []corev1.EnvVa
 	return envVars
 }
 
-func newMetal3Deployment(info *ProvisioningInfo) *appsv1.Deployment {
+func newMetal3Deployment(info *ProvisioningInfo, sshKey string) *appsv1.Deployment {
```
sadasu, line 827: Does sshKey need to be a separate parameter when it is already part of ProvisioningInfo?

sadasu:
I have a couple of minor comments. Would be nice to get those fixed before the merge.

derekhiggins:
> I have a couple of minor comments. Would be nice to get those fixed before the merge.

thanks, updated, ProvisioningInfo wasn't being passed into those functions when this patch was first created.

sadasu:
> > I have a couple of minor comments. Would be nice to get those fixed before the merge.
> 
> thanks, updated, ProvisioningInfo wasn't being passed into those functions when this patch was first created.

Yes, that is a more recent change. Thanks for updating your patch to accommodate that.

sadasu (commented):
/lgtm

provisioning/baremetal_pod_test.go
```diff
@@ -151,34 +151,51 @@ func TestNewMetal3Containers(t *testing.T) {
...
 		t.Run(tc.name, func(t *testing.T) {
 			t.Logf("Testing tc : %s", tc.name)
-			info := &ProvisioningInfo{Images: &images, ProvConfig: &metal3iov1alpha1.Provisioning{Spec: *tc.config}}
+			info := &ProvisioningInfo{Images: &images, ProvConfig: &metal3iov1alpha1.Provisioning{Spec: *tc.config}, SSHKey: tc.sshkey}
 			actualContainers := newMetal3Containers(info)
 			assert.Equal(t, tc.expectedContainers, len(actualContainers), fmt.Sprintf("%s : Expected number of Containers : %d Actual number of Containers : %d", tc.name, tc.expectedContainers, len(actualContainers)))
+			for _, container := range actualContainers {
+				if getSSHKey[container.Name] {
+					assert.Contains(t, container.Env, corev1.EnvVar{Name: "IRONIC_RAMDISK_SSH_KEY", Value: tc.sshkey}, fmt.Sprintf("Expected %s container to contain a sshkey %s", container.Name, tc.sshkey))
+				} else {
```
asalkeld, line 191: this test could be cleaned up to have an expected map of envs so we are testing all env variables..

asalkeld (commented):
there is two minor nits, but it's been on review for a long time..
/lgtm

controllers/provisioning_controller.go
```diff
@@ -37,6 +38,8 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/client"
 	"sigs.k8s.io/controller-runtime/pkg/controller/controllerutil"
 
+	"github.com/ghodss/yaml"
```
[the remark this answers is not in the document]
derekhiggins: np, done

provisioning/baremetal_pod_test.go
derekhiggins: I took a look at doing this, to test all env variables each entry in the tCases list would need a list of structs (one for each container), I can do that as a follow up if you want but would like to try and merge this as is to avoid complicating/delaying it more.

sadasu (approved):
/approve

asalkeld (commented):
/lgtm

asalkeld (commented):
/lgtm

PR #116, opened 2021-03-05
Only set IMAGES_JSON if not previously set

derekhiggins:
Allow us to use the env variable $IMAGES_JSON if set,
if not then default to the /etc/cluster-baremetal-operator
value. With this we can set the value in dev-scripts/metal3-dev/local-cbo.sh


sadasu (approved):
Thanks!

derekhiggins:
dev-scripts patch to make use of this is here 
https://github.com/openshift-metal3/dev-scripts/pull/1198

PR #105, opened 2021-02-11
Update BMH CRD from metal3-io/baremetal-operator

honza:

PR #78, opened 2020-11-26
Annotate manifests for single-node-developer cluster profile

guillaumerose:
This implements phase 1 of openshift/enhancements#482

PR #70, opened 2020-11-17
Add default value to BMH errorCount field

andfasano:
Add default value to BMH errorCount field to allow smoother upgrades.

Related to:
[1] https://github.com/metal3-io/baremetal-operator/pull/722
[2] https://github.com/openshift/machine-api-operator/pull/747

PR #118, opened 2021-03-10
Update BMH CRD

hardys:
This is to align with the rebase proposed via:
https://github.com/openshift/baremetal-operator/pull/133

### DIFF
--- a/manifests/0000_31_cluster-baremetal-operator_03_baremetalhost.crd.yaml
+++ b/manifests/0000_31_cluster-baremetal-operator_03_baremetalhost.crd.yaml
@@ -180,6 +180,106 @@ spec:
               online:
                 description: Should the server be online?
                 type: boolean
+              raid:
+                description: RAID configuration for bare metal server
+                properties:
+                  hardwareRAIDVolumes:
+                    description: The list of logical disks for hardware RAID, if rootDeviceHints isn't used, first volume is root volume.
+                    items:
+                      description: HardwareRAIDVolume defines the desired configuration of volume in hardware RAID
+                      properties:
+                        level:
+                          description: 'RAID level for the logical disk. The following levels are supported: 0;1;2;5;6;1+0;5+0;6+0.'
+                          enum:
+                          - "0"
+                          - "1"
+                          - "2"
+                          - "5"
+                          - "6"
+                          - 1+0
+                          - 5+0
+                          - 6+0
+                          type: string
+                        name:
+                          description: Name of the volume. Should be unique within the Node. If not specified, volume name will be auto-generated.
+                          maxLength: 64
+                          type: string
+                        numberOfPhysicalDisks:
+                          description: Integer, number of physical disks to use for the logical disk. Defaults to minimum number of disks required for the particular RAID level.
+                          minimum: 1
+                          type: integer
+                        rotational:
+                          description: Select disks with only rotational or solid-state storage
+                          type: boolean
+                        sizeGibibytes:
+                          description: Size (Integer) of the logical disk to be created in GiB. If unspecified or set be 0, the maximum capacity of disk will be used for logical disk.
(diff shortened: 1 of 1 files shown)


manifests/0000_31_cluster-baremetal-operator_03_baremetalhost.crd.yaml
```diff
@@ -1,13 +1,12 @@
+
 ---
 apiVersion: apiextensions.k8s.io/v1
 kind: CustomResourceDefinition
 metadata:
-  name: baremetalhosts.metal3.io
   annotations:
-    include.release.openshift.io/self-managed-high-availability: "true"
-    include.release.openshift.io/single-node-developer: "true"
```
hardys, line 8: @honza did you manually merge these in the previous update? 

`get-bmh-crd.sh`
 doesn't add them, so I'm not sure where these come from, or why we need them tbh.

hardys: Ack, thanks, I guess we could merge these in the hack script with `yq` but it'd be good to clarify where we look to keep the openshift-specific additions updated first so I'll do it manually for now :)

manifests/0000_31_cluster-baremetal-operator_03_baremetalhost.crd.yaml
```diff
@@ -1,3 +1,4 @@
+
```
kirankt, line 1: Nit. Can we remove this line?

kirankt (approved):

sadasu:
This CRD change needs to be made here https://github.com/openshift/cluster-baremetal-operator/blob/master/api/v1alpha1/provisioning_types.go. "make manifests" will take care of actually generating the manifest.

manifests/0000_31_cluster-baremetal-operator_03_baremetalhost.crd.yaml
hardys: It's just copied from https://github.com/metal3-io/baremetal-operator/blob/master/config/crd/bases/metal3.io_baremetalhosts.yaml

I can remove it manually but IMHO we should fix this CRD update process to avoid any manual edits  (e.g put any downstream stuff into `get-bmh-crd.sh` but otherwise just copy from either upstream or the openshift BMO fork.

hardys:
> This CRD change needs to be made here https://github.com/openshift/cluster-baremetal-operator/blob/master/api/v1alpha1/provisioning_types.go. "make manifests" will take care of actually generating the manifest.

Ok, why does `hack/get-bmh-crd.sh` exist then?  That just copies the yaml?

honza:
You should be able to use the `hack/get-bmh-crd.sh` script to update the CRD. As mentioned above, this process isn't fully automatic. You'll need to tweak the CRD after running the script: the script deletes some annotations already in the CBO that we need to keep.

sadasu:
> > This CRD change needs to be made here https://github.com/openshift/cluster-baremetal-operator/blob/master/api/v1alpha1/provisioning_types.go. "make manifests" will take care of actually generating the manifest.
> 
> This is updating the BMH CRD from the BMO repo, via `hack/get-bmh-crd.sh` - this just copies the yaml, and the `provisioning_types.go` does not contain any code definition of the BMH resource AFAICS:
> 
> ```
> $ grep type provisioning_types.go 
> type ProvisioningNetwork string
> type ProvisioningSpec struct {
> type ProvisioningStatus struct {
> type Provisioning struct {
> type ProvisioningList struct {
> ```
> 
> That also wasn't the approach used in #105 #100 #78 #70 #64 #51 so perhaps @honza can confirm, but I think this is the right process?

That is correct. We had planned on coming up with a better way to sync the BMH CRD between BMO and CBO. This is a good time to do that.

PR #119, opened 2021-03-18
Set default values for resources.requests for all containers

asalkeld:
It seems like this test was just added:
https://github.com/openshift/origin/blob/master/test/extended/operators/resources.go#L30

here is a failure
https://prow.ci.openshift.org/view/gs/origin-ci-test/pr-logs/pull/openshift_cluster-baremetal-operator/84/pull-ci-openshift-cluster-baremetal-operator-master-e2e-metal-ipi/1372397778717642752

### DIFF
--- a/provisioning/baremetal_pod.go
+++ b/provisioning/baremetal_pod.go
@@ -23,6 +23,7 @@ import (
 
 	appsv1 "k8s.io/api/apps/v1"
 	corev1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/api/resource"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	appsclientv1 "k8s.io/client-go/kubernetes/typed/apps/v1"
 	"k8s.io/utils/pointer"
@@ -582,6 +583,17 @@ func newMetal3PodTemplateSpec(images *Images, config *metal3iov1alpha1.Provision
 			Operator:          corev1.TolerationOpExists,
 			TolerationSeconds: pointer.Int64Ptr(120),
 		},
+	}
+
+	for _, cont := range containers {
+		// these are "default" values to pass openshift e2e tests.
+		// to set specific values, define them in the container definition.
+		if _, ok := cont.Resources.Requests[corev1.ResourceCPU]; !ok {
+			cont.Resources.Requests[corev1.ResourceCPU] = resource.MustParse("10m")
+		}
+		if _, ok := cont.Resources.Requests[corev1.ResourceMemory]; !ok {
+			cont.Resources.Requests[corev1.ResourceMemory] = resource.MustParse("50Mi")
+		}
 	}
 
 	return &corev1.PodTemplateSpec{


provisioning/baremetal_pod.go
```diff
@@ -584,6 +585,17 @@ func newMetal3PodTemplateSpec(images *Images, config *metal3iov1alpha1.Provision
 		},
 	}
 
+	for _, cont := range containers {
+		// these are "default" values to pass openshift e2e tests.
```
andfasano, line 589: This sounds a little bit too generic (which tests?). I guess it's better to mention the underlying requirement, ie enforcing the OpenShift convention on resources and limits for control plane components (https://github.com/openshift/enhancements/blob/master/CONVENTIONS.md#resources-and-limits)

provisioning/baremetal_pod.go
```diff
@@ -584,6 +585,17 @@ func newMetal3PodTemplateSpec(images *Images, config *metal3iov1alpha1.Provision
...
+		// to set specific values, define them in the container definition.
+		if _, ok := cont.Resources.Requests[corev1.ResourceCPU]; !ok {
+			cont.Resources.Requests[corev1.ResourceCPU] = resource.MustParse("10m")
+		}
+		if _, ok := cont.Resources.Requests[corev1.ResourceMemory]; !ok {
+			cont.Resources.Requests[corev1.ResourceMemory] = resource.MustParse("50Mi")
+		}
+	}
```
andfasano, line 597: How those values have been calculated ? Could make sense to try using the formula suggested in the conventions doc for determining them more accurately?

asalkeld: I set the same as this: https://github.com/openshift/cluster-baremetal-operator/blob/master/config/cluster-baremetal-operator/cluster-baremetal-operator.yaml#L37-L40

asalkeld: any idea what a better default would be?

stbenjam:
We are going to start blocking releases here shortly due to the test failures, I think using what's in https://github.com/openshift/cluster-baremetal-operator/blob/master/config/cluster-baremetal-operator/cluster-baremetal-operator.yaml?rgh-link-date=2021-03-18T09%3A44%3A47Z#L37-L40 is fair, and we can adjust both values if you have a better idea what they should be.

honza:
/hold

It's missing requests for the image cache.

andfasano:
Ref https://github.com/openshift/origin/pull/25982

provisioning/baremetal_pod.go
smarterclayton: Yes, please don't set "default" values.  You need to set "correct" and "accurate" values ;)

smarterclayton:
/hold

The bug link in the origin PR will give you some time to fix this.  Don't have to rush, follow the guidelines and try to get the right values set (if guidelines in conventions aren't clear, please ask and doug and i can help)

asalkeld:
@honza is working on it, closing.

PR #124, opened 2021-04-08
Add management workload annotations

lack:
In support of the workload partitioning feature
(https://github.com/openshift/enhancements/pull/703), we need to add
annotations to all management pods and namespaces so they can be
properly identified and assigned to segregated management cores on
clusters configured to do so.

### DIFF
--- a/provisioning/baremetal_pod.go
+++ b/provisioning/baremetal_pod.go
@@ -56,6 +56,10 @@ const (
 	externalTrustBundleConfigMapName = "cbo-trusted-ca"
 )
 
+var podTemplateAnnotations = map[string]string{
+	"workload.openshift.io/management": `{"effect": "PreferredDuringScheduling"}`,
+}
+
 var deploymentRolloutStartTime = time.Now()
 var deploymentRolloutTimeout = 5 * time.Minute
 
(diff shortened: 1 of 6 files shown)


dhellmann:
The deployment created in https://github.com/openshift/cluster-baremetal-operator/blob/1d0501f2204eb11ef6422e4dbfa7c1067160780b/provisioning/baremetal_pod.go#L783 needs the annotation too.

provisioning/baremetal_pod.go
```diff
@@ -56,6 +56,10 @@ const (
 	externalTrustBundleConfigMapName = "cbo-trusted-ca"
 )
 
+var daemonsetAnnotations = map[string]string{
```
dhellmann, line 59: Nit: It's just a regular Deployment, not a DaemonSet.

dhellmann (commented):
This looks good to me. I would like @sadasu to double check that all of the deployments are updated.

/approve

asalkeld:
seems ok to me, wait for @sadasu to lgtm

sadasu:
AFAIK, this PR now needs to be associated with a bug for it to merge.

lack:
> AFAIK, this PR now needs to be associated with a bug for it to merge.

Correct.  We should be getting that set up shortly.  Thanks :)

PR #125, opened 2021-04-12
Bug 1948787: secret.StringData is a WriteOnly convience field, stop using it to read

asalkeld:
In updateTlsSecret() we do the following check:
```
	existingCert := secret.StringData[corev1.TLSCertKey]
	if existingCert != "" {
```
This should always fail as StringData is WriteOnly. This will result in continuous writes to the secret (that we are watching).

Note: from the godoc
```
	// stringData allows specifying non-binary secret data in string form.
	// It is provided as a write-only convenience method.
	// All keys and values are merged into the data field on write, overwriting any existing values.
	// It is never output when reading from the API.
	// +k8s:conversion-gen=false
	// +optional
	StringData map[string]string `json:"stringData,omitempty" protobuf:"bytes,4,rep,name=stringData"`
```
/cc @stbenjam 
This maybe contributing to the secret watch issue.

### DIFF
--- a/provisioning/baremetal_secrets.go
+++ b/provisioning/baremetal_secrets.go
@@ -32,8 +32,6 @@ const (
 	inspectorSecretName = "metal3-ironic-inspector-password"
 	inspectorUsername   = "inspector-user"
 	tlsSecretName       = "metal3-ironic-tls" // #nosec
-	tlsCertificateKey   = "tls.crt"
-	tlsPrivateKeyKey    = "tls.key"
 )
 
 // CreateMariadbPasswordSecret creates a Secret for Mariadb password
@@ -186,8 +184,8 @@ func updateTlsSecret(client coreclientv1.SecretsGetter, targetNamespace string, 
 		changed = true
 	}
 
-	existingCert := secret.StringData[tlsCertificateKey]
-	if existingCert != "" {
+	existingCert := secret.Data[corev1.TLSCertKey]
+	if len(existingCert) > 0 {
 		expired, err := IsTlsCertificateExpired(existingCert)
 		if err != nil {
 			return errors.Wrap(err, "failed to determine expiration date of TLS certificate")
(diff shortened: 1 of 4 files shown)


sadasu:
Good catch @asalkeld

stbenjam:
Beautiful:

```
count: 1243881, first: 2021-04-12T07:05:12-04:00, last: 2021-04-12T08:55:43-04:00, duration: 1h50m31.738458s
288372x              system:serviceaccount:kube-system:namespace-controller
92229x               system:serviceaccount:openshift-apiserver:openshift-apiserver-sa
65119x               system:admin
61339x               system:kube-controller-manager
```

We are no longer in the top 5.

stbenjam:
Sorry folks, @andfasano pointed out there's actually a typo in my commands (verb=watch instead of --verb=watch), so effectively all these commands were without filtering.  There were still a high number of requests from CBO, but they appear to be `get` not `watch` and this PR still seems to help.

stbenjam:
The serial error is expected: https://bugzilla.redhat.com/show_bug.cgi?id=1948364

/override ci/prow/e2e-metal-ipi-serial-ipv4

provisioning/baremetal_secrets.go
```diff
@@ -236,9 +234,9 @@ func CreateOrUpdateTlsSecret(client coreclientv1.SecretsGetter, targetNamespace
 			Name:      tlsSecretName,
 			Namespace: targetNamespace,
 		},
-		StringData: map[string]string{
-			tlsCertificateKey: cert.certificate,
-			tlsPrivateKeyKey:  cert.privateKey,
+		Data: map[string][]byte{
```
deads2k, line 237: are you looking for `ApplySecret` from https://github.com/openshift/library-go/blob/master/pkg/operator/resource/resourceapply/core.go#L237 ?

Many resources have gotchas and that package was written by the original API authors to handle the gotchas.  You may want to consider grabbing a dep on it.

andfasano:
Note: by comparing the results looks like that also the updates have been greatly reduced

stbenjam:
/retitle Bug 1948787: secret.StringData is a WriteOnly convience field, stop using it to read

stbenjam:
As discussed during team scrum we'll land this one and then [have a look at library-go in a follow-up](https://github.com/openshift/cluster-baremetal-operator/pull/125#discussion_r611730098)

/lgtm

PR #126, opened 2021-04-12
Bug 1903700: Fix Pod Selectors in metal3 pods created by CBO

sadasu:
This PR:
1. Removes accommodations made to deal with upgrading a MAO created metal3 Pod
2. Deletes any old metal3 pod with old/incorrect Pod Selectors and create a new one with the correct ones

### DIFF
--- a/provisioning/baremetal_pod.go
+++ b/provisioning/baremetal_pod.go
@@ -721,31 +721,16 @@ func envWithProxy(proxy *configv1.Proxy, envVars []corev1.EnvVar) []corev1.EnvVa
 	return envVars
 }
 
-func newMetal3Deployment(targetNamespace string, images *Images, config *metal3iov1alpha1.ProvisioningSpec, selector *metav1.LabelSelector, proxy *configv1.Proxy) *appsv1.Deployment {
-	if selector == nil {
-		selector = &metav1.LabelSelector{
-			MatchLabels: map[string]string{
-				"k8s-app":    metal3AppName,
-				cboLabelName: stateService,
-			},
-		}
-	}
-	k8sAppLabel := metal3AppName
-	apiLabelValue := ""
-	for k, v := range selector.MatchLabels {
-		if k == "k8s-app" {
-			k8sAppLabel = v
-		}
-		if k == "api" {
-			apiLabelValue = v
-		}
+func newMetal3Deployment(targetNamespace string, images *Images, config *metal3iov1alpha1.ProvisioningSpec, proxy *configv1.Proxy) *appsv1.Deployment {
+	selector := &metav1.LabelSelector{
+		MatchLabels: map[string]string{
+			"k8s-app":    metal3AppName,
+			cboLabelName: stateService,
+		},
 	}
 	podSpecLabels := map[string]string{
-		"k8s-app":    k8sAppLabel,
+		"k8s-app":    metal3AppName,
 		cboLabelName: stateService,
-	}
-	if apiLabelValue != "" {
-		podSpecLabels["api"] = apiLabelValue
 	}
 	template := newMetal3PodTemplateSpec(images, config, &podSpecLabels, proxy)
 	return &appsv1.Deployment{
@@ -756,7 +741,7 @@ func newMetal3Deployment(targetNamespace string, images *Images, config *metal3i
 				cboOwnedAnnotation: "",
 			},
 			Labels: map[string]string{
-				"k8s-app":    k8sAppLabel,
+				"k8s-app":    metal3AppName,
 				cboLabelName: stateService,
 			},
 		},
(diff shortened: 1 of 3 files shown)


sadasu:
Tried to keep the number of Gets for the Deployment to the minimum by checking for the Pod Selector only when ApplyDeployment() failed. Since the Pod's Label Selector is immutable, the ApplyDeployment should fail when we try to update an existing metal3 deployment with outdated (MAO based) Label Selector.

provisioning/baremetal_pod.go
```diff
@@ -798,13 +782,29 @@ func EnsureMetal3Deployment(info *ProvisioningInfo) (updated bool, err error) {
...
-		return
+		// Check if ApplyDeployment failed because the existing Pod had an outdated
+		// Pod Selector.
+		selector, get_err := getMetal3DeploymentSelector(info.Client.AppsV1(), info.Namespace)
+		if get_err != nil || selector == metal3Deployment.Spec.Selector {
+			return
+		}
+		// This is an older deployment with the incorrect Pod Selector.
+		// Delete deployment and re-create
+		_ = DeleteMetal3Deployment(info)
```
andfasano, line 793: Any particular reason to ignore the error here?

sadasu: If the delete didn't succeed, the error from the create would fail with an error that the resource already exists. So, in the specific case, I felt it was OK to not specifically report the error with the delete because it was immediately followed by a create.

sadasu:
@zaneb did the original PR envision any other new Label Selectors to be added to the Metal3 Pod or is this sufficient?

zaneb:
> @zaneb did the original PR envision any other new Label Selectors to be added to the Metal3 Pod or is this sufficient?

The existing code was already creating new Deployments with the label/selectors that I expected us to use long-term. So the only change required is to stop working around existing Deployments with the old labels and just delete them instead.

provisioning/baremetal_pod.go
zaneb: We won't see the reason that the delete didn't succeed in the logs if we don't return it.

provisioning/baremetal_pod.go
```diff
@@ -798,13 +782,29 @@ func EnsureMetal3Deployment(info *ProvisioningInfo) (updated bool, err error) {
...
+		_, create_err := info.Client.AppsV1().Deployments(info.Namespace).Create(
+			context.Background(),
+			metal3Deployment,
+			metav1.CreateOptions{})
```
zaneb, line 797: I personally wouldn't try to do a Create here. Just requeue and on the next reconcile ApplyDeployment will create it again. It's risky to try to create the Deployment in two different ways in two different places.

provisioning/baremetal_pod.go
```diff
@@ -798,13 +782,29 @@ func EnsureMetal3Deployment(info *ProvisioningInfo) (updated bool, err error) {
 		info.EventRecorder, metal3Deployment, expectedGeneration)
 	if err != nil {
 		err = fmt.Errorf("unable to apply Metal3 deployment: %w", err)
-		return
+		// Check if ApplyDeployment failed because the existing Pod had an outdated
+		// Pod Selector.
+		selector, get_err := getMetal3DeploymentSelector(info.Client.AppsV1(), info.Namespace)
+		if get_err != nil || selector == metal3Deployment.Spec.Selector {
```
zaneb, line 788: Pointer comparison is unlikely to ever match here ;) Does LabelSelector have a DeepEqual() method?

provisioning/baremetal_pod.go
sadasu: Well, the create's error would say that the create failed because the metal3 deployment already exists. That way we would know the delete succeeded.

provisioning/baremetal_pod.go
sadasu: Yeah, duh! Yes, it does. Fixed!

provisioning/baremetal_pod.go
sadasu: Agreed. Fixed.

provisioning/baremetal_pod.go
```diff
@@ -798,13 +783,23 @@ func EnsureMetal3Deployment(info *ProvisioningInfo) (updated bool, err error) {
...
 		err = fmt.Errorf("unable to apply Metal3 deployment: %w", err)
-		return
+		// Check if ApplyDeployment failed because the existing Pod had an outdated
+		// Pod Selector.
+		selector, get_err := getMetal3DeploymentSelector(info.Client.AppsV1(), info.Namespace)
+		if get_err != nil || equality.Semantic.DeepEqual(selector, metal3Deployment.Spec.Selector) {
+			return
+		}
+		// This is an older deployment with the incorrect Pod Selector.
+		// Delete deployment now and re-create in the next reconcile
```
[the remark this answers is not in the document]
sadasu: Yes, the code is currently replying on the reconcile being triggered due to the watch on the Deployment which is currently being deleted. Do you see any downside to that approach?

provisioning/baremetal_pod.go
```diff
@@ -798,13 +783,23 @@ func EnsureMetal3Deployment(info *ProvisioningInfo) (updated bool, err error) {
...
+		if delete_err := DeleteMetal3Deployment(info); delete_err != nil {
+			err = fmt.Errorf("unable to delete Metal3 deployment with incorrect Pod Selector: %w", err)
```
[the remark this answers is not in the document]
sadasu: Done.

provisioning/baremetal_pod.go
```diff
@@ -798,13 +783,25 @@ func EnsureMetal3Deployment(info *ProvisioningInfo) (updated bool, err error) {
...
+		// The operator is watching deployments so the reconcile should be triggered when metal3 deployment
+		// is deleted.
+		if delete_err := DeleteMetal3Deployment(info); delete_err != nil {
+			err = fmt.Errorf("unable to delete Metal3 deployment with incorrect Pod Selector: %w", delete_err)
+			return
+		}
 	}
-
 	if updated {
 		resourcemerge.SetDeploymentGeneration(&info.ProvConfig.Status.Generations, deployment)
```
zaneb, line 802: This code is now reachable in the event of an error due to a non-matching selector, so in that case we will set the expected generation back to 0. I've convinced myself that this is OK, since we are deleting the Deployment already. If ApplyDeployment finds the old Deployment gone, it will create a new one without looking at the generation; and if it finds the old one in the cache the selector won't match and it will try (unsuccessfully) to update again, also ignoring the generation.

zaneb (commented):
/lgtm

PR #127, opened 2021-04-13
Bug 1948787: follow up : Use ApplySecret()

asalkeld:
- use resourceapply.ApplySecret().
- make the create secrets function signature the same as the other provisioner functions

### DIFF
--- a/controllers/provisioning_controller.go
+++ b/controllers/provisioning_controller.go
@@ -235,12 +235,8 @@ func (r *ProvisioningReconciler) Reconcile(ctx context.Context, req ctrl.Request
 		}
 	}
 
-	//Create Secrets needed for Metal3 deployment
-	if err := provisioning.CreateAllSecrets(r.KubeClient.CoreV1(), ComponentNamespace, baremetalConfig, r.Scheme); err != nil {
-		return ctrl.Result{}, err
-	}
-
 	for _, ensureResource := range []ensureFunc{
+		provisioning.EnsureAllSecrets,
 		provisioning.EnsureMetal3Deployment,
 		provisioning.EnsureMetal3StateService,
 		provisioning.EnsureImageCache,
--- a/provisioning/baremetal_crypto.go
+++ b/provisioning/baremetal_crypto.go
@@ -87,7 +87,7 @@ func generateTlsCertificate(provisioningIP string) (TlsCertificate, error) {
 	}, nil
 }
 
-func IsTlsCertificateExpired(certificate []byte) (bool, error) {
+func isTlsCertificateExpired(certificate []byte) (bool, error) {
 	certs, err := cert.ParseCertsPEM(certificate)
 	if err != nil {
 		return false, err
--- a/provisioning/baremetal_crypto_test.go
+++ b/provisioning/baremetal_crypto_test.go
@@ -31,7 +31,7 @@ func TestGenerateTlsCertificate(t *testing.T) {
 		assert.NotEqual(t, cert.privateKey, "", "empty private key")
 	}
 
-	expired, err := IsTlsCertificateExpired(cert.certificate)
+	expired, err := isTlsCertificateExpired(cert.certificate)
 	if err != nil {
 		t.Errorf("Unexpected error while checking a certificate: %s", err)
 	} else {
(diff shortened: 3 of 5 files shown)


asalkeld:
/cc @deads2k @stbenjam 
Not the prettiest solution coping the function, but otherwise it will result in the same overwriting of secrets..

provisioning/applysecret.go
```diff
@@ -0,0 +1,104 @@
...
+
+type updateDataFn func(existing, required *corev1.Secret) error
+
+func doNotUpdateData(existing, required *corev1.Secret) error {
+	return nil
+}
+
+// applySecret merges objectmeta, applies data if the secret does not exist.
+// Note: this is mostly a copy of resourceapply.ApplySecret() except we allow to choose what
+//       data to update. This is because we don't want to keep rotating passwords.
```
asalkeld, line 27: Note: ^

provisioning/applysecret.go
```diff
@@ -0,0 +1,104 @@
...
+
+	existingCopy := existing.DeepCopy()
+
+	resourcemerge.EnsureObjectMeta(resourcemerge.BoolPtr(false), &existingCopy.ObjectMeta, required.ObjectMeta)
+
+	// Allow the caller to decide what data to update.
+	err = updateData(existingCopy, required)
+	if err != nil {
+		return err
+	}
```
asalkeld, line 63: this is the only bit that is different

provisioning/applysecret.go
```diff
@@ -0,0 +1,104 @@
...
+
+	existingCopy.Type = required.Type
+
+	// Server defaults some values and we need to do it as well or it will never equal.
+	if existingCopy.Type == "" {
+		existingCopy.Type = corev1.SecretTypeOpaque
+	}
```
sadasu, line 70: Could this be :
if required.Type == "" {
    existingCopy.Type = corev1.SecretTypeOpaque
} else {
        existingCopy.Type = required.Type
}

provisioning/applysecret.go
```diff
@@ -0,0 +1,104 @@
...
+		}
+	}
+
+	// if the field was immutable on a secret, we're going to be stuck until we delete it.  Try to delete and then create
+	_ = client.Secrets(required.Namespace).Delete(context.TODO(), existingCopy.Name, metav1.DeleteOptions{})
+
+	// clear the RV and track the original actual and error for the return like our create value.
+	existingCopy.ResourceVersion = ""
+	_, err = client.Secrets(required.Namespace).Create(context.TODO(), existingCopy, metav1.CreateOptions{})
+
```
sadasu, line 102: Should we be creating "required" here instead of existingCopy?

provisioning/applysecret.go
```diff
@@ -0,0 +1,104 @@
...
+	if klog.V(4).Enabled() {
+		klog.Infof("Secret %s/%s changes: %v", required.Namespace, required.Name, resourceapply.JSONPatchSecretNoError(existing, existingCopy))
+	}
+
+	/*
+	 * Kubernetes validation silently hides failures to update secret type.
+	 * https://github.com/kubernetes/kubernetes/blob/98e65951dccfd40d3b4f31949c2ab8df5912d93e/pkg/apis/core/validation/validation.go#L5048
+	 * We need to explicitly opt for delete+create in that case.
+	 */
```
sadasu, line 84: Since we are essentially writing our own applySecret() should we take care of a scenario where the secret.Type is changing? AFAIK, we do not expect any of our Secrets to change their type. So, in that case, should we just be returning an error instead of deleting and creating the Secret?

sadasu: Ok. I see thet existingCopy.Type has been updated to required.Type or corev1.SecretTypeOpaque.

provisioning/applysecret.go
```diff
@@ -0,0 +1,104 @@
...
+	if existingCopy.Type == existing.Type {
```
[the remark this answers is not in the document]
sadasu: Ok. I see that earlier existingCopy.Type has been set to required.Type or corev1.SecretTypeOpaque.

provisioning/applysecret.go
asalkeld: @sadasu this is mostly a copy of https://github.com/openshift/library-go/blob/master/pkg/operator/resource/resourceapply/core.go#L237
but what it does is get the existing object and copy over fields from the "required".

provisioning/applysecret.go
asalkeld: not sure, i thought it would be easier to compare if they are mostly the same?

PR #129, opened 2021-04-18
Use library-go instead of controller-runtime

asalkeld:
I have used https://github.com/openshift/generic-admission-server for the webhooks.


deads2k:
> * There doesn't seem to be any Webhook helper code in library-go, I was  going to look at something like this https://github.com/openshift/sriov-network-operator/tree/master/pkg/webhook (@deads2k please advise ..)

Is that an attempt to make something like https://github.com/openshift/generic-admission-server ?

controllers/provisioning_controller.go
```diff
@@ -114,82 +113,72 @@ func (r *ProvisioningReconciler) readProvisioningCR(ctx context.Context) (*metal
...
 		klog.Info("Provisioning CR not found")
-		return result, nil
+		return nil
+	}
+
+	if !r.webHookEnabled {
+		if provisioning.WebhookDependenciesReady(r.osClient) &&
+			slice.Contains(baremetalConfig.ObjectMeta.Finalizers, metal3iov1alpha1.ProvisioningFinalizer) {
+			klog.Info("restarting to enable the webhook")
+			os.Exit(1)
```
deads2k, line 153: this reads oddly.  Your controller is restarting?

asalkeld: We can't enable webhooks from the start, as service-ca is not ready to inject the ca in time. So we start without the webhook (which is fine as the installer creates the provisioning CR - so it's correct). Then later when service-ca is up, we restart with the webhook enabled. It is definitely one of the ugly things I'd like to fix with library-go.

stbenjam:
@asalkeld Can you rebase?

The fact e2e-metal-ipi and ovn-ipv6 both passed is pretty good sign.  It'd be nice to see a few successful runs

main.go
```diff
@@ -17,121 +17,163 @@ package main
...
+		controllerContext.EventRecorder,
+		releaseVersion,
+		imagesJSONFilename,
+	)
+
+	metal3Informers.Start(ctx.Done())
+	kubeInformersForNamespace.Start(ctx.Done())
+	configInformers.Start(ctx.Done())
+
+	go provisioningController.Run(ctx, 1)
```
deads2k, line 114: To aggregate status from FooDegraded, FooAvailable on a resource like kubeapiserver.operator.openshift.io to clusteroperator.config.openshift.io/kube-apiserver we use https://github.com/openshift/library-go/blob/7a65fdb398e28782ee1650959a5e0419121e97ae/pkg/operator/status/status_controller.go#L59-L67 .

This allows us to add various degraded checks easily.

webhooks/provisioning_webhook.go
```diff
@@ -0,0 +1,86 @@
...
+}
+
+// +kubebuilder:rbac:groups=flowcontrol.apiserver.k8s.io,resources=prioritylevelconfigurations;flowschemas,verbs=get;list;watch
+// +kubebuilder:rbac:groups=admission.metal3.io,resources=provisionings,verbs=create
+
+var _ apiserver.AdmissionHook = &ProvisioningValidatingWebHook{}
+
+func (a *ProvisioningValidatingWebHook) ValidatingResource() (plural schema.GroupVersionResource, singular string) {
+	return schema.GroupVersionResource{
+		Group:    "admission.metal3.io",
```
asalkeld, line 44: the amount of time i took to figure out i had this ^ wrong is mind boggling

dhellmann: Perhaps the team shouldn't adopt library-go until there is better documentation for using it?

asalkeld: It was in the example (so my bad), but I am happy to add docs..
Also this is not really library-go's fault as it is using https://github.com/openshift/generic-admission-server for the webhook

asalkeld: https://github.com/openshift/library-go/pull/1066

main.go
asalkeld: @deads2k that controller is not very obvious.. still trying to figure out what exactly it does

asalkeld:
There is a real e2e failure that amounts to
```
oc explain provisionings --api-version=metal3.io/v1alpha1
error: couldn't find resource for "metal3.io/v1alpha1, Kind=AdmissionReview"
```
also...
```
oc explain provisionings
error: couldn't find resource for "admission.metal3.io/v1alpha1, Kind=AdmissionReview"
```

deads2k:
> There is a real e2e failure that amounts to
> 
> ```
> oc explain provisionings --api-version=metal3.io/v1alpha1
> error: couldn't find resource for "metal3.io/v1alpha1, Kind=AdmissionReview"
> ```
> 
> also...
> 
> ```
> oc explain provisionings
> error: couldn't find resource for "admission.metal3.io/v1alpha1, Kind=AdmissionReview"
> ```

You'll want `oc get --raw /apis/admission.metal3.io/v1alpha1` and have @soltysh check the content to see if it's an `oc explain` bug (very possible) or a discovery bug.

asalkeld:
> > There is a real e2e failure that amounts to
> > ```
> > oc explain provisionings --api-version=metal3.io/v1alpha1
> > error: couldn't find resource for "metal3.io/v1alpha1, Kind=AdmissionReview"
> > ```
> > 
> > 
> > also...
> > ```
> > oc explain provisionings
> > error: couldn't find resource for "admission.metal3.io/v1alpha1, Kind=AdmissionReview"
> > ```
> 
> You'll want `oc get --raw /apis/admission.metal3.io/v1alpha1` and have @soltysh check the content to see if it's an `oc explain` bug (very possible) or a discovery bug.

oc get --raw /apis/admission.metal3.io/v1alpha1
{"kind":"APIResourceList","apiVersion":"v1","groupVersion":"admission.metal3.io/v1alpha1","resources":[{"name":"provisionings","singularName":"","namespaced":false,"group":"admission.k8s.io","version":"v1beta1","kind":"AdmissionReview","verbs":["create"]}]}

main.go
asalkeld: > > > There is a real e2e failure that amounts to
> > > ```
> > > oc explain provisionings --api-version=metal3.io/v1alpha1
> > > error: couldn't find resource for "metal3.io/v1alpha1, Kind=AdmissionReview"
> > > ```
> > > 
> > > 
> > > also...
> > > ```
> > > oc explain provisionings
> > > error: couldn't find resource for "admission.metal3.io/v1alpha1, Kind=AdmissionReview"
> > > ```
> > 
> > 
> > You'll want `oc get --raw /apis/admission.metal3.io/v1alpha1` and have @soltysh check the content to see if it's an `oc explain` bug (very possible) or a discovery bug.
> 
> oc get --raw /apis/admission.metal3.io/v1alpha1
> {"kind":"APIResourceList","apiVersion":"v1","groupVersion":"admission.metal3.io/v1alpha1","resources":[{"name":"provisionings","singularName":"","namespaced":false,"group":"admission.k8s.io","version":"v1beta1","kind":"AdmissionReview","verbs":["create"]}]}

I have worked around this by changing the APIService/Webhook path (I had a look at what hive was doing) https://github.com/openshift/cluster-baremetal-operator/pull/129/commits/2d1f5ef5d1b0f178b50ef12fe7677c915dd07236

asalkeld:
I am closing this, but it is still around if we want to bring this back. 
Why are we closing this?
- bmo and with cluster-api coming the capm3 controllers will all use controller-runtime
- this will be our only operator in the team that uses library-go and it doesn't warrant the mental load of the whole team learning a different way of doing things.

Personally, I quite like library-go, but it is an uphill battle as anything upstream will not use it.
@dhellmann @zaneb @sadasu

PR #133, opened 2021-04-21
Use Watch instead of Owns for the Proxy, as we do not have ownership

asalkeld:
Note: the proxy object is not owned by cbo so
```
Owns(&osconfigv1.Proxy{})
```
will not cause a Reconcile when the proxy object is changed.
This is now changed to a Watch() that maps the proxy object to our CR.
/cc @stbenjam


stbenjam:
> will not cause a Reconcile when the proxy object is changed.

It does in my testing, but this does seem more correct.

/approve
/assign @sadasu

asalkeld:
@stbenjam not sure how it worked, as there are no owner references on the proxy object.
```
oc get proxy -o yaml
apiVersion: v1
items:
- apiVersion: config.openshift.io/v1
  kind: Proxy
  metadata:
    creationTimestamp: "2021-04-21T10:25:29Z"
    generation: 1
    name: cluster
    resourceVersion: "597"
    uid: b37a6de2-1742-4e45-9448-dfc22c8c059f
  spec:
    trustedCA:
      name: ""
  status: {}
kind: List
metadata:
  resourceVersion: ""
  selfLink: ""
```
and this is the code that enqueues the Owns() objects https://github.com/kubernetes-sigs/controller-runtime/blob/master/pkg/handler/enqueue_owner.go#L170

stbenjam:
Edit the proxy object and you'll see the Metal3 pod restart, I just tried it.

```
$ oc edit proxy cluster                          

--- Auto-Commands ---
proxy.config.openshift.io/cluster edited
$ watch oc get pods -n openshift-machine-api
```

A few minutes later it restarts...

```
NAME                                           READY   STATUS        RESTARTS   AGE
cluster-autoscaler-operator-86549547b6-zlx8v   2/2     Running       0          8h
cluster-baremetal-operator-86f85478fc-66tfs    2/2     Running       0          8h
machine-api-controllers-5d897648bc-4xhjb       0/7     Terminating   3          8h
machine-api-controllers-5f77748b4d-64twd       7/7     Running       0          25s
machine-api-operator-bf66d85b9-zv6lq           2/2     Running       0          8h
metal3-5bd4585757-4plnh                        10/10   Terminating   0          8h
metal3-5bd4585757-848s2                        0/10    Init:0/3      0          0s
metal3-image-cache-b5qzn                       1/1     Running       0          8h
metal3-image-cache-bwz2g                       1/1     Running       0          8h
metal3-image-cache-xfx2d                       1/1     Running       0          8h
```

asalkeld:
mmm 🤔 , if it works, i should probably just leave it.. seems odd though.

sadasu:
@asalkeld @stbenjam we had already considered this while reviewing the original fix.
Also, controller-runtime recommends we use Owns instead. See: https://github.com/kubernetes-sigs/controller-runtime/blob/master/pkg/builder/controller.go#L123-L125

PR #134, opened 2021-04-23
generate our own client and use it

asalkeld:
/cc @sadasu you asked me to pull this out.
I personally like this, but don't think it will effect the throttling as the client still gets created and used
for the watches.

Note: the move of api/v1alpha1 to apis/metal.io/v1alpha1 is to make client-gen happy.


sadasu:
@asalkeld Thanks for separating these 2 commits out. I am fine with this change.

apis/metal3.io/v1alpha1/doc.go
```diff
@@ -13,23 +13,7 @@ See the License for the specific language governing permissions and
 limitations under the License.
 */
 
-// Package v1alpha1 contains API Schema definitions for the metal3io v1alpha1 API group
-// +kubebuilder:object:generate=true
+// Package v1alpha1 contains API Schema definitions for the metal3.io/v1alpha1 API group
```
zaneb, line 16: This is not part of the metal3 project, so I don't think we should be appropriating that namespace.

zaneb:
Can you elaborate on what the benefit of this is?
I'm not familiar with the mechanics, but I would have hoped that kubebuilder would have generated us something that is adequate to the task and which will become common with more other controllers over time.

asalkeld:
> Can you elaborate on what the benefit of this is?
> I'm not familiar with the mechanics, but I would have hoped that kubebuilder would have generated us something that is adequate to the task and which will become common with more other controllers over time.

@zaneb I have another PR up to migrate to library-go, so basically I am getting away from anything controller-runtime. This is separated out
just to make it easier to digest/review.

apis/metal3.io/v1alpha1/doc.go
asalkeld: ? we already have. This is just a comment spello

PR #135, opened 2021-04-26
Bug 1953736: Fix annotations for the ServiceMonitor

sadasu:
Adding the include.release.openshift.io/self-managed-high-availability: "true" annotation to the ServiceMonitor manifest so that it is installed by CVO

### DIFF
--- a/manifests/0000_90_cluster-baremetal-operator_03_servicemonitor.yaml
+++ b/manifests/0000_90_cluster-baremetal-operator_03_servicemonitor.yaml
@@ -3,6 +3,7 @@ kind: ServiceMonitor
 metadata:
   annotations:
     exclude.release.openshift.io/internal-openshift-hosted: "true"
+    include.release.openshift.io/self-managed-high-availability: "true"
   labels:
     k8s-app: cluster-baremetal-operator
   name: cluster-baremetal-operator-servicemonitor


sadasu:

known issue unrelated to this PR

kirankt (approved):
/lgtm

config/prometheus/service.yaml
```diff
@@ -6,15 +6,14 @@ metadata:
...
-    include.release.openshift.io/self-managed-high-availability: "true"
     service.alpha.openshift.io/serving-cert-secret-name: cluster-baremetal-operator-tls
   labels:
     k8s-app: cluster-baremetal-operator
 spec:
   type: ClusterIP
   ports:
   - name: https
-    port: 8443
+    port: 9444
```
zaneb, line 16: FWIW, neither CBO nor MAO are using host networking, so there is no port conflict.

There _would_ be a problem if they were both using the same labels so that either of the Services' selectors ended up selecting both Pods, but `k8s-app: cluster-baremetal-operator` should be pretty unique.

zaneb (commented):
The TLS error is likely because the port you're now connecting to (9444) is not the port the server is running on (presumably 8443).

I'm guessing the original error with the agnostic job was because the manifest was getting installed on a platform where the CBO wasn't actually started by CVO maybe?

ardaguclu:
@sadasu @zaneb after spending sometime on this e2e-agnostic failure, there are two problems;

- Certificate, serveName mismatch(correct: `cluster-baremetal-operator-service.openshift-machine-api.svc`)
- Lack of rbac roles to access tokenreviews and subjectaccessreviews resources

You can have a look at my playground PR https://github.com/openshift/cluster-baremetal-operator/pull/188

wking:
Bringing back some context, the profile annotations had been dropped in #111, when the single consuming alert was dropped (because it was noisy and redundant).  I don't see any PrometheusRules in master or this PR, although the master README does [mention metrics][1] (without listing the metrics being provided).  Is getting those metrics working the goal of this PR?

[1]: https://github.com/openshift/cluster-baremetal-operator/blame/0d79d56dcef615d9dcc06bd2e1df80da97a67826/README.md#L83-L85

sadasu:
/closed
We don't have any CBO level alerts that we want to report at this time.

PR #136, opened 2021-04-26
WIP: Remove monitoring of CBO metrics via prometheus

sadasu:


asalkeld:
why do we want to do this?

PR #137, opened 2021-04-29
Update controller-runtime version to v0.8.3

sadasu:
Some recent fixes in controller-runtime might be useful for us. Especially the one regarding adding a timeout while waiting for the cache to sync.


asalkeld:
will make a conflict with mine, but otherwise seems like a good idea if we don't merge my one.

asalkeld (commented):
/lgtm

PR #138, opened 2021-04-30
WIP: cbo should be enabled on all platforms

asalkeld:
We want the ability to run cbo on any platform so that it can be used for the hub and spoke model.
eg. hub AWS cluster creates a spoke baremetal cluster.
/cc @sadasu


asalkeld:
/hold
Not sure there is need for this..

PR #139, opened 2021-04-30
Bug 1953795: Set Ironic webserver_verify_ca

bfournie:
Set the env variable WEBSERVER_CACERT_FILE which is used by Ironic
for the ironic.conf setting ``webserver_verify_ca``.


stbenjam:
It's only needed in conductor, or is it possible the other Ironic containers might need to know this?

bfournie:
> It's only needed in conductor, or is it possible the other Ironic containers might need to know this?

Yes, its only used when downloading the image (e.g. https://github.com/openstack/ironic/blob/master/ironic/common/image_service.py#L127).

PR #140, opened 2021-04-30
Bug 1952448: Update condition for starting the metal3StaticIpManager container

sadasu:
The staticIPManager container is not required when the ProvisioningNetwork
is Disabled or the ProvisioningIP is not provided.


kirankt (approved):
/lgtm

PR #141, opened 2021-05-04
Bug 1954768: Use AdmissionRegistration v1 not v1beta1 as it is going away

asalkeld:
validatingwebhookconfigurations.v1beta1.admissionregistration.k8s.io will stop functioning in 4.9 so this will break upgrades.

- also use resourceapply.ApplyValidatingWebhookConfiguration()


kirankt (approved):
/lgtm

PR #142, opened 2021-05-05
Revert "Bug 1953795: Set Ironic webserver_verify_ca"

stbenjam:
Reverts openshift/cluster-baremetal-operator#139

This is the wrong cert to use for the assisted images, it uses the service ca, this needs more investigation in another release to make the service, ingress, and trusted CA bundles available to our containers, as well as investigating using TLS for the httpd hosted by Ironic.


stbenjam:
This needs a /lgtm to merge from someone else

PR #143, opened 2021-05-10
Bug 1946079: Pass the API VIP to the downloader container

honza:
Depends on https://github.com/openshift/ironic-rhcos-downloader/pull/40

### DIFF
--- a/provisioning/baremetal_pod_test.go
+++ b/provisioning/baremetal_pod_test.go
@@ -131,7 +131,7 @@ func TestNewMetal3InitContainers(t *testing.T) {
 	for _, tc := range tCases {
 		t.Run(tc.name, func(t *testing.T) {
 			t.Logf("Testing tc : %s", tc.name)
-			actualContainers := newMetal3InitContainers(&images, tc.config, nil)
+			actualContainers := newMetal3InitContainers(&images, tc.config, nil, "")
 			assert.Equal(t, len(tc.expectedContainers), len(actualContainers), fmt.Sprintf("%s : Expected number of Init Containers : %d Actual number of Init Containers : %d", tc.name, len(tc.expectedContainers), len(actualContainers)))
 		})
 	}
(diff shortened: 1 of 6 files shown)


sadasu:
@honza could you please add a little more information about the API VIP? As far as i know, this provides a way for remote workers not on the provisioning network to be able to download their image from the baremetal control plane (metal3 pod). Does not depend on the ProvisioningNetwork state.

sadasu:
@Ian @asalkeld do we need to add any validation for this new config?

hardys:
The API VIP is already stored in the platformStatus in the cluster infrastructure CR, can we use that instead of adding the provisioning CR field?

```
$ oc get infrastructures.config.openshift.io cluster -o yaml | yq .status.platformStatus
{
  "baremetal": {
    "apiServerInternalIP": "fd2e:6f44:5dd8:c956::5",
    "ingressIP": "fd2e:6f44:5dd8:c956::4"
  },
  "type": "BareMetal"
}
```

stbenjam:
We have an epic to enable CBO on other platforms than baremetal, so I'm not sure relying on the baremetal platform in the Infrastructure CR is the right thing to do.

Also does the infrastructure CR exist outside of IPI  installs?

Could we do a DNS look-up instead? See if an A record exists, if not then see if there's an AAAA record?

provisioning/baremetal_pod_test.go
```diff
@@ -208,7 +208,7 @@ func TestProxyAndCAInjection(t *testing.T) {
 	}{
 		{
 			name:       "init containers have proxy and CA information",
-			containers: newMetal3InitContainers(&images, managedProvisioning().build(), &proxy),
+			containers: newMetal3InitContainers(&images, managedProvisioning().build(), &proxy, ""),
 		},
 		{
 			name:       "metal3 containers have proxy and CA information",
```
sadasu, line 214: Can we add a test where we pass in a valid API VIP and check if the container.Env for the "metal3-machine-os-downloader" contains this API VIP?

asalkeld:
> We have an epic to enable CBO on other platforms than baremetal, so I'm not sure relying on the baremetal platform in the Infrastructure CR is the right thing to do. This will be empty on every other platform.
> 
> Also does the infrastructure CR exist outside of IPI installs?
> 
> Could we do a DNS look-up instead? See if an A record exists, if not then see if there's an AAAA record?

This is a really good point..
Note: that epic won't land this sprint, but we will have the same issue with capi cc @hardys

honza:
Closing in favour of #148

PR #144, opened 2021-05-11
Allow boot iso source configuration

elfosardo:
Allow setting the source boot iso parameter to enable or disable local
caching of the boot iso image.

### DIFF
--- a/api/v1alpha1/provisioning_types.go
+++ b/api/v1alpha1/provisioning_types.go
@@ -38,6 +38,16 @@ const (
 	ProvisioningSingletonName = "provisioning-configuration"
 )
 
+// BootIsoSource is the origin of the boot iso image
+// +kubebuilder:validation:Enum=local;http
+type BootIsoSource string
+
+// BootIsoSource values
+const (
+	BootIsoSourceLocal BootIsoSource = "local"
+	BootIsoSourceHttp  BootIsoSource = "http"
+)
+
 // ProvisioningSpec defines the desired state of Provisioning
 type ProvisioningSpec struct {
 	// ProvisioningInterface is the name of the network interface
@@ -110,6 +120,14 @@ type ProvisioningSpec struct {
 	// openshift-machine-api namespace. When set to true, this provisioning
 	// configuration would be used for baremetal hosts across all namespaces.
 	WatchAllNamespaces bool `json:"watchAllNamespaces,omitempty"`
+
+	// BootIsoSource provides a way to set the location where the iso image
+	// to boot the nodes will be served from.
+	// By default the boot iso image is cached locally and served from
+	// the Provisioning service (Ironic) nodes using an auxiliary httpd server.
+	// If the boot iso image is already served by an httpd server, setting
+	// this option to http allows to directly provide the image from there.
+	BootIsoSource BootIsoSource `json:"bootIsoSource,omitempty"`
 }
 
 // ProvisioningStatus defines the observed state of Provisioning
--- a/config/crd/bases/metal3.io_provisionings.yaml
+++ b/config/crd/bases/metal3.io_provisionings.yaml
@@ -31,6 +31,12 @@ spec:
           spec:
             description: ProvisioningSpec defines the desired state of Provisioning
             properties:
+              bootIsoSource:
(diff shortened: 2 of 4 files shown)


api/v1alpha1/provisioning_types.go
```diff
@@ -110,6 +110,9 @@ type ProvisioningSpec struct {
 	// openshift-machine-api namespace. When set to true, this provisioning
 	// configuration would be used for baremetal hosts across all namespaces.
 	WatchAllNamespaces bool `json:"watchAllNamespaces,omitempty"`
+
+	// IronicBootIsoSource
+	IronicBootIsoSource string `json:"ironicBootIsoSource,omitempty"`
```
kirankt, line 115: Not sure adding a new field is the way to go. I'm working on the installer changes and I think the current strategy is to send multiple URLs (live-iso, live-initramfs, live-rootfs, live-iso, openstack qcow2). Once these are set in the ProvisioningConfig, the CBO will pass it without change. Then in the ironic-rhcos-downloader, we'll have to unwrap it. If you want we can collaborate on how to get it done correctly.

kirankt: Eventually the idea is to remove the qcow2 completely.

elfosardo: @kirankt there may be some confusion here due to the name of the field/variable.
IronicBootIsoSource simply indicates if the boot iso image will be downloaded, cached, and served from the httpd server on the conductor node (local), or if it will be served by the original httpd server (http).
This is related to this ironic upstream change https://opendev.org/openstack/ironic/commit/4874f3c44e7b49714ca4fb158d2551b8f6f5fee4

hardys: @kirankt This isn't related to the new approach for using the live-iso+IPA, it's for the existing users of the BMH `format: live-iso` which currently caches images (e.g from assisted installer) by default.

I think probably https://github.com/openshift/ironic-image/pull/169  is sufficient, we'll disable caching unconditionally for that use-case, then this PR could be revisited in future if we discover a need to make it configurable.

stbenjam: I realize it's WIP but can we avoid saying Ironic? We try to keep the implementation detail hidden. This field also needs more documentation, and should probably be an enum type like ProvisioningNetwork, instead of an open-ended string.

api/v1alpha1/provisioning_types.go
```diff
@@ -110,6 +120,9 @@ type ProvisioningSpec struct {
 	// openshift-machine-api namespace. When set to true, this provisioning
 	// configuration would be used for baremetal hosts across all namespaces.
 	WatchAllNamespaces bool `json:"watchAllNamespaces,omitempty"`
+
+	// BootIsoSource
```
stbenjam, line 124: This should contain docs about the field. It's user-facing and shows up in the `explain` output for a CRD.

api/v1alpha1/provisioning_types.go
```diff
@@ -38,6 +38,16 @@ const (
...
 )
 
+// BootIsoSource is the origin of the boot iso image
+// +kubebuilder:validation:Enum=local;http
+type BootIsoSource string
+
+// BootIsoSource values
+const (
+	BootIsoSourceLocal BootIsoSource = "local"
+	BootIsoSourceHttp  BootIsoSource = "http"
```
stbenjam, line 48: Did we bike shed on the names we want to use?

`cache` and `direct` have more obvious meanings to me than `local` and `http`.

stbenjam:
`make generate` should fix the generate-check failure, the CRD needs to be updated

elfosardo:
@stbenjam thanks for the help!
Going to submit the changes ASAP

api/v1alpha1/provisioning_types.go
elfosardo: this makes sense, although the values of "local" and "http" can't be changed as they're interpreted by Ironic
or maybe I misunderstood and you meant to change BootIsoSourceLocal and BootIsoSourceHttp ?

elfosardo:
> `make generate` should fix the generate-check failure, the CRD needs to be updated

I updated the CRD but it seems still failing :/

api/v1alpha1/provisioning_types.go
asalkeld: I think it makes sense to name them after the ironic values

asalkeld (commented):
/approve

kirankt (approved):
/lgtm

PR #145, opened 2021-05-17
Bug 1947293: Document ipv6 provisioning network size limitation

squidboylan:
IPv6 provisioning networks cannot be larger than a /64 due to a limitation in dnsmasq.

### DIFF
--- a/config/crd/bases/metal3.io_provisionings.yaml
+++ b/config/crd/bases/metal3.io_provisionings.yaml
@@ -51,7 +51,7 @@ spec:
                 - Disabled
                 type: string
               provisioningNetworkCIDR:
-                description: ProvisioningNetworkCIDR is the network on which the baremetal nodes are provisioned. The provisioningIP and the IPs in the dhcpRange all come from within this network.
+                description: ProvisioningNetworkCIDR is the network on which the baremetal nodes are provisioned. The provisioningIP and the IPs in the dhcpRange all come from within this network. IPv6 networks cannot be larger than a /64.
                 type: string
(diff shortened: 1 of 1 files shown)


squidboylan:
The wording is a little confusing, but this is the clearest way to explain it to me, I'm happy to change it if others think there is a better way to explain it.

config/crd/bases/metal3.io_provisionings.yaml
```diff
@@ -51,7 +51,7 @@ spec:
                 - Disabled
                 type: string
               provisioningNetworkCIDR:
-                description: ProvisioningNetworkCIDR is the network on which the baremetal nodes are provisioned. The provisioningIP and the IPs in the dhcpRange all come from within this network.
+                description: ProvisioningNetworkCIDR is the network on which the baremetal nodes are provisioned. The provisioningIP and the IPs in the dhcpRange all come from within this network. IPv6 networks cannot be larger than a /64.
```
andfasano, line 54: Do you think it could be worth to add the same motivation introduced in the PR description also here (`dnsmasq`) for the sake of clarity?

squidboylan: It seems unnecessary to add this as I'm not sure users should care, but I'm happy to update it if you think it's appropriate.

stbenjam:
/approve
/override ci/prow/e2e-metal-ipi-ovn-ipv6 

Known flake: https://bugzilla.redhat.com/show_bug.cgi?id=1960758

PR #147, opened 2021-05-25
Bug 1962592: Use a cache URL with the .svc.cluster.local suffix

derekhiggins:
By default cluster.local is part of NoProxy, using the longer
version ensure we match the NoProxy variable and keep the traffic
internal.

Fixes: https://github.com/openshift/cluster-network-operator/issues/1105


kirankt (approved):
/approve

PR #148, opened 2021-05-27
Bug 1946079: pass IP_OPTIONS env down to the OS downloader container

asalkeld:
This should be equal to
"ip=dhcp" (when ipv4 only)
"ip=dhcp6" (when ipv6 only)
"" (when dual stack)

the downloader will then do the following
```
if [ -n "$IP_OPTIONS" ] ; then
    LIBGUESTFS_BACKEND=direct virt-edit -a "$RHCOS_IMAGE_FILENAME_QCOW" -m /dev/sda3 /boot/loader/entries/ostree-1-rhcos.conf -e "s/^options/options ${IP_OPTIONS}/"
fi
```

/cc @sadasu @honza 

This is an alternative to https://github.com/openshift/cluster-baremetal-operator/pull/143

### DIFF
--- a/controllers/provisioning_controller.go
+++ b/controllers/provisioning_controller.go
@@ -18,7 +18,10 @@ package controllers
 import (
 	"context"
 	"fmt"
+	"net"
+	"net/url"
 	"os"
+	"strings"
 	"time"
 
 	"github.com/pkg/errors"
@@ -60,6 +63,7 @@ type ProvisioningReconciler struct {
 	ReleaseVersion string
 	ImagesFilename string
 	WebHookEnabled bool
+	APIServerHost  string
 }
 
 type ensureFunc func(*provisioning.ProvisioningInfo) (bool, error)
@@ -184,13 +188,10 @@ func (r *ProvisioningReconciler) Reconcile(ctx context.Context, req ctrl.Request
 		return ctrl.Result{}, err
 	}
 
-	// Get cluster-wide proxy information
-	clusterWideProxy, err := r.OSClient.ConfigV1().Proxies().Get(context.Background(), "cluster", metav1.GetOptions{})
+	info, err := r.provisioningInfo(baremetalConfig, &containerImages)
 	if err != nil {
 		return ctrl.Result{}, err
 	}
-
-	info := r.provisioningInfo(baremetalConfig, &containerImages, clusterWideProxy)
 
 	// Check if Provisioning Configuartion is being deleted
 	deleted, err := r.checkForCRDeletion(ctx, info)
@@ -299,7 +300,26 @@ func (r *ProvisioningReconciler) Reconcile(ctx context.Context, req ctrl.Request
 	return result, nil
 }
 
-func (r *ProvisioningReconciler) provisioningInfo(provConfig *metal3iov1alpha1.Provisioning, images *provisioning.Images, proxy *osconfigv1.Proxy) *provisioning.ProvisioningInfo {
+func (r *ProvisioningReconciler) provisioningInfo(provConfig *metal3iov1alpha1.Provisioning, images *provisioning.Images) (*provisioning.ProvisioningInfo, error) {
+	clusterWideProxy, err := r.OSClient.ConfigV1().Proxies().Get(context.Background(), "cluster", metav1.GetOptions{})
+	if err != nil {
+		return nil, err
+	}
+
+	ips, err := net.LookupIP(r.APIServerHost)
+	if err != nil {
+		return nil, err
+	}
+
+	ns := provisioning.NetworkStackType(0)
+	for _, ip := range ips {
+		if len(ip) == net.IPv4len {
+			ns |= provisioning.NetworkStackV4
+		} else if len(ip) == net.IPv6len {
+			ns |= provisioning.NetworkStackV6
+		}
+	}
+
 	return &provisioning.ProvisioningInfo{
 		Client:        r.KubeClient,
 		EventRecorder: events.NewLoggingEventRecorder(ComponentName),
@@ -307,8 +327,9 @@ func (r *ProvisioningReconciler) provisioningInfo(provConfig *metal3iov1alpha1.P
 		Scheme:        r.Scheme,
 		Namespace:     ComponentNamespace,
 		Images:        images,
-		Proxy:         proxy,
-	}
+		Proxy:         clusterWideProxy,
+		NetworkStack:  ns,
+	}, nil
 }
 
 //Ensure Finalizer is present on the Provisioning CR when not deleted and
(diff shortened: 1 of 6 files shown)


controllers/provisioning_controller.go
```diff
@@ -299,16 +297,47 @@ func (r *ProvisioningReconciler) Reconcile(ctx context.Context, req ctrl.Request
...
+		} else if len(ip) == net.IPv6len {
+			haveV6 = true
+		}
+	}
+	ns := provisioning.NetworkStackV4
+	if haveV4 && haveV6 {
+		ns = provisioning.NetworkStackDual
+	} else if haveV6 {
+		ns = provisioning.NetworkStackV6
+	}
```
asalkeld, line 329: this seems really clumsy (and not tested), if anyone can think of a better way, please let me know...

honza:
/retitle Bug 1946079: pass IP_OPTIONS env down to the OS downloader container

controllers/provisioning_controller.go
```diff
@@ -299,16 +299,56 @@ func (r *ProvisioningReconciler) Reconcile(ctx context.Context, req ctrl.Request
 	return result, nil
 }
 
-func (r *ProvisioningReconciler) provisioningInfo(provConfig *metal3iov1alpha1.Provisioning, images *provisioning.Images, proxy *osconfigv1.Proxy) *provisioning.ProvisioningInfo {
+func apiServer() (string, error) {
+	config, err := ctrl.GetConfig()
+	if err != nil {
+		return "", err
```
sadasu, line 305: If an error is encountered at this point, CBO continues and tries again during the next Reconcile(). From my understanding it appears that this condition may not be fixed in a future reconcile.

controllers/provisioning_controller.go
```diff
@@ -299,16 +299,56 @@ func (r *ProvisioningReconciler) Reconcile(ctx context.Context, req ctrl.Request
...
+	}
+
+	apiServerURL, err := url.Parse(config.Host)
+	if err != nil {
+		return "", err
```
sadasu, line 310: Same comment as above.

provisioning/provisioning_info.go
```diff
@@ -9,6 +9,14 @@ import (
 	"github.com/openshift/library-go/pkg/operator/events"
 )
 
+type NetworkStackType int
+
+const (
+	NetworkStackV4   NetworkStackType = 1 << iota
+	NetworkStackV6   NetworkStackType = 1 << iota
+	NetworkStackDual NetworkStackType = (NetworkStackV4 & NetworkStackV6)
+)
```
sadasu, line 18: This is clever!

controllers/provisioning_controller.go
asalkeld: i'll move it to the SetupManager, so if it fails the pod will restart and if it succeeds the host will be cached

asalkeld:
the e2e-metal-ipi-ovn-ipv6 test is unrelated (failure of ingress operator), after chatting with @sadasu we are going to ignore it.
/override e2e-metal-ipi-ovn-ipv6

controllers/provisioning_controller.go
```diff
@@ -365,13 +386,35 @@ func (r *ProvisioningReconciler) deleteMetal3Resources(info *provisioning.Provis
...
+	}
+	host := config.Host
+	if strings.Contains(apiServerURL.Host, ":") {
+		host = strings.Split(apiServerURL.Host, ":")[0]
+	}
```
stbenjam, line 402: The IPv6 failures were real, and this is the source.  Core operators like ingress that fail often indicate workers didn't deploy, and workers didn't deploy since CBO didn't bring up Metal3. 

CBO logs:

``E0604 06:00:58.813292       1 controller.go:267] controller-runtime/manager/controller/provisioning "msg"="Reconciler error" "error"="lookup [fd02: no such host" "name"="provisioning-configuration" "namespace"="" "reconciler group"="metal3.io" "reconciler kind"="Provisioning" ``

The API host might be something like `[fd02::1]:6443` and this code is splitting it -- I think this is trying to remove the port? You'll get `[fd02` instead.

Here's an example using `SplitHostPort` that should work for everything: https://play.golang.org/p/6VqhuZQmtGc

stbenjam:
- It looks like CBO is getting an IP on the service network for the API VIP (see above comment, it's fd02::1), since it's using the SDN networking to reach the API VIP.  I'm not sure we're really determinining the right IP OPTIONS to pass here. The machine networks could be totally different IP version

- How is dualstack being detected? It looks like we're only ever seeing if it's IPv4 or IPv6.

provisioning/baremetal_pod.go
```diff
@@ -263,18 +263,30 @@ func createInitContainerIpaDownloader(images *Images) corev1.Container {
 	return initContainer
 }
 
-func createInitContainerMachineOsDownloader(images *Images, config *metal3iov1alpha1.ProvisioningSpec) corev1.Container {
+func createInitContainerMachineOsDownloader(info *ProvisioningInfo) corev1.Container {
+	optionValue := ""
+	switch info.NetworkStack {
```
dcbw, line 268: Isn't NetworkStack a bitfield due to https://github.com/openshift/cluster-baremetal-operator/pull/148/files#diff-7ad72d08a06fe28b03a15c84289b815f693443562d0c67e6f461c7e3f5ee0707R315? This isn't going to give the right result in dual-stack scenarios since neither option will match.

stbenjam: Right, we don't hit either case here, so it's left blank which is the behavior for dualstack.  The behavior here is more implicit, I think it'd be clearer if there was a case for NetworkStackDualstack.

asalkeld: https://github.com/openshift/cluster-baremetal-operator/pull/154

provisioning/image_cache.go
```diff
@@ -137,13 +137,14 @@ func newImageCacheContainers(images *Images, proxy *osconfigv1.Proxy) []corev1.C
...
 }
 
-func newImageCachePodTemplateSpec(targetNamespace string, images *Images, provisioningConfig *metal3iov1alpha1.ProvisioningSpec, proxy *osconfigv1.Proxy) (*corev1.PodTemplateSpec, error) {
-	cacheConfig, err := imageCacheConfig(targetNamespace, *provisioningConfig)
+func newImageCachePodTemplateSpec(info *ProvisioningInfo) (*corev1.PodTemplateSpec, error) {
+	cacheConfig, err := imageCacheConfig(info.Namespace, info.ProvConfig.Spec)
 	if err != nil {
 		return nil, err
 	}
+	info.ProvConfig.Spec = *cacheConfig
```
zaneb, line 145: This is very dangerous, because it sticks the locally-modified Spec that is used just for the image cache into the global ProvisioningInfo that is passed to every function. We're only getting away with it because setting up the image cache is currently the last function called.

asalkeld: https://github.com/openshift/cluster-baremetal-operator/pull/179

PR #149, opened 2021-06-01
Pass the node mac addresses to containers currently needing provisioningInterface

asalkeld:
Since we don't know what node it will be started on, pass all master node macs in the env as PROVISIONING_MACS

/hold


config/crd/bases/metal3.io_provisionings.yaml
```diff
@@ -51,7 +51,7 @@ spec:
                 - Disabled
                 type: string
               provisioningNetworkCIDR:
-                description: ProvisioningNetworkCIDR is the network on which the baremetal nodes are provisioned. The provisioningIP and the IPs in the dhcpRange all come from within this network. IPv6 networks cannot be larger than a /64.
```
squidboylan, line 54: Is there a reason this was removed? Maybe just a mistake during a merge conflict?

asalkeld: the version of controller-gen was bumped because of needing to import baremetal operator and this changed because of the controller-gen bump

kirankt (approved):
/lgtm

PR #150, opened 2021-06-03
Add VirtualMediaViaExternalNetwork flag to the Provisioning CRD

sadasu:


sadasu:
/hold
this is a 4.9 feature

api/v1alpha1/provisioning_types.go
```diff
@@ -54,6 +54,7 @@ type ProvisioningSpec struct {
 	// ProvisioningNetworkCIDR is the network on which the
 	// baremetal nodes are provisioned. The provisioningIP and the
 	// IPs in the dhcpRange all come from within this network.
+	// IPv6 networks cannot be larger than a /64.
```
sadasu, line 57: This change should have been part of https://github.com/openshift/cluster-baremetal-operator/pull/145. Adding it here so that the generated manifest contained the documentation change that was intended for the above PR. 
I also believe that this check for IPv6 network size could be added as an additional validation.

asalkeld (commented):
/lgtm

PR #151, opened 2021-06-04
Bug 1946079: Fix parsing of IPv6 APIserver IP address

sadasu:


controllers/provisioning_controller.go
```diff
@@ -398,7 +398,7 @@ func apiServerHost() (string, error) {
 	}
 	host := config.Host
 	if strings.Contains(apiServerURL.Host, ":") {
-		host = strings.Split(apiServerURL.Host, ":")[0]
+		host, _, _ = net.SplitHostPort(apiServerURL.Host)
```
stbenjam, line 401: What do you think about checking the error? It's probably relatively safe that url.Parse gave us back good data but just in case...

```suggestion
		host, _, err = net.SplitHostPort(apiServerURL.Host)
		if(err != nil) {
			return "", err
		}
```

controllers/provisioning_controller.go
```diff
@@ -398,7 +398,7 @@ func apiServerHost() (string, error) {
 	}
 	host := config.Host
 	if strings.Contains(apiServerURL.Host, ":") {
```
stbenjam, line 400: I think it's OK to skip this check, net.SplitHostPort will do the right thing if there's no port

controllers/provisioning_controller.go
sadasu: Good idea to check anyway. An ill formed API server address is unlikely but as you suggested, it is probably a good idea to check.

controllers/provisioning_controller.go
```diff
@@ -398,7 +398,7 @@ func apiServerHost() (string, error) {
 	}
 	host := config.Host
```
sadasu, line 399: This line can be skipped too, I think.

sadasu:
/retest
the IP_OPTIONS values are as expected.

PR #153, opened 2021-06-07
Bug 1968615: Restrict watches to openshift-machine-api

andfasano:
Previously CBO was watching over all the namespaces, thus caching all the resources with an high memory consumption (even those ones that didn't really belong to the component)

This PR fixes the issue by limiting the cache just to the component namespace `openshift-machine-api` (updates on cluster scoped resources as `Provisioning` will be received as usual)

Fixes #1968615


zaneb:
Nice, this solves the problem not only for Secrets but also Deployments and all the other stuff we watch :+1:
/lgtm

PR #157, opened 2021-06-09
WIP: Using the api-int URL from Infrastructure CR to determine protocol

sadasu:


sadasu:
Cannot fully test this until https://github.com/openshift/ironic-rhcos-downloader/pull/40 merges. This fix implements one of the alternatives discussed where we get the api-int URL from the Infrastructure CR. the api-int is on the machine network and protocol on that should be passes along to the ironic-rhcos-downloader container.

sadasu:
/closed
in favor of https://github.com/openshift/cluster-baremetal-operator/pull/158

PR #158, opened 2021-06-10
Bug 1946079: Fix network stack calculation

asalkeld:
There was at least 3 things wrong with how we figure out the NetworkStack

1. the NetworkStack enum was broken - blush
2. we were assuming that when len(ip) == IPv6Len it was an actual ipv6 address (you can have an embedded ipv4 address in a v6 one - news to me)
3. this is now using the internal apiserver url not, the public one

See https://github.com/openshift/ironic-rhcos-downloader/pull/40#issuecomment-858206600
This PR supersedes https://github.com/openshift/cluster-baremetal-operator/pull/154
and includes @sadasu commit in her PR https://github.com/openshift/cluster-baremetal-operator/pull/157
/cc @sadasu @honza @andfasano

### DIFF
--- a/controllers/provisioning_controller_test.go
+++ b/controllers/provisioning_controller_test.go
@@ -2,6 +2,8 @@ package controllers
 
 import (
 	"context"
+	"net"
+	"reflect"
 	"testing"
 
 	"github.com/stretchr/testify/assert"
@@ -13,6 +15,7 @@ import (
 	configv1 "github.com/openshift/api/config/v1"
 	fakeconfigclientset "github.com/openshift/client-go/config/clientset/versioned/fake"
 	metal3iov1alpha1 "github.com/openshift/cluster-baremetal-operator/api/v1alpha1"
+	"github.com/openshift/cluster-baremetal-operator/provisioning"
 )
 
 func setUpSchemeForReconciler() *runtime.Scheme {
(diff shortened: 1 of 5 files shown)


provisioning/provisioning_info.go
```diff
@@ -14,7 +14,7 @@ type NetworkStackType int
 const (
 	NetworkStackV4   NetworkStackType = 1 << iota
 	NetworkStackV6   NetworkStackType = 1 << iota
-	NetworkStackDual NetworkStackType = (NetworkStackV4 & NetworkStackV6)
+	NetworkStackDual NetworkStackType = (NetworkStackV4 | NetworkStackV6)
```
asalkeld, line 17: groan, how did I do this :facepalm:

controllers/provisioning_controller_test.go
```diff
@@ -144,3 +147,75 @@ func TestProvisioning(t *testing.T) {
...
+	}{
+		{
+			name: "v4 basic",
+			ips:  []net.IP{net.ParseIP("192.168.0.1")},
+			want: provisioning.NetworkStackV4,
+		},
+		{
+			name: "v4 in v6 format: basic",
+			ips:  []net.IP{net.ParseIP("::FFFF:192.168.0.1")},
+			want: provisioning.NetworkStackV4,
```
asalkeld, line 166: I am suspicious that we got this kind of address in the CI test for https://github.com/openshift/ironic-rhcos-downloader/pull/40#issuecomment-858206600

provisioning/provisioning_info.go
zaneb: I read this earlier and also did not notice.

zaneb (commented):
/approve

sadasu:
/lgtm
Nice addition of unit tests around this. Hopefully, that helped you catch some issues that were missed earlier.

sadasu (commented):
/lgtm
Nice addition of unit tests.

asalkeld:
1 test failed (not related, skipping)
```
: [sig-api-machinery] AdmissionWebhook [Privileged:ClusterAdmin] should be able to deny pod and configmap creation [Conformance] [Suite:openshift/conformance/parallel/minimal] [Suite:k8s] 
```
/override ci/prow/e2e-metal-ipi

provisioning/baremetal_pod.go
```diff
@@ -263,15 +263,20 @@ func createInitContainerIpaDownloader(images *Images) corev1.Container {
...
 		optionValue = "ip=dhcp"
 	case NetworkStackV6:
 		optionValue = "ip=dhcp6"
+	case NetworkStackDual:
+		optionValue = ""
 	}
+	return optionValue
+}
 
+func createInitContainerMachineOsDownloader(info *ProvisioningInfo) corev1.Container {
```
zaneb, line 279: This function is used by both the os-downloader in the metal3 pod and the one in the image cache, so we end up applying the command line change twice before the image reaches the host.
This _probably_ doesn't break anything, since it works by doing `s/^options/options ip=dhcp/` (in perl) but it's not ideal.

zaneb: Actually, I think it might break something: the checksum will be different depending on which host the API VIP is on. There might be a scenario in which the image and checksum get read either side of an API VIP failover, in which case they won't match.
I don't know that we test without a provisioning network in CI so this never gets covered.
I opened #163 to fix this.

PR #163, opened 2021-06-16
Don't set IP options on kernel command line twice

zaneb:
We only want to pass the IP_OPTIONS environment variable to the
machine-os-downloader container in the metal3 Deployment, not the ones
in the metal3-image-cache DaemonSet.

If we pass it to both then the options get added twice on some cache
Pods (the ones not running on the Node that has the API VIP), with the
result that the checksum will be different depending where the API VIP
is pointing.


stbenjam:
This sounds like something we should backport to 4.8

asalkeld (commented):
/lgtm

honza:
Nice catch, lgtm

PR #165, opened 2021-06-17
Bug 1972753: Only start static ip set if provisioning net not disabled

derekhiggins:
Only add metal3-static-ip-set if we have a provisioning ip and
and provisioning network is enabled. This brings the check for
both metal3-static-ip-set and metal3-static-ip-manager inline
with each other. It doesn't make sense to have one without the
other.


sadasu (commented):
/approve

PR #166, opened 2021-06-24
Bug 1975711: Only start static ip set if provisioning net not disabled

derekhiggins:
Only add metal3-static-ip-set if we have a provisioning ip and
and provisioning network is enabled. This brings the check for
both metal3-static-ip-set and metal3-static-ip-manager inline
with each other. It doesn't make sense to have one without the
other.


sadasu (approved):

asalkeld (commented):
/retest
/lgtm

asalkeld (commented):
/lgtm

sdodson:
Re-validating bugzilla status after prow config updates.
/bugzilla refresh

mfojtik:
[patch-manager] :rocket: Approved for z-stream by score: 0.20

adding cherry-pick approved

PR #169, opened 2021-06-29
Bug 1973724: reorder the initContainers, so that static-ip-set happens prior to the image download

asalkeld:
/cc @zaneb


zaneb:
The only concern with  doing this was that @hardys [mentioned](https://bugzilla.redhat.com/show_bug.cgi?id=1973724#c6) there is a 5-minute timeout after which the provisioning VIP will fail over if we haven't finished downloading the image and started the pod. Per Steve's comment there though, it should be fair to assume that 5 minutes is enough.

Theoretically, because the images are now stored in the host filesystem, it should be possible to run this init container outside of the Pod. However, knowing when we need to re-run it (when the image changes? when an image-cache pod fails?) is challenging.

/lgtm
/retest

PR #172, opened 2021-06-30
Pass IRONIC_EXTERNAL_IP to Ironic container

sadasu:
If VirtualMediaViaExternalNetwork is set and the ProvisioningNetwork
is not Disabled, remote workers on the external network need to use
the host IP of the master node on the external network to talk back to
Ironic.


sadasu:
Current e2e-metal-ipi failures should be taken care of by https://github.com/openshift/machine-config-operator/pull/2650.

imain:
LGTM!  I've tested this change now and it works as advertised.  Be great to get this in soon.

kirankt (approved):
/lgtm

PR #174, opened 2021-07-07
Add support for live images and configuration of coreos IPA

kirankt:
This PR adds support to embed IPA ignition into a coreos live ISO. See: openshift/ironic-image#188

It also adds support to either fetch live-iso or the old qcow2 based URLs.

Support is also added to send DEPLOY_ISO_URL to BMO


provisioning/baremetal_pod.go
```diff
@@ -293,7 +321,7 @@ func createInitContainerMachineOsDownloader(info *ProvisioningInfo, setIpOptions
 	initContainer := corev1.Container{
 		Name:            "metal3-machine-os-downloader",
 		Image:           info.Images.MachineOsDownloader,
-		Command:         []string{"/usr/local/bin/get-resource.sh"},
+		Command:         []string{"/usr/local/bin/get-live-images.sh"},
```
dtantsur, line 324: You need to change how `env` is built above.

provisioning/baremetal_pod.go
```diff
@@ -266,6 +267,33 @@ func createInitContainerIpaDownloader(images *Images) corev1.Container {
 	return initContainer
 }
 
+func createInitContainerCoreOSIPA(info *ProvisioningInfo) corev1.Container {
+	config := &info.ProvConfig.Spec
+	initContainer := corev1.Container{
+		Name:            "configure-coreos-ipa",
```
dtantsur, line 273: maybe prefix with `metal3-` for consistency and discoverability?

kirankt: +1. Makes sense.

provisioning/baremetal_pod.go
kirankt: Can you please elaborate why we need to rebuild the `env`? The image URL is sent via the installer to CBO via provisioning config and we control what gets sent here.

provisioning/baremetal_pod.go
```diff
@@ -300,6 +328,14 @@ func createInitContainerMachineOsDownloader(info *ProvisioningInfo, setIpOptions
 	env := []corev1.EnvVar{
 		buildEnvVar(machineImageUrl, &info.ProvConfig.Spec),
 	}
+	imageURL := getMetal3DeploymentConfig(machineImageUrl, &info.ProvConfig.Spec)
```
dtantsur, line 331: nit: you can probably use the result of `buildEnvVar` above

provisioning/baremetal_pod_test.go
```diff
@@ -112,6 +112,10 @@ func TestNewMetal3InitContainers(t *testing.T) {
 					Name:  "metal3-static-ip-set",
 					Image: images.StaticIpManager,
 				},
+				{
+					Name:  "configure-coreos-ipa",
```
dtantsur, line 116: Hmm, why is the old name used here?

dtantsur (changes requested):
Please also update createContainerMetal3BaremetalOperator to pass `DEPLOY_ISO`.

provisioning/baremetal_pod_test.go
kirankt: Doh. Fixed.

kirankt:
> Please also update createContainerMetal3BaremetalOperator to pass `DEPLOY_ISO`.
Should this be DEPLOY_ISO_URL env variable pointing to the cached iso image?

provisioning/baremetal_pod.go
```diff
@@ -258,28 +257,35 @@ func newMetal3InitContainers(info *ProvisioningInfo) []corev1.Container {
...
-		Image:           images.IpaDownloader,
-		Command:         []string{"/usr/local/bin/get-resource.sh"},
+		Name:            "metal3-configure-coreos-ipa",
+		Image:           info.Images.Ironic,
+		Command:         []string{"/bin/configure-coreos-ipa"},
 		ImagePullPolicy: "IfNotPresent",
 		SecurityContext: &corev1.SecurityContext{
 			Privileged: pointer.BoolPtr(true),
 		},
 		VolumeMounts: []corev1.VolumeMount{imageVolumeMount},
```
dtantsur, line 275: You may also need to mount the TLS volume so that it knows that TLS is used.

dtantsur: Ref: https://github.com/openshift/ironic-image/blob/master/scripts/configure-coreos-ipa#L15

provisioning/image_cache.go
```diff
@@ -77,9 +76,32 @@ func imageCacheConfig(targetNamespace string, config metal3iov1alpha1.Provisioni
...
+		var allURLs string
+		URLs := strings.Split(config.ProvisioningOSDownloadURL, ",")
+		for _, URL := range URLs {
+			newURL, err := transformURL(targetNamespace, URL)
+			if err != nil {
+				return nil, err
+			}
+			allURLs = allURLs + newURL + ","
+		}
+		config.ProvisioningOSDownloadURL = strings.TrimSuffix(allURLs, ",")
```
zaneb, line 96: nit: you could add them to a slice and then use strings.Join() instead.

provisioning/image_cache.go
```diff
@@ -77,9 +76,32 @@ func imageCacheConfig(targetNamespace string, config metal3iov1alpha1.Provisioni
 		Scheme: "http",
 		Host: net.JoinHostPort(fmt.Sprintf("%s.%s.svc.cluster.local", stateService, targetNamespace),
 			baremetalHttpPort),
-		Path: fmt.Sprintf("/images/%s/%s", imageName, imageName),
+		Path: fmt.Sprintf("/images/%s/cached-%s", imageName, imageName),
```
zaneb, line 79: Is this correct for both qcow2 and live-iso? I would guess that it's not returning the correct thing when there is no -live image, because in that case the machine-os-downloader container is still running the old `get-resource.sh` and not the new `get_live_images.sh`.

Also, this doesn't match the comment on line 73, so please update it.

provisioning/baremetal_pod.go
```diff
@@ -297,9 +308,20 @@ func ipOptionForMachineOsDownloader(info *ProvisioningInfo) string {
 }
 
 func createInitContainerMachineOsDownloader(info *ProvisioningInfo, setIpOptions bool) corev1.Container {
+	imageURL := getMetal3DeploymentConfig(machineImageUrl, &info.ProvConfig.Spec)
 	env := []corev1.EnvVar{
-		buildEnvVar(machineImageUrl, &info.ProvConfig.Spec),
+		{
+			Name:  machineImageUrl,
+			Value: *imageURL,
+		},
```
zaneb, line 316: It's not clear to me that this is different from what `buildEnvVar()` does?

provisioning/image_cache.go
kirankt: TIL. Thanks. I'll make the change.

provisioning/baremetal_pod.go
kirankt: Just trying to reuse the imageURL variable. I 'll move it back to buildEnvVar. Its more readable.

provisioning/image_cache.go
kirankt: The filename prefixed with `cached` is the original name. The one without is a symlink. So, this should be OK.

https://github.com/openshift/ironic-rhcos-downloader/blob/d4f21a6e47399ae796a7c0ea80c211b8405d0fb1/get-resource.sh#L104

provisioning/baremetal_pod.go
```diff
@@ -449,6 +496,7 @@ func createContainerMetal3BaremetalOperator(images *Images, config *metal3iov1al
 			},
 			buildEnvVar(deployKernelUrl, config),
 			buildEnvVar(deployRamdiskUrl, config),
+			buildEnvVar(deployIsoUrl, config),
```
dtantsur, line 499: Given what we discussed last week, I wonder if we should skip this part, so that Ironic keeps building an ISO itself, using the kernel and ramdisk provided. Maybe as a follow-up patch if this one is already tested to work with virtual media.

provisioning/image_cache.go
```diff
@@ -60,12 +61,10 @@ func imageVolume() corev1.Volume {
 	}
 }
 
-func imageCacheConfig(targetNamespace string, config metal3iov1alpha1.ProvisioningSpec) (*metal3iov1alpha1.ProvisioningSpec, error) {
-	// The download URL looks something like:
-	// https://releases-art-rhcos.svc.ci.openshift.org/art/storage/releases/rhcos-4.2/42.80.20190725.1/rhcos-42.80.20190725.1-openstack.qcow2?sha256sum=123
```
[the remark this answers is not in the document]
kirankt: I agree about the lack of comments. I've tried to add a few in the new code and it hopefully helps.

provisioning/baremetal_pod.go
```diff
@@ -258,28 +260,64 @@ func newMetal3InitContainers(info *ProvisioningInfo) []corev1.Container {
...
 
-func createInitContainerIpaDownloader(images *Images) corev1.Container {
+// getRegistryPullSecret fetches the registry config auth file which is then passed
+// as an env variable to the Ironi-Image's configure-coreos-ipa endpoint. Note that
+// we cannot use LocalObjectReference because the secret is in a different NS.
+func getRegistryPullSecret(info *ProvisioningInfo, targetNamespace string) string {
+	// Check whether the Client is present before using it
+	if info.Client != nil {
+		secretClient := info.Client.CoreV1().Secrets(targetNamespace)
+		secret, err := secretClient.Get(context.TODO(), pullSecretName, metav1.GetOptions{})
```
zaneb, line 275: IIUC since 887809db97969ea5eaccacb5713a42a8f13edda3 this will never be in the cache.

Obviously you tried this and it worked. @andfasano does that mean we're always reading directly from the API (i.e. for everything except the Provisioning resource itself) and never the cache?

provisioning/image_cache.go
kirankt: Ahh, I see the problem. Yes, this will fail spectacularly, if we always prefix with `cache-`

kirankt: 🤦 :
```
sh-4.4# ls -l /shared/html/images/cached-rhcos-48.84.202105190318-0-live-initramfs.x86_64.img/
total 85944
-rw-r--r--. 1 root root 88003520 Jul 19 23:16 cached-cached-rhcos-48.84.202105190318-0-live-initramfs.x86_64.img
```

kirankt: I'll open a PR on ironic-rhcos-downloader and add the symlink for the live images as well. That should fix this problem. Thanks for catching this one and clarifying it for me.

kirankt: PR here: https://github.com/openshift/ironic-rhcos-downloader/pull/59

zaneb:
After a discussion today about upgrades, we concluded that it probably does not make sense to overload the meaning of the `ProvisioningOSDownloadURL` field. Ultimately the IPA image and the Provisioning image are different things and would be best represented by separate fields. There may be a release or two where we need to specify both, and we may ultimately want users to be able to edit the URLs, particularly in the case of a disconnected environment.

Potentially a structure like:

```
PreprovisioningOSDownloadURLs struct {
    ISO    string
    Kernel string
    InitRD string
    RootFS string
}
```

would be the easiest to read/write.

To ensure that the two things are orthogonal we'll effectively need two separate `metal3-machine-os-downloader` init containers, using the `get-resource.sh` and `get-live-images.sh` endpoints, respectively, with each only enabled if there is a relevant image URL provided.

Finally, it is not yet clear how we can handle upgrades from releases where the IPA URLs are not provided. At least until we figure it out (so in this PR at a minimum), we'll want to continue including a `metal3-ipa-downloader` init container and providing its built-in kernel+ramdisk to BMO whenever the IPA kernel+initrd+rootfs URLs are not all present in the Provisioning resource.

provisioning/baremetal_pod.go
sadasu: Yes, this Secret in the a Namespace other than "openshift-machine-api" so it will be read directly from the API server and not from the cache.

provisioning/baremetal_pod.go
```diff
@@ -258,28 +260,64 @@ func newMetal3InitContainers(info *ProvisioningInfo) []corev1.Container {
...
+		if err != nil {
+			return ""
+		}
+		// Return the value of .dockerconfigjson key
+		return base64.StdEncoding.EncodeToString(secret.Data[openshiftConfigSecretKey])
```
sadasu, line 280: There is a reason this data is in a Secret. We are reading it and passing its contents to Ironic. Is that OK?

sadasu:
The ironic-ipa-downloader container image is still part of the release unless we remove it from here https://github.com/openshift/cluster-baremetal-operator/blob/master/manifests/image-references#L25. I understand this PR will not include removal of this image or the metal3-ipa-downloader container.

sadasu:
Can we update this commit to just add the initContainer as the commit message suggests. Can we add the changes to the image-cache logic as a seperate commit/PR?

provisioning/baremetal_pod.go
kirankt: If we don't pass this then the IPA won't be able to pull the ironic-agent image from the "official" registry. It'll fall back to Dmitry's quay.io version, which is not OK for an official release.

provisioning/baremetal_pod.go
sadasu: https://github.com/openshift/ironic-image/pull/196 should take care of this above concern I raised.

provisioning/baremetal_pod.go
sadasu: openshift/ironic-image#196 should take care of this above concern.

kirankt: We will still be able to read it from oc describe pod metal3..., right?

kirankt:
Looks like these CI errors are due to the healtz port conflict between BMO and CAPBM pods landing on the same master.

provisioning/image_cache.go
```diff
@@ -138,11 +167,17 @@ func newImageCacheContainers(images *Images, proxy *osconfigv1.Proxy) []corev1.C
 }
 
 func newImageCachePodTemplateSpec(info *ProvisioningInfo) (*corev1.PodTemplateSpec, error) {
-	cacheConfig, err := imageCacheConfig(info.Namespace, info.ProvConfig.Spec)
+	imageURL, err := imageCacheConfig(info.Namespace, info.ProvConfig.Spec)
 	if err != nil {
 		return nil, err
 	}
-	info.ProvConfig.Spec = *cacheConfig
```
zaneb, line 145: Glad we got rid of this: https://github.com/openshift/cluster-baremetal-operator/pull/148#discussion_r674993687

provisioning/image_cache.go
```diff
@@ -138,11 +167,17 @@ func newImageCacheContainers(images *Images, proxy *osconfigv1.Proxy) []corev1.C
...
+
+	initContainers := []corev1.Container{
+		createInitContainerMachineOsDownloader(info, *imageURL, false),
```
zaneb, line 176: The same logic is required as in the metal3 Deployment - we may need two MachineOsDownloader containers.

provisioning/image_cache.go
```diff
@@ -60,27 +61,55 @@ func imageVolume() corev1.Volume {
...
 	}
 	imageName := path.Base(fileCompressionSuffix.ReplaceAllString(downloadURL.Path, ""))
 
-	// The first level of cache downloads the file and makes an uncompressed
-	// version of it available to this second-level cache somewhere like:
-	// http://metal3-state.openshift-machine-api:6180/images/rhcos-42.80.20190725.1-openstack.qcow2/rhcos-42.80.20190725.1-openstack.qcow2
+	// The first level of cache downloads the file and makes an version of it available
+	// to this second-level cache which then transforms it into the following format:
+	// Original format: http://<URL>:6180/images/rhcos-42.80.20190725.1-openstack.qcow2/cached-rhcos-42.80.20190725.1-openstack.qcow2
+	// Transformed format: http://metal3-state.openshift-machine-api:6181/images/rhcos-42.80.20190725.1-openstack.qcow2/cached-rhcos-42.80.20190725.1-openstack.qcow2
```
zaneb, line 76: I don't believe either of these is correct. The _original_ format (i.e. the URL that is being passed in to this function) is something like:

```
https://releases-art-rhcos.svc.ci.openshift.org/art/storage/releases/rhcos-4.2/42.80.20190725.1/rhcos-42.80.20190725.1-openstack.qcow2.gz?sha256sum=123
```

We then transform it into:

```
http://metal3-state.openshift-machine-api.svc.cluster.local:6180/images/rhcos-42.80.20190725.1-openstack.qcow2/cached-rhcos-42.80.20190725.1-openstack.qcow2
```

which is where the first-level cache serves it from. Finally, the second-level cache will make it available at:

```
http://cluster.local:6181/images/rhcos-42.80.20190725.1-openstack.qcow2/cached-rhcos-42.80.20190725.1-openstack.qcow2
```

The previous comments were considerably more accurate, although neither was entirely correct.

provisioning/image_cache.go
```diff
@@ -60,27 +61,55 @@ func imageVolume() corev1.Volume {
...
 		Path: fmt.Sprintf("/images/%s/%s", imageName, imageName),
 	}
-	config.ProvisioningOSDownloadURL = cacheURL.String()
-	return &config, nil
+	return cacheURL.String(), nil
+}
+
+func imageCacheConfig(targetNamespace string, config metal3iov1alpha1.ProvisioningSpec) (*string, error) {
+	// If the URL contains a comma-seperated list and are live images, split and transform the URLs
+	if getPreProvisioningOSDownloadURLs(&config) != nil {
```
zaneb, line 89: Maybe this function should just return a list of strings so we don't effectively have to write it twice.

provisioning/image_cache.go
```diff
@@ -60,27 +61,55 @@ func imageVolume() corev1.Volume {
...
+		URLs := []string{config.PreProvisioningOSDownloadURLs.InitramFsURL,
+			config.PreProvisioningOSDownloadURLs.KernelURL,
+			config.PreProvisioningOSDownloadURLs.RootFsURL,
+			config.PreProvisioningOSDownloadURLs.IsoURL}
+		var newURLs []string
+		for _, URL := range URLs {
```
zaneb, line 95: I think you meant to add a check for `URL == ""` about here.

provisioning/baremetal_pod.go
kirankt: I was looking at Secret selectors only work in the current namespace. I didn't find a way to input the value into the env without this method. One potential solution would be to copy the registry secret into the current ns and then use the LocalObjectReference to read it.

api/v1alpha1/provisioning_types.go
```diff
@@ -48,6 +48,22 @@ const (
...
+	IsoURL string `json:"iso,omitempty"`
+
+	// KernelURL is an Image URL to be used for PXE deployments
+	KernelURL string `json:"kernel,omitempty"`
+
+	// InitramFsURL Image URL to be used for PXE deployments
+	InitramFsURL string `json:"initramFs,omitempty"`
+
+	// RootFsURL Image URL to be used for PXE deployments
+	RootFsURL string `json:"rootFs,omitempty"`
```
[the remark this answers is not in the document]
kirankt: +1. @sadasu and I agreed that these names were placeholders and we intended to change them today. Just stepping back and looking at the `Fs` does  look funky. Will change.

provisioning/baremetal_pod.go
```diff
@@ -257,11 +259,37 @@ func newMetal3InitContainers(info *ProvisioningInfo) []corev1.Container {
...
 
-	initContainers = append(initContainers, createInitContainerMachineOsDownloader(info, true))
+	// If the PreProvisioningOSDownloadURLs are not empty then we download the IPA
+	liveURLs := getPreProvisioningOSDownloadURLs(&info.ProvConfig.Spec)
+	if liveURLs != nil {
+		initContainers = append(initContainers, createInitContainerMachineOsDownloader(info, *liveURLs, true))
+		// Configure the LiveISO by embedding ignition and other startup files
+		initContainers = append(initContainers, createInitContainerCoreOSIPA(info))
+	} else {
+		initContainers = append(initContainers, createInitContainerIpaDownloader(info.Images))
```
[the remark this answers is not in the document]
sadasu: Thanks for clarifying @zaneb! We did have questions around that.

provisioning/image_cache.go
kirankt: Done. Thank you.

provisioning/baremetal_pod.go
```diff
@@ -257,11 +259,37 @@ func newMetal3InitContainers(info *ProvisioningInfo) []corev1.Container {
...
+		initContainers = append(initContainers, createInitContainerMachineOsDownloader(info, info.ProvConfig.Spec.ProvisioningOSDownloadURL, true))
```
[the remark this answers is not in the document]
sadasu: So, does that mean that when both the qcow2 image is provided in the ProvisioningOSDownloadURL **and** the Preprovisioning URLs are present, we want to download all of them?
In this version of the implementation we are assuming that the Preprovisioning URLs take precedence and when they are present, we essentially ignore the ProvisioningOSDownloadURL.

provisioning/image_cache.go
kirankt: Thanks for the context. This line was a head scratcher to me too. I was worried about imagecache changing the spec also.

provisioning/baremetal_pod.go
```diff
@@ -283,6 +311,45 @@ func createInitContainerIpaDownloader(images *Images) corev1.Container {
 	return initContainer
 }
 
+// This initContainer configures RHCOS Live ISO images by embedding the IPA
+// agent ignition. See:
+// https://github.com/openshift/ironic-image/blob/master/scripts/configure-coreos-ipa
+func createInitContainerCoreOSIPA(info *ProvisioningInfo) corev1.Container {
```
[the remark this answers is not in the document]
kirankt: Oh yeah. That make sense. These functions are so long, I hadn't paid attention to its name... Thanks. Will fix it.

provisioning/baremetal_pod.go
sadasu: > We will still be able to read it from oc describe pod metal3..., right?
When I run "oc describe pod <metal3-pod> -n openshift-machine-api", I see this for a secret passed in via an env var:
HTTP_BASIC_HTPASSWD:     <set to the key 'htpasswd' in secret 'metal3-ironic-inspector-password'>  Optional: false

@kirankt maybe this password also shows up in a similar way. Could you pls confirm?

kirankt: No. This API is not available for Secrets outside the current namespace.

provisioning/image_cache.go
```diff
@@ -197,7 +197,7 @@ func newImageCachePodTemplateSpec(info *ProvisioningInfo) (*corev1.PodTemplateSp
 				},
 			},
 			InitContainers: injectProxyAndCA([]corev1.Container{
-				createInitContainerMachineOsDownloader(info, false),
+				createInitContainerMachineOsDownloader(info, info.ProvConfig.Spec.ProvisioningOSDownloadURL, false),
 				createInitContainerIpaDownloader(info.Images),
```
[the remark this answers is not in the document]
kirankt: We made the change last night after seeing that the `createInitContainerMachineOsDownloader` after being called from two different contexts. So we added a parameter to remove the inference.

provisioning/image_cache.go
kirankt: I think adding the parameter fixes this issue. Correct?

provisioning/image_cache.go
kirankt: The comment makes use of the word 'Original', which is unfortunate not to mention inaccurate. I should have mentioned first level cache url instead of the original. I'll rewrite the comment.

provisioning/image_cache.go
kirankt: I think so. That make sense. I'll make the change.

provisioning/image_cache.go
kirankt: Thanks for clarifying the comments. I've updated them.

provisioning/image_cache.go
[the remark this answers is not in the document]
zaneb: I meant the IPA downloader. So our upgrades situation isn't quite as bad as I thought - after an upgrade we will always have an up-to-date IPA from the previous release. Although we'd still have no way to update it without continuing to ship the IPA downloader, and any new masters added to the cluster would not get it.

provisioning/baremetal_pod.go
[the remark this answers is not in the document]
zaneb: Yes. The other URLs are for _Pre_-provisioning. We also need the qcow2 image for _Provisioning_ if we are to keep provisioning from existing MachineSets. They are orthogonal.

Only the CoreOS kernel+initramfs+rootfs vs. the IPADownloader are alternatives to each other.

See https://github.com/openshift/cluster-baremetal-operator/pull/174#issuecomment-883704094

provisioning/baremetal_pod.go
zaneb: > I was looking at Secret selectors only work in the current namespace. I didn't find a way to input the value into the env without this method. One potential solution would be to copy the registry secret into the current ns and then use the LocalObjectReference to read it.

Yes, that's the obvious way.

We also should watch that pull secret in case it changes. That is going to be fun...

provisioning/baremetal_pod.go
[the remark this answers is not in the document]
sadasu: Thanks @zaneb! I had forgotten about the existing Machinesets.

provisioning/baremetal_pod.go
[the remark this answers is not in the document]
zaneb: This is in the wrong place now. It should only be added if we don't have the live URLs.

provisioning/image_cache.go
zaneb: If it's not called twice then it isn't fixed.
Also the parameter hasn't been added yet.

provisioning/baremetal_secrets.go
```diff
@@ -135,6 +138,32 @@ password = %s
...
+		return err
+	}
+
+	secret := &corev1.Secret{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      pullSecretName,
+			Namespace: info.Namespace,
+		},
+		StringData: map[string]string{
+			openshiftConfigSecretKey: string(openshiftConfigSecret.Data[openshiftConfigSecretKey]),
```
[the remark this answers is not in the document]
kirankt: So, the value of openshiftConfigSecret.Data[openshiftConfigSecretKey] is just a []byte which turns out to be a plain string. I made this same mistake by requesting @dtantsur to change here: https://github.com/openshift/ironic-image/pull/195#discussion_r671404017. 

In any case, I *think* the opaque data returned via the env variable is base64 encoded, so no double encoding going on here. In any case, in this iteration, I'm trying not to break the qcow2 support that exist in the repo today.

api/v1alpha1/provisioning_types.go
```diff
@@ -143,6 +159,11 @@ type ProvisioningSpec struct {
 	// PXE deployments will always use the Provisioning Network and will not be
 	// affected by this flag.
 	VirtualMediaViaExternalNetwork bool `json:"virtualMediaViaExternalNetwork,omitempty"`
+
+	// PreprovisioningOSDownloadURLs set of boot Image URLs that would be necessary to provision a worker
+	// either using virtual media or PXE. The PXE URLs should be accessible from the
+	// provisioning network.
```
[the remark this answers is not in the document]
kirankt: I did not modify that comment. We should fix it in a different PR.

provisioning/image_cache.go
kirankt: I overhauled the image cache code a bit to fix this. PTAL. Thanks

api/v1alpha1/provisioning_types.go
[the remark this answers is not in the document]
zaneb: This is new code in this PR.

provisioning/baremetal_config.go
```diff
@@ -95,6 +95,33 @@ func getProvisioningOSDownloadURL(config *metal3iov1alpha1.ProvisioningSpec) *st
...
 
+// Check whether the PreProvisionOSDownloadURLs are set. If yes, we
+// construct a comma-separated list of RHCOS live images and return it
+func getPreProvisioningOSDownloadURLs(config *metal3iov1alpha1.ProvisioningSpec) []string {
+	var liveURLs []string
+	if config.PreProvisioningOSDownloadURLs.InitramfsURL != "" {
+		liveURLs = append(liveURLs, config.PreProvisioningOSDownloadURLs.InitramfsURL)
+	}
+	if isCoreOSIPAAvailable(config) {
+		liveURLs = append(liveURLs, config.PreProvisioningOSDownloadURLs.IsoURL)
```
zaneb, line 106: This should be InitramfsURL.

provisioning/baremetal_config.go
```diff
@@ -95,6 +95,33 @@ func getProvisioningOSDownloadURL(config *metal3iov1alpha1.ProvisioningSpec) *st
 	return nil
 }
 
+// Check whether the PreProvisionOSDownloadURLs are set. If yes, we
+// construct a comma-separated list of RHCOS live images and return it
+func getPreProvisioningOSDownloadURLs(config *metal3iov1alpha1.ProvisioningSpec) []string {
+	var liveURLs []string
+	if config.PreProvisioningOSDownloadURLs.InitramfsURL != "" {
+		liveURLs = append(liveURLs, config.PreProvisioningOSDownloadURLs.InitramfsURL)
```
zaneb, line 103: This should be IsoURL on both of the 2 lines above.

provisioning/baremetal_pod.go
```diff
@@ -257,7 +268,27 @@ func newMetal3InitContainers(info *ProvisioningInfo) []corev1.Container {
 		initContainers = append(initContainers, createInitContainerStaticIpSet(info.Images, &info.ProvConfig.Spec))
 	}
 
-	initContainers = append(initContainers, createInitContainerMachineOsDownloader(info, true))
+	// If the PreProvisioningOSDownloadURLs are set, we fetch the URLs of either CoreOS ISO and IPA assets or in some
+	// cases only the IPA assets
+	liveURLs := getPreProvisioningOSDownloadURLs(&info.ProvConfig.Spec)
+	if liveURLs != nil {
```
zaneb, line 274: It's always non-nil now (which is good!), so this should be `if len(liveURLSs) > 0`

provisioning/baremetal_pod.go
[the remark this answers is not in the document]
zaneb: :+1:

provisioning/baremetal_secrets.go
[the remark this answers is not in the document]
zaneb: OK, thanks. The whole Secret interface is a giant foot-gun :/ To this day I don't understand why it is base64-encoded. Anything you can put in a string and base64-encode you can also... put in a string.

provisioning/image_cache.go
```diff
@@ -129,6 +147,35 @@ func createContainerImageCache(images *Images) corev1.Container {
 	return container
 }
 
+func newImageCacheInitContainers(info *ProvisioningInfo) ([]corev1.Container, error) {
+	initContainers := []corev1.Container{}
+	// If the PreProvisioningOSDownloadURLs are set, we fetch the URLs of either CoreOS ISO and IPA URLs or in some
+	// cases only the IPA URLs
+	liveURLs := getPreProvisioningOSDownloadURLs(&info.ProvConfig.Spec)
+	if liveURLs != nil {
```
zaneb, line 155: check for length >0

api/v1alpha1/provisioning_types.go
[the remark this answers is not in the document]
sadasu: @zaneb I added this comment in https://github.com/openshift/cluster-baremetal-operator/pull/175.
I agree that the MachineOSDownloader should be able to pull them from any network.
The reason, I added this was:
As of today, we cannot PXE boot workers if they are not on the provisioning network. Does that still hold true for PXE booting via PreProvisioning PXE boot URLs? If so, would a better comment be "while using the Kernel, initramfs and rootfs URLs to PXE boot, the workers should be on the Provisioning network and that requirement does not exist if the workers are using the LiveISO URL to boot via virtualmedia".

provisioning/baremetal_config.go
sadasu, line 103: Should this be IsoURL?

provisioning/baremetal_config.go
sadasu, line 106: Should this be initramfsURL?

provisioning/image_cache.go
```diff
@@ -129,6 +147,35 @@ func createContainerImageCache(images *Images) corev1.Container {
...
+		newURLs, err := getTransformedURLs(liveURLs, info.Namespace)
+		if err != nil {
+			return nil, err
+		}
+		initContainers = append(initContainers, createInitContainerMachineOsDownloader(info, strings.Join(newURLs, ","), true, false))
+	}
+	// Download the transformed URL containing qcow2
+	if info.ProvConfig.Spec.ProvisioningOSDownloadURL != "" {
+		origURLs := []string{info.ProvConfig.Spec.ProvisioningOSDownloadURL}
+		newURLs, err := getTransformedURLs(origURLs, info.Namespace)
```
zaneb, line 165: nit: could just call transformURL here.

provisioning/baremetal_config.go
kirankt: Yes. Correct. I meant IsoURL in both places.

provisioning/baremetal_config.go
kirankt: Correct.

kirankt: Yes. Correct.

provisioning/baremetal_config.go
kirankt: Yes. Correct. I meant IsoURL in both places.

provisioning/baremetal_pod.go
kirankt: Changed.

api/v1alpha1/provisioning_types.go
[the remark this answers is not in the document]
zaneb: > @zaneb I added this comment in #175.
We're not planning to merge that though, right?

> As of today, we cannot PXE boot workers if they are not on the provisioning network. Does that still hold true for PXE booting via PreProvisioning PXE boot URLs?

Yes.

> If so, would a better comment be "while using the preprovisioning URLs to PXE boot, the workers should be on the Provisioning network and that requirement does not exist if the workers are using the LiveISO URL to boot via virtualmedia".

Even this is not strictly true, because if there is no LiveISO then the PXE files will be used for virtualmedia as well (with ironic building the ISO).

imho this is not a particularly useful context in which to remind people that PXE booting happens over the Provisioning network.

provisioning/image_cache.go
kirankt: +1. I tried to build a helper but it was not so useful. Removing it. Thanks.

provisioning/image_cache.go
kirankt: +1

provisioning/baremetal_pod.go
```diff
@@ -257,7 +268,27 @@ func newMetal3InitContainers(info *ProvisioningInfo) []corev1.Container {
...
+		initContainers = append(initContainers, createInitContainerMachineOsDownloader(info, strings.Join(liveURLs, ","), true, true))
+
+		// If the ISO URL is also specified, start the createInitContainerConfigureCoreOSIPA init container
+		if info.ProvConfig.Spec.PreProvisioningOSDownloadURLs.IsoURL != "" {
+			// Configure the LiveISO by embedding ignition and other startup files
+			initContainers = append(initContainers, createInitContainerConfigureCoreOSIPA(info))
```
sadasu, line 280: Add :
// If the CoreOS IPA assets are not available we will use the IPA downloader
if !isCoreOSIPAAvailable(&info.ProvConfig.Spec) {
		initContainers = append(initContainers, createInitContainerIpaDownloader(info.Images))
}
here?

zaneb: It's already there at line 288, we don't need two.

sadasu:
Thanks @kirankt !

api/v1alpha1/provisioning_types.go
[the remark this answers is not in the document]
sadasu: > We're not planning to merge that separately though, right?
> 
No, I closed that PR.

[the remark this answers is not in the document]
sadasu: > imho this is not a particularly useful context in which to remind people that PXE booting happens over the Provisioning network.

Ageed. We'll have to be careful with our documentation given the user is presented with so many options with this new feature.

provisioning/baremetal_pod.go
zaneb: https://bugzilla.redhat.com/show_bug.cgi?id=1986098

PR #175, opened 2021-07-20
WIP: Add PreprovisioningOSDownloadURLs to the Provisioning CRD

sadasu:
These new fields can be used to supply CBO with the image URLs
needed for RHCOS-based ramdisk IPA container


api/v1alpha1/provisioning_types.go
```diff
@@ -48,6 +48,22 @@ const (
...
 )
 
+// PreprovisioningOSDownloadURLs allows specification of a set of
+// OS URLs that can be used by metal3 to pre-provision workers
+type PreprovisioningOSDownloadURLs struct {
+	// Image URL to be used for live ISO deployments
+	LiveIsoUrl string `json:"liveIsoUrl,omitempty"`
+
+	// Kernel Image URL to be used for PXE deployments
+	KernelImageUrl string `json:"kernelImageUrl,omitempty"`
```
kirankt, line 58: I would remove the Image from this name. I think having something simple like IsoUrl, KernelUrl, RootFsUrl, InitramFsUrl would be sufficient.

sadasu: Since the type is called PreprovisioningOSDownloadURLs, both "Image" and "Urls" might be unnecessary. Just calling them liveIso, kernel, initramfs, rootfs. I am open to any and all suggestions because names aren't easy.

api/v1alpha1/provisioning_types.go
```diff
@@ -48,6 +48,22 @@ const (
...
+// OS URLs that can be used by metal3 to pre-provision workers
+type PreprovisioningOSDownloadURLs struct {
+	// Image URL to be used for live ISO deployments
+	LiveISO string `json:"liveISO,omitempty"`
+
+	// Kernel Image URL to be used for PXE deployments
+	Kernel string `json:"kernel,omitempty"`
+
+	// InitRD Image URL to be used for PXE deployments
+	InitRD string `json:"initRD,omitempty"`
```
kirankt, line 61: This should be called InitramFS to match the exact file name. Initrd is an alternate name but here we're explicit about names and the other three keys here match the file names except LiveISO, which should just be ISO, IMHO.

sadasu:
/retest
I was able to apply this manifest on my local cluster.

api/v1alpha1/provisioning_types.go
sadasu: Updated based on the above comment.

sadasu:
These CRD changes are now part of https://github.com/openshift/cluster-baremetal-operator/pull/174

PR #176, opened 2021-07-21
Set external IP env var for ironic conductor too.

imain:
We need to set this for ironic conductor also as it is the thing
responsible for setting up the callback.

### DIFF
--- a/provisioning/baremetal_pod.go
+++ b/provisioning/baremetal_pod.go
@@ -607,6 +607,7 @@ func createContainerMetal3IronicConductor(images *Images, config *metal3iov1alph
 			buildEnvVar(provisioningInterface, config),
 			buildSSHKeyEnvVar(sshKey),
 			setIronicHtpasswdHash(htpasswdEnvVar, ironicrpcSecretName),
+			setIronicExternalIp(externalIpEnvVar, config),
 		},
 		Ports: []corev1.ContainerPort{
 			{


asalkeld:
any chance of some unit tests while you are at it? We are short on them..

provisioning/baremetal_pod.go
```diff
@@ -607,6 +607,7 @@ func createContainerMetal3IronicConductor(images *Images, config *metal3iov1alph
 			buildEnvVar(provisioningInterface, config),
 			buildSSHKeyEnvVar(sshKey),
 			setIronicHtpasswdHash(htpasswdEnvVar, ironicrpcSecretName),
+			setIronicExternalIp(externalIpEnvVar, config),
```
zaneb, line 610: I'm not clear why this isn't being done through buildEnvVar, but it matches the one for ironic-api.

zaneb (commented):
/lgtm

provisioning/baremetal_pod.go
hardys: It looks like because there is some additional logic around Disabled provisioning network, and AFAICS the conditional in buildEnvVar only handles `PROVISIONING_IP` not `IRONIC_EXTERNAL_IP`

Maybe that can be refactored in future, but as you say this looks Ok for now?

PR #180, opened 2021-07-23
Customize metal3 health endpoint to avoid port conflicts

andfasano:
This PR fixes the port conflicts recently discovered on CI between metal3 cluster-cloud-controller-manager-operator: both are using the default port for the health endpoint `9440`, and both are using the host network. Since cccmo is usually deployed first by CVO, when both deployments land on the same node then metal3 won't start.

### DIFF
--- a/provisioning/baremetal_pod.go
+++ b/provisioning/baremetal_pod.go
@@ -411,6 +411,7 @@ func createContainerMetal3BaremetalOperator(images *Images, config *metal3iov1al
 			},
 		},
 		Command:         []string{"/baremetal-operator"},
+		Args:            []string{"--health-addr", ":9446"},
 		ImagePullPolicy: "IfNotPresent",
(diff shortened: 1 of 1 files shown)


kirankt (approved):
/lgtm

kirankt:
I have verified this locally and changing the healtz port avoids potential conflicts.

provisioning/baremetal_pod.go
```diff
@@ -411,6 +411,7 @@ func createContainerMetal3BaremetalOperator(images *Images, config *metal3iov1al
 			},
```
zaneb, line 411: Should we add a named port for this?

zaneb (commented):
One day we should probably actually implement a liveness probe.

/approve

provisioning/baremetal_pod.go
andfasano: I guess that probably is not required

PR #181, opened 2021-07-26
Generate a cert for localhost when using VirtualMediaViaExternalNetwork

squidboylan:
When we provision using VirtualMediaViaExternalNetwork the cert
generated is for the provisioning network instead of the external IP, in
this case we decided it is fine to generate the cert for localhost
instead as this is what we do when the provisioning network is disabled.

Co-authored-by: Ian Main <imain@redhat.com>

### DIFF
--- a/provisioning/baremetal_secrets.go
+++ b/provisioning/baremetal_secrets.go
@@ -14,6 +14,8 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/client"
 	"sigs.k8s.io/controller-runtime/pkg/controller/controllerutil"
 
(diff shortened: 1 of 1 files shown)


provisioning/baremetal_secrets.go
```diff
@@ -14,6 +14,8 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/client"
 	"sigs.k8s.io/controller-runtime/pkg/controller/controllerutil"
 
+	metal3iov1alpha1 "github.com/openshift/cluster-baremetal-operator/api/v1alpha1"
+
```
asalkeld, line 18: kill the newline here

PR #183, opened 2021-07-26
Bug 1986098: Use cached client for all normal access

asalkeld:
The main focus is to use the controller-runtime client as much as possible inorder to benefit from it's caching.
How ever there are a number of places that we access resources that are no cached.
1. non-namespaced resources like Proxy and Infrastructure
2. the webhook needs to be enabled before the client cache is sync'd
3. there is a secret and a configMap that we read from other namespaces

We *could* use the multiNamespaced cache, but this would blow out the memory used by caching the new namespaces when we
only need one resource from that namespace.
So this PR proposes a hybrid approach:
1. keep with caching only the openshift-machine-api namespace 
2. use a new module to isolate the other client usages to make it clearer when to do so


controllers/clusteroperator.go
```diff
@@ -125,14 +126,19 @@ func (r *ProvisioningReconciler) createClusterOperator() (*osconfigv1.ClusterOpe
 		return nil, fmt.Errorf("could not convert deserialized asset into ClusterOperoator")
 	}
 
-	return r.OSClient.ConfigV1().ClusterOperators().Create(context.Background(), defaultCO, metav1.CreateOptions{})
+	err = r.Client.Create(ctx, defaultCO, &client.CreateOptions{})
+	if err != nil {
+		return nil, err
+	}
```
asalkeld, line 132: maybe need a fresh Get here?

provisioning/baremetal_pod.go
```diff
@@ -1018,21 +1018,21 @@ func EnsureMetal3Deployment(info *ProvisioningInfo) (updated bool, err error) {
 	}
 
 	deploymentRolloutStartTime = time.Now()
-	deployment, updated, err := resourceapply.ApplyDeployment(info.Client.AppsV1(),
+	deployment, updated, err := resourceapply.ApplyDeployment(info.KubeClient.AppsV1(),
```
asalkeld, line 1021: So this resourceapply stuff only takes the native clients not the controller-runtime client. We need to either move to library-go client or change to  a different "apply" library..
https://github.com/kubernetes-sigs/cluster-api/blob/master/util/patch/patch.go#L83

provisioning/baremetal_pod.go
```diff
@@ -996,49 +995,76 @@ func newMetal3Deployment(info *ProvisioningInfo) *appsv1.Deployment {
...
+
+	err := c.Get(context.Background(), client.ObjectKeyFromObject(existing), existing)
 	if existing != nil && err == nil {
 		return existing.Spec.Selector, nil
 	}
 	return nil, err
 }
 
-func EnsureMetal3Deployment(info *ProvisioningInfo) (updated bool, err error) {
+func EnsureMetal3Deployment(info *ProvisioningInfo) (bool, error) {
```
asalkeld, line 1010: the fun starts here :-)

provisioning/state_service.go
```diff
@@ -40,23 +40,37 @@ func newMetal3StateService(targetNamespace string) *corev1.Service {
...
+
+	return updated, err
 }
 
-func DeleteMetal3StateService(info *ProvisioningInfo) error {
-	return client.IgnoreNotFound(info.Client.CoreV1().Services(info.Namespace).Delete(context.Background(), stateService, metav1.DeleteOptions{}))
+func DeleteMetal3StateService(ctx context.Context, info *ProvisioningInfo) error {
+	obj := &corev1.Service{ObjectMeta: metav1.ObjectMeta{Name: stateService, Namespace: info.Namespace}}
+	err := client.IgnoreNotFound(info.Client.Delete(ctx, obj, &client.DeleteOptions{}))
+	reportDeleteEvent(info.EventRecorder, obj, err)
```
asalkeld, line 75: need "return err"

provisioning/webhook.go
```diff
@@ -77,7 +77,7 @@ func EnableValidatingWebhook(info *ProvisioningInfo, mgr manager.Manager) error
 	// we might not have a baremetalCR (when disabled), so we have no where to store
 	// the expectedGeneration, so just fake it.
 	expectedGeneration := int64(0)
-	_, _, err := resourceapply.ApplyValidatingWebhookConfiguration(info.Client.AdmissionregistrationV1(), info.EventRecorder, instance, expectedGeneration)
+	_, _, err := resourceapply.ApplyValidatingWebhookConfiguration(info.KubeClient.AdmissionregistrationV1(), info.EventRecorder, instance, expectedGeneration)
```
ardaguclu, line 80: ProvisioningInfo is created in here https://github.com/openshift/cluster-baremetal-operator/blob/5f4ae6841cd64cf74e9c8cd31523dc84ff7bda6f/main.go#L110 without setting KubeClient. But it accesses it in here.

provisioning/provisioning_info.go
```diff
@@ -18,7 +19,8 @@ const (
 )
 
 type ProvisioningInfo struct {
-	Client             kubernetes.Interface
+	KubeClient         kubernetes.Interface
```
ardaguclu, line 22: AFAICS, KubeClient is used 4 times and Client is used frequently. Isn't it possible to use one of them(Possibly `client.Client`) to uniform client usage?. Otherwise, I saw a couple of upstream repositories that there are lots of different client objects and contributors seem like using them randomly(or I don't understand the procedure which one is used in which cases :) ).

asalkeld: i could try and separate them for clarity. but the reason I am keeping them separate is the r.Client is a cached client for the namespace we watching. but if we access things explicitly in other namespaces i believe we still need to access it using the kubeclient. but i'll test to make sure (i would have thought it would work fine, it would just have read from the API).
For the webhook i have to use kubeclient as we need to use it before the cached client is initialised.

zaneb (commented):
Ugh, when I spotted this issue I didn't realise it was another case of controller-runtime/library-go incompatibility :disappointed: 

The first 3 patches look like an easy win. I'm now less sure that saving excess reads is worth having to effectively reimplement large portions of library-go.

controllers/provisioning_controller.go
```diff
@@ -333,15 +330,15 @@ func (r *ProvisioningReconciler) Reconcile(ctx context.Context, req ctrl.Request
...
 	proxy := &osconfigv1.Proxy{}
-	err := r.Client.Get(context.Background(), client.ObjectKey{Name: "cluster"}, proxy)
+	err := r.Client.Get(ctx, client.ObjectKey{Name: "cluster"}, proxy)
 	if err != nil {
 		return nil, err
 	}
 
 	return &provisioning.ProvisioningInfo{
-		Client:             r.KubeClient,
+		Client:             r.Client,
```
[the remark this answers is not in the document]
asalkeld: I am now using multiNamespace watcher to include openshift-config and kube-system...

asalkeld:
```
Aug  2 23:02:46.116: INFO: operator=cluster-baremetal-operator, watchrequestcount=140, upperbound=50, ratio=2.8
Aug  2 23:02:46.116: INFO: Operator cluster-baremetal-operator produces more watch requests than expected
```
We are watching 2 new namespaces, so this is obvious, we might just have to tweek the  e2e test

sadasu:
Proxy and Platform configurations donot change once the cluster is up. So, could we read them once in main() and store them in ProvisioningInfo instead of reading during each reconcile loop? That is a small win.

I am not for re-implementing the apply*() methods.

asalkeld:
> Proxy and Platform configurations donot change once the cluster is up. So, could we read them once in main() and store them in ProvisioningInfo instead of reading during each reconcile loop? That is a small win.
> 
> I am not for re-implementing the apply*() methods that are already implemented in library-go.
Then there is no point to this PR all the main requests are going through the apply functions and they can't be used with the cached client.

asalkeld:
failed tests
metal-ipi
```
[sig-cli] Kubectl client Simple pod should return command exit codes [Suite:openshift/conformance/parallel]
[sig-arch] Managed cluster should ensure platform components have system-* priority class associated [Suite:openshift/conformance/parallel]
```
e2e-metal-ipi-ovn-ipv6
``` 
[sig-cli] Kubectl client Simple pod should return command exit codes [Suite:openshift/conformance/parallel] [
[sig-cli] oc explain should contain proper fields description for special types [Suite:openshift/conformance/parallel]
[Conformance][sig-api-machinery][Feature:APIServer] local kubeconfig "localhost.kubeconfig" should be present on all masters and work [Suite:openshift/conformance/parallel/minimal]
```

asalkeld:
ci/prow/e2e-metal-ipi-ovn-ipv6
```
[sig-cli] Kubectl client Simple pod should return command exit codes [Suite:openshift/conformance/parallel] [Suite:k8s]
```
/retest

PR #185, opened 2021-07-27
Bug 1986462: Separate the names of machine os downloader when two copies are started 

kirankt:
When specifying both PreProvisioningOSDownloadURLs and ProvisioningOSDownloadURL in the provisioning config CR, CBO will fail to reconcile becasuse two copies of machine-os-downloader init containers with the same name are started. The names of these initcontainers should be different when both of them are spawned together.

### DIFF
--- a/provisioning/baremetal_config_test.go
+++ b/provisioning/baremetal_config_test.go
@@ -176,6 +176,25 @@ func disabledProvisioning() *provisioningBuilder {
 	}
 }
 
+func configWithPreProvisioningOSDownloadURLs() *provisioningBuilder {
+	return &provisioningBuilder{
+		metal3iov1alpha1.ProvisioningSpec{
+			ProvisioningInterface:     "eth0",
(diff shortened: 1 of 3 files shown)


asalkeld (commented):
can you create a new unit test section like this https://github.com/openshift/cluster-baremetal-operator/blob/master/provisioning/baremetal_pod_test.go#L330-L346
but with the PreProvisioningOSDownloadURLs defined

kirankt:
> can you create a new unit test section like this https://github.com/openshift/cluster-baremetal-operator/blob/master/provisioning/baremetal_pod_test.go#L330-L346
> but with the PreProvisioningOSDownloadURLs defined

This PR only deals with the initContainers. We don't really mess with the existing containers in the pod. But I will add a section in the initContainers. Realistically there should be another PR that should validate the preProvisioningOSDownloadURLs and perform unit testing on them.

sadasu (commented):
/lgtm

PR #189, opened 2021-08-03
cbo should be enabled on baremetal and "none" platforms

asalkeld:
/cc @sadasu 
/hold


romfreiman:
few comments:
1. Lets try to understand whether OpenStack is a requirement
2. if  CBO is enabled, should it get the manifests for workload partitioning (see here: https://github.com/openshift/enhancements/blob/master/enhancements/workload-partitioning/management-workload-partitioning.md#example-manifests) cc @dhellmann
3. Will it affect SNO performance? cc: @browsell

sadasu:
Do these changes also include disabling metal3 when the Topology is "External"? https://github.com/openshift/enhancements/pull/871/files#diff-7fe1b077c80e38c1b2725c8fcd5cf98a3049943033fa28a25e51cdd4b22a427cR93

asalkeld:
> Do these changes also include disabling metal3 when the Topology is "External"? https://github.com/openshift/enhancements/pull/871/files#diff-7fe1b077c80e38c1b2725c8fcd5cf98a3049943033fa28a25e51cdd4b22a427cR93

@sadasu no, this is just the initial PR I did, I'll wait until the enhancement is merged then update.

controllers/provisioning_controller.go
```diff
@@ -245,19 +241,15 @@ func (r *ProvisioningReconciler) Reconcile(ctx context.Context, req ctrl.Request
...
-			// Provisioning configuration is not valid.
-			// Requeue request.
-			klog.Error(err, "invalid config in Provisioning CR")
-			err = r.updateCOStatus(ReasonInvalidConfiguration, err.Error(), "Unable to apply Provisioning CR: invalid configuration")
-			if err != nil {
-				return ctrl.Result{}, fmt.Errorf("unable to put %q ClusterOperator in Degraded state: %v", clusterOperatorName, err)
-			}
-			// Temporarily not requeuing request
-			return ctrl.Result{}, nil
+	// Always re-validate the provisioning configuration is valid.
```
sadasu, line 230: Good point.

sadasu:
IsEnabled() could return false if the Platform or the Topology are unsupported. So, could we modify https://github.com/openshift/cluster-baremetal-operator/pull/189/commits/da8b35c8ea239541b69b66d2e0706d2d3bf71df2#diff-7ad72d08a06fe28b03a15c84289b815f693443562d0c67e6f461c7e3f5ee0707R467 to reflect that?

controllers/provisioning_enabled.go
```diff
@@ -0,0 +1,60 @@
...
+
+	switch infra.Status.Platform {
+	case osconfigv1.BareMetalPlatformType, osconfigv1.OpenStackPlatformType, osconfigv1.NonePlatformType, osconfigv1.VSpherePlatformType:
+		return true, nil
+	default:
+		return false, nil
+	}
+}
+
+func EnabledFeatures(ctx context.Context, osClient osclientset.Interface) (v1alpha1.EnabledFeatures, error) {
```
sadasu, line 32: Both these methods are reading the same resources. Even though they are read from the cache, we could make it more efficient by setting EnabledFeatures while determining if this is a supported platform.

main.go
```diff
@@ -92,16 +93,23 @@ func main() {
...
-		ImagesFilename: imagesJSONFilename,
-		WebHookEnabled: enableWebhook,
+		Client:          mgr.GetClient(),
+		Scheme:          mgr.GetScheme(),
+		OSClient:        osClient,
+		KubeClient:      kubeClient,
+		ReleaseVersion:  releaseVersion,
+		ImagesFilename:  imagesJSONFilename,
+		WebHookEnabled:  enableWebhook,
+		EnabledFeatures: enabledFeatures,
```
sadasu, line 112: We can determine enabledFeatures while calling IsEnabled() in SetupWithManager().

PR #194, opened 2021-09-02
Bug 2000445: Set LIVE_ISO_FORCE_PERSISTENT_BOOT_DEVICE=Never

derekhiggins:
Enable the functionality to set force_persistent_boot_device
to "Never" for live ISO nodes. This will allow nodes to be
booted once to CD then rebooted to HD.


derekhiggins:
Hi @mfojtik, wondering if you can also approve this as its needed in conjunction with https://github.com/openshift/baremetal-operator/pull/174

PR #195, opened 2021-09-07
Bug 2000081: add a new field "ProvisioningMacAddresses" to the provisioning CRD

asalkeld:
This makes an explicit field for what we are trying to achieve so that the user can modify as to their needs.
https://coreos.slack.com/archives/C01RJHA6BRC/p1631185821227600?thread_ts=1631179467.225300&cid=C01RJHA6BRC


manifests/0000_31_cluster-baremetal-operator_06_deployment.yaml
```diff
@@ -76,6 +80,7 @@ spec:
           name: config
         - mountPath: /etc/tls/private
           name: cluster-baremetal-operator-tls
+      hostNetwork: true
```
zaneb, line 83: If we put the CBO on the host network then we need to move a bunch of ports and register them in the port registry.

controllers/provisioning_controller.go
```diff
@@ -333,8 +325,13 @@ func (r *ProvisioningReconciler) Reconcile(ctx context.Context, req ctrl.Request
...
 
-func (r *ProvisioningReconciler) provisioningInfo(provConfig *metal3iov1alpha1.Provisioning, images *provisioning.Images, sshkey string) (*provisioning.ProvisioningInfo, error) {
-	proxy, err := r.OSClient.ConfigV1().Proxies().Get(context.Background(), "cluster", metav1.GetOptions{})
+func (r *ProvisioningReconciler) provisioningInfo(ctx context.Context, provConfig *metal3iov1alpha1.Provisioning, images *provisioning.Images, sshkey string) (*provisioning.ProvisioningInfo, error) {
+	proxy, err := r.OSClient.ConfigV1().Proxies().Get(ctx, "cluster", metav1.GetOptions{})
+	if err != nil {
+		return nil, err
+	}
+
+	macAddresses, err := r.masterMacAddresses(ctx)
```
zaneb, line 333: Q: why are we even relying on the BootMACAddress given that many of the metal3 drivers (e.g. IPMI) don't even require it to be present?

asalkeld: that was the basis for this epic: https://issues.redhat.com/browse/MPINSTALL-6?focusedCommentId=16032926&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-16032926

controllers/provisioning_controller.go
```diff
@@ -445,18 +440,23 @@ func (r *ProvisioningReconciler) apiServerInternalHost(ctx context.Context) (str
...
+		return err
 	}
 	for _, bmh := range bmhl.Items {
 		if strings.Contains(bmh.Name, "master") && len(bmh.Spec.BootMACAddress) > 0 {
 			macs = append(macs, bmh.Spec.BootMACAddress)
 		}
 	}
-	return macs, nil
+	provConfig.Spec.ProvisioningMacAddresses = macs
+	return r.Client.Update(ctx, provConfig)
```
asalkeld, line 459: we can later move this to the installer

hardys: Yeah I was wondering if we should just calculate this in the installer, and CBO can then just consume the provisioning CR value like other install-time-provided things.

That also partially resolves the issue that bootMACAddress of the BMH is optional, but in the install-config it's mandatory

controllers/provisioning_controller.go
hardys: @zaneb FYI although provisioning can work for IPMI without bootMACAddress - adopt will then fail without it ref https://bugzilla.redhat.com/show_bug.cgi?id=1854142

@derekhiggins looked into changing that Ironic behavior ref https://bugzilla.redhat.com/show_bug.cgi?id=1876573 but AFAIK we've not found a way to remove that constraint, so the installer currently does treat bootMACAddress as mandatory in the install-config schema (hence, it seemed reasonable to leverage this existing data to identify the provisioning NIC)

hardys: Perhaps it would be reasonable to only consume the install-config data (which is mandatory) to set this provisioning CR value, avoiding any confusion with the BMH field which is theoretically-optional-but-actually-required?

zaneb: Do we need to fix https://github.com/metal3-io/baremetal-operator/blob/master/pkg/bmc/ipmi.go#L60 then?
Do other examples (like https://github.com/metal3-io/baremetal-operator/blob/master/pkg/bmc/irmc.go#L36) have the same problem?

hardys: @derekhiggins can you confirm - IIRC the adopt behavior always requires the MAC regardless of the driver (but on deploy some drivers work without a MAC because they can correlate a unique BMC address), but not completely sure I'm recalling the details correctly?

derekhiggins: @hardys ya there are some inconsistencies between drivers and when exactly the MAC is required, iirc pxe requires it on adopt but redfish-vmedia doesn't (I didn't look at any others) the safest option is to always specify it. I'm fuzzy on the exact details of when each driver requires it, I'd have to revisit this bug to see if it can be made more consistent.

zaneb: I opened https://github.com/metal3-io/baremetal-operator/issues/979 to track this.

controllers/provisioning_controller.go
sadasu: Will it always be a one-time config to happen at install time? What if one of the master nodes gets replaced, should this list of bootMacAddresses have get recalculated?

PR #200, opened 2021-09-15
Bug 2004597: Remove duplicate ramdisk log container

derekhiggins:
The runlogwatch.sh script is now intended to watch
both directories, so we only need one instance of
this container.

Also update the ramdisk naming to match the other containers.

Side Note:  gathering the inspection logs is also currently broken due to a bug,
this is fixed upstream https://github.com/metal3-io/ironic-image/pull/289


asalkeld (commented):
/lgtm

sadasu:
/approve
How did this happen? :-)

PR #202, opened 2021-09-22
Set IRONIC_AGENT_IMAGE in the ipa init container

asalkeld:


zaneb:
LGTM but not to CI apparently :thinking:

kirankt (approved):
/lgtm

PR #99, opened 2021-02-03
Add standard metrics support for metal3

sadasu:
For Prometheus to be able to scrape metal3 metrics:

Added kubeProxyContainer to the metal3 deployment
Created a NodePort Service (metal3-metrics-service) to expose CMO metrics port
Created a Service Monitor (metal3-metrics-servicemonitor) in order to be discovered automatically by the prometheus operator
Updated rbacs in the openshift-machine-api namespace for the service monitor

PR #203, opened 2021-09-23
WIP: Enable webhook for bmo

ardaguclu:


provisioning/baremetal_webhook.go
```diff
@@ -0,0 +1,165 @@
...
+
+	"sigs.k8s.io/controller-runtime/pkg/client"
+
+	admissionregistration "k8s.io/api/admissionregistration/v1"
+	"k8s.io/utils/pointer"
+
+	"github.com/openshift/library-go/pkg/operator/resource/resourceapply"
+
+	"k8s.io/apimachinery/pkg/util/intstr"
+
```
asalkeld, line 21: can you cleanup the imports into groups please?

ardaguclu: I relied on goimports but like you said it definitely needs some manual grouping. I fixed it. Thank you @asalkeld.

provisioning/baremetal_webhook.go
```diff
@@ -0,0 +1,159 @@
...
+			Name:      validatingWebhookService,
+			Namespace: targetNamespace,
+			Annotations: map[string]string{
+				"include.release.openshift.io/self-managed-high-availability": "true",
+				"include.release.openshift.io/single-node-developer":          "true",
+				"service.beta.openshift.io/serving-cert-secret-name":          baremetalWebhookSecretName,
+			},
+			Labels: map[string]string{
+				"k8s-app":    metal3AppName,
+				cboLabelName: stateService,
```
zaneb, line 95: I don't want to use this label as the target. The purpose of the metal3-state service is to point to the Pod that holds the DB and httpd. It exists specifically so that things that are stateless, like baremetal-operator, can eventually be _removed_ from that Pod.
If we want a service to point at the baremetal-operator, we should create another label for it so that they can move independently in future.

provisioning/baremetal_webhook.go
```diff
@@ -0,0 +1,159 @@
...
+				},
+			},
+		},
+	}
+}
+
+func newBaremetalOperatorWebhook(targetNamespace string) *admissionregistration.ValidatingWebhookConfiguration {
+	failurePolicy := admissionregistration.Fail
+	sideEffect := admissionregistration.SideEffectClassNone
+	instance := &admissionregistration.ValidatingWebhookConfiguration{
```
zaneb, line 118: While this looks fine, it also seems like this is something that could be applied directly as a manifest based on the upstream one from the BMO repo with some light kustomization, just as we do for the CRD. That advantage of doing that instead of doing it in the operator is (a) less code, but also (b) they can't get out of sync. WDYT?

provisioning/baremetal_webhook.go
```diff
@@ -0,0 +1,159 @@
...
+	"k8s.io/utils/pointer"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+
+	osclientset "github.com/openshift/client-go/config/clientset/versioned"
+	"github.com/openshift/library-go/pkg/operator/resource/resourceapply"
+	"github.com/openshift/library-go/pkg/operator/resource/resourcemerge"
+)
+
+const (
+	validatingWebhookService           = "baremetal-operator-webhook-service"
```
zaneb, line 21: One day we'll finally resurrect the metrics PR (#99) and there'll be another thing that needs to point to the baremetal-operator. I wonder if it's worth naming this in such a way that it is not webhook-specific.

provisioning/baremetal_webhook.go
```diff
@@ -0,0 +1,159 @@
...
+		},
+		Spec: corev1.ServiceSpec{
+			Type: corev1.ServiceTypeClusterIP,
+			Selector: map[string]string{
+				"k8s-app":    metal3AppName,
+				cboLabelName: stateService,
+			},
+			Ports: []corev1.ServicePort{
+				{
+					Name:       httpPortName,
```
zaneb, line 106: Something like "validating-webhook" might be better than "http"?

provisioning/baremetal_webhook.go
ardaguclu: Thank you @zaneb, that is a good point and I'll create another label.

provisioning/baremetal_webhook.go
ardaguclu: In upstream, we can safely apply webhook manifests because cluster-api ensures that cert-manager is running and healthy. In downstream, we rely on service-ca but there is no guarantee that it is healthy. Probably, that's why, CBO also uses code instead manifests as in here https://github.com/openshift/cluster-baremetal-operator/blob/af0136a0dba7f42f2a493ba7a09dc33f81d45c41/provisioning/webhook.go#L38. 

And I agree about less code and always up to date, but ValidatingWebhookConfiguration has very few fields that can be changeable once it is set.

provisioning/baremetal_webhook.go
ardaguclu: If I understand correct, you mean once metrics api is ready, we can use this service with additional port bindings. If we rename this to general purpose, maybe we should move this service from `baremetal_webhook.go` to `baremetal_service.go`.

IMO, Instead re-using this service for different port bindings, I'd rather propose we can create new service resource (with same labels) in `baremetal_metrics.go` or even better metrics service yaml under manifests.

zaneb: Yeah, the current PR creates a separate Service for metrics.
I'm just wondering how many Services we need pointing to the same Container in the same Pod.

provisioning/baremetal_webhook.go
```diff
@@ -0,0 +1,160 @@
...
+	validatingWebhookHttpPortName      = "http"
+)
+
+// EnsureBaremetalOperatorWebhook ensures ValidatingWebhook resources are ready to serve.
+func EnsureBaremetalOperatorWebhook(info *ProvisioningInfo) (bool, error) {
+	if !info.BaremetalWebhookEnabled {
+		// In case service-ca operator is degraded and we detect it in reconciliation loop
+		// We should delete validatingwebhook resources(ValidatingWebhookConfiguration, Service)
+		// to allow BMO continues serving. If we detect in first run, it is safe to run these
+		// deletions.
```
asalkeld, line 33: I seriously doubt that any other operator does this. I'd suggest just leave it in place.

provisioning/baremetal_webhook.go
```diff
@@ -0,0 +1,160 @@
...
+
+// DeleteValidatingWebhook deletes ValidatingWebhookConfiguration and
+// service resources.
+func DeleteValidatingWebhook(info *ProvisioningInfo) error {
+	err := client.IgnoreNotFound(info.Client.CoreV1().Services(info.Namespace).Delete(context.Background(), validatingWebhookService, metav1.DeleteOptions{}))
+	if err != nil {
+		return err
+	}
+
+	err = client.IgnoreNotFound(info.Client.AdmissionregistrationV1().
```
asalkeld, line 76: return client.IgnoreNotFound(...)

provisioning/baremetal_webhook.go
```diff
@@ -0,0 +1,160 @@
...
+					Port:       validatingWebhookServiceHttpsPort,
+					TargetPort: intstr.FromInt(int(webhookPort)),
+				},
+			},
+		},
+	}
+}
+
+func newBaremetalOperatorWebhook(targetNamespace string) *admissionregistration.ValidatingWebhookConfiguration {
+	failurePolicy := admissionregistration.Fail
```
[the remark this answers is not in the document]
asalkeld: @ardaguclu not really.
we use service-ca just to inject the cert, once that is done it should be fine.
any errors then are network errors or us being not available.
We do that check so we don't register for the webhook before service-ca can inject our cert.

PR #205, opened 2021-09-30
Add tests for missing ProvisioningIP in validation

zaneb:
Explicitly verify that validation fails if the ProvisioningIP is
missing, unless the Provisioning network is disabled.


asalkeld (commented):
/lgtm

PR #206, opened 2021-10-11
Serve virtual media ISOs over TLS

dtantsur:
Especially in the live ISO case, the virtual media images may contain
sensitive information. Now that we've established that Redfish hardware
does not verify TLS certificates, we can use our self-signed certificate
to protect the communication with the BMC.

A knob to turn this feature off is provided for hardware that diverges.

Requires:
- https://github.com/openshift/ironic-image/pull/218
- https://github.com/openshift/enhancements/pull/917
- https://github.com/openshift-metal3/dev-scripts/pull/1301

### DIFF
--- a/provisioning/baremetal_pod.go
+++ b/provisioning/baremetal_pod.go
@@ -49,6 +49,7 @@ const (
 	inspectorCredentialsVolume       = "metal3-inspector-basic-auth"
 	ironicTlsVolume                  = "metal3-ironic-tls"
 	inspectorTlsVolume               = "metal3-inspector-tls"
+	vmediaTlsVolume                  = "metal3-vmedia-tls"
 	htpasswdEnvVar                   = "HTTP_BASIC_HTPASSWD" // #nosec
 	mariadbPwdEnvVar                 = "MARIADB_PASSWORD"    // #nosec
 	ironicInsecureEnvVar             = "IRONIC_INSECURE"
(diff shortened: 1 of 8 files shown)


asalkeld:
looks good to me, but maybe wait for the enhancement to merge?

provisioning/baremetal_pod.go
```diff
@@ -612,34 +627,47 @@ func createContainerMetal3Mariadb(images *Images) corev1.Container {
...
-		},
+		Command:      []string{"/bin/runhttpd"},
+		VolumeMounts: volumes,
 		Env: []corev1.EnvVar{
 			buildEnvVar(httpPort, config),
 			buildEnvVar(provisioningIP, config),
 			buildEnvVar(provisioningInterface, config),
 			buildSSHKeyEnvVar(sshKey),
 			buildEnvVar(provisioningMacAddresses, config),
+			buildEnvVar(vmediaHttpsPort, config),
```
sadasu, line 658: Is it OK to set this even when DisableVirtualMediaTLS is set to True?

provisioning/baremetal_pod.go
```diff
@@ -612,34 +627,47 @@ func createContainerMetal3Mariadb(images *Images) corev1.Container {
...
 		},
 		Ports: []corev1.ContainerPort{
 			{
 				Name:          httpPortName,
 				ContainerPort: int32(port),
 				HostPort:      int32(port),
 			},
+			{
```
sadasu, line 666: Is it OK to set this even when DisableVirtualMediaTLS is set to True?

provisioning/baremetal_pod.go
dtantsur: It won't trigger anything in ironic-image (there is a default value for the port anyway). I tried to reduce complexity by having fewer conditionals, but I can change this.

provisioning/baremetal_pod.go
dtantsur: I think so? I'm not sure what a ContainerPort does if nothing is listening.

asalkeld (commented):
/lgtm

asalkeld (commented):
/lgtm

PR #217, opened 2021-11-19
Bug 2024900: Not enable CBO webhook in unsupported platform

ardaguclu:
This PR adds additional check to not enable `ValidatingWebhookConfiguration` for non baremetal platforms like AWS, Azure, GCP. Because Cluster Baremetal Operator is already disabled in these platforms.


sttts:
Sgtm

PR #219, opened 2021-12-02
Make ProvisioningOSDownloadURL optional

zaneb:
In future we will provision with the CoreOS installer, so we won't need
a ProvisioningOSDownloadURL. Stop enforcing its presence in the webhook.


sadasu (commented):
/lgtm

PR #220, opened 2021-12-11
Get images from machine-os-images

zaneb:


zaneb:
These changes were incorporated back into #208 
/close

PR #221, opened 2021-12-11
Add the image-customization controller to the images configmap

zaneb:
The image-customization-controller will not be added to the release payload until it is referenced in the images configmap of CBO. (Or alternatively has the label `io.openshift.release.operator=true` applied.)

This patch is part of #208, but we want the installer change https://github.com/openshift/installer/pull/5425 to land first. It cannot do so unless the image-customization-controller is in the release payload.

We have successfully built an image in OSBS:
```
ose-image-customization-controller-container-v4.10.0-202112102123.p0.gfaf44b3.assembly.stream successfully tagged into rhaos-4.10-rhel-8-candidate by osbs
```


zaneb:
/hold
Let's wait until we get confirmation on what the "published back to the nightly test jobs" bit of "ensure their image is built in OSBS at least once and published back to the nightly test jobs" means.

zaneb:
I see multiple successful builds of this container downstream, and multiple accepted nightly builds of the product since it started being built, so I believe this is now safe to merge.
/hold cancel

PR #191, opened 2021-08-16
Bug 1991641: make provisioning CR status more explicit and remove the ownership on the CO

asalkeld:
- Make the state of a missing CR more explicit by creating a ```ReasonProvisioningCRNotFound```.
- Fix a code path when this was not set.
- Remove the ownership on the clusterOperator as it was both ineffective and possibly not desired.

/cc @sadasu

PR #224, opened 2021-12-22
Bug 2011824: Fix CO message when Provisioning CR is not present

sadasu:
This is a backport of https://github.com/openshift/cluster-baremetal-operator/pull/191 and https://github.com/openshift/cluster-baremetal-operator/pull/214 that together improved the CO handling especially the status reported when the Provisioning CR is not present.


zaneb:
I'm not sure this needs to be backported. It was necessary in 4.10 because CBO now runs on other platforms where the installer doesn't install a provisioning CR. Unless there's some particular call for it in 4.9 I don't see a reason to backport.

sadasu:
Fixed in 4.10, no need to backport this.

PR #225, opened 2021-12-23
Bug 2035239: Add pod listing permission to extract pod host IP

ardaguclu:
To extract pod's host IP address, CBO tries to list pods in here https://github.com/openshift/cluster-baremetal-operator/blob/b7bba707a8d6c3a3e6b63dc75afdf93ec60183a1/provisioning/baremetal_pod.go#L1071

However, CBO does not have allowance to list pods. This PR adds this.

### DIFF
--- a/provisioning/baremetal_pod.go
+++ b/provisioning/baremetal_pod.go
@@ -1063,16 +1063,31 @@ func DeleteMetal3Deployment(info *ProvisioningInfo) error {
 }
 
 func getPodHostIP(podClient coreclientv1.PodsGetter, targetNamespace string) (string, error) {
+	labelSelector := &metav1.LabelSelector{
+		MatchLabels: map[string]string{
+			"k8s-app":    metal3AppName,
+			cboLabelName: stateService,
+		}}
+
+	selector, err := metav1.LabelSelectorAsSelector(labelSelector)
+	if err != nil {
+		return "", err
+	}
+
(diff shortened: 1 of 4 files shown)


provisioning/baremetal_pod.go
```diff
@@ -1065,7 +1065,6 @@ func DeleteMetal3Deployment(info *ProvisioningInfo) error {
 func getPodHostIP(podClient coreclientv1.PodsGetter, targetNamespace string) (string, error) {
 	listOptions := metav1.ListOptions{
 		LabelSelector: metal3AppName,
-		FieldSelector: "status.hostIP",
```
ardaguclu, line 1068: I removed this because it failed with;
```
 invalid selector: 'status.hostIP'; can't understand 'status.hostIP'" "name"="provisioning-configuration"
```

Maybe it is because of lack of `get` permission but I did this change for a quick fix

andfasano: As an additional safety measure, can you please also verify that `len(podList.Items) == 1` instead of `> 0`?

ardaguclu: What do you think for the latest version @andfasano ?

andfasano: Thanks, it seems ok, let's see the virtualmedia job

PR #227, opened 2022-01-04
Bug 2037483: Allow CBO to list Pods only in the openshift-machine-api namespace

sadasu:
CBO should be able to list the Pods running in the openshift-machine-api namespace to determine the Pod's IP to create the i-c-c deployment. Also, removing CBO's cluster wide permissions to get and list Pods.

### DIFF
--- a/config/rbac/role.yaml
+++ b/config/rbac/role.yaml
@@ -30,13 +30,6 @@ rules:
   - patch
   - watch
 - apiGroups:
-  - ""
-  resources:
-  - pods
-  verbs:
-  - get
-  - list
-- apiGroups:
   - admissionregistration.k8s.io
   resources:
   - validatingwebhookconfigurations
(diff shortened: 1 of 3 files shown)


zaneb:
I think the current permissions added in #225 allow us to list Pods in every namespace - so it will be working, but we should _definitely_ narrow it down to just openshift-machine-api.

ardaguclu:
Good point, maybe we can also remove these permissions https://github.com/openshift/cluster-baremetal-operator/blob/ca77bd52efa74cbdd06f7cf38d177acd88864f01/config/rbac/role.yaml#L32 in this PR

sadasu:
/test e2e-agnostic

Seems like a temporary error

ardaguclu:
/test e2e-metal-ipi-virtualmedia

Maybe this job is the main indicator

PR #228, opened 2022-01-05
Remove asalkeld and kirankt from Owners

kirankt:
Remove asalkeld and kirankt from Owners

### DIFF
--- a/OWNERS
+++ b/OWNERS
@@ -2,10 +2,8 @@
 
 approvers:
  - andfasano
- - asalkeld
  - hardys
  - markmc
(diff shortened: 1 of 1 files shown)


sadasu:
@kirankt we need a BZ for this.

kirankt:
> @kirankt we need a BZ for this.

Ugh. This is not urgent. We can hold off until 4.11 branch opens. Thanks.

zaneb:
I think there's a way to override the BZ requirement, but possibly only @hardys can do that.

PR #229, opened 2022-01-05
Bug 2038088: Remove references to ipa-downloader

elfosardo:


dtantsur:
/lgtm

Will need a valid bug to get merged in the near future.

provisioning/baremetal_pod_test.go
```diff
@@ -99,16 +98,16 @@ func TestNewMetal3InitContainers(t *testing.T) {
...
-					Name:  "metal3-ipa-downloader",
-					Image: images.IpaDownloader,
+					Name:  "metal3-static-ip-set",
+					Image: images.StaticIpManager,
 				},
 				{
-					Name:  "metal3-machine-os-downloader",
-					Image: images.MachineOsDownloader,
+					Name:  "machine-os-images",
+					Image: images.IronicAgent,
```
zaneb, line 106: This doesn't look right - machine-os-images does not use the IronicAgent image, it uses the MachineOSImages image.

It appears the test is only checking the number of containers... I think we should test the contents or just put the number here.

elfosardo: I actually test the names of the containers, although I didn't add the test here, so the names are correct
I probably overlooked the Image name by mistake, it was not intentional
I'm going to fix this and maybe we can add the check on the containers names in a different PR

zaneb:
Thanks for handling this!
/approve

PR #230, opened 2022-01-06
Bug 2034527: Base IPA kernel params on provisioning network IP version

derekhiggins:
We need the kernel param for dhcp version to be based on the
provisioning network version for IPA and the external network
version for the machine image. These can differ.


zaneb:
I'm  having second thoughts about whether this is sufficient: https://github.com/openshift/ironic-agent-image/pull/29#issuecomment-1006978285

PR #237, opened 2022-01-17
Bug 2041510: Enable webhook for bmceventsubscriptions

honza:
@ardaguclu Is this enough to enable it? Do I need to do anything else?

### DIFF
--- a/provisioning/baremetal_webhook.go
+++ b/provisioning/baremetal_webhook.go
@@ -146,6 +146,33 @@ func newBaremetalOperatorWebhook(targetNamespace string) *admissionregistration.
 					},
 				},
 			},
+			{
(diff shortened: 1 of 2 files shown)


omichaeli:
@honza Should the bmceventsubscription be in plural?

honza:
> @honza Should the bmceventsubscription be in plural?

Yes. :man_facepalming:

ardaguclu:
@honza this looks good to me and should be enough.

I highly suggest to add a test to check that endpoint is healthy like in here https://github.com/openshift/origin/blob/b5c202c718553f4124030ce2b21a1d7059ee37cf/test/extended/baremetal/hosts.go#L244.

/lgtm
/e2e-agnostic

bfournie (approved):

PR #239, opened 2022-01-20
Bug 2040671: Fix the way the network stack is determined

sadasu:
We were looking at apiServerInternalURI and apiServerURL fields in the Infrastructure CR to figure out if the installation was a dualstack one. The problem with that approach was that these two URLs always had only one IP associated with them even in dualstack scenarios. That resulted in CBO passing the wrong IP_OPTIONS to metal3 containers.

With this revised approach, we are looking at the ServiceNetwork in the Network CR . In dualstack installs, the ServiceNetwork is expected to have 2 sets of IPs one for IPv4 and the other for IPv6.


zaneb:
The cluster [failed to build](https://prow.ci.openshift.org/view/gs/origin-ci-test/pr-logs/pull/openshift_cluster-baremetal-operator/239/pull-ci-openshift-cluster-baremetal-operator-master-e2e-metal-ipi-ovn-dualstack/1484183415744892928) because of errors with the authentication and console operators.
Let's try again to make sure.
/retest

zaneb:
It's still passing
```
    - name: IP_OPTIONS
      value: ip=dhcp
```

```
2022-01-20T16:07:47.042548326Z I0120 16:07:47.042494       1 provisioning_controller.go:495] "Network stack calculation" APIServerInternalHost="api.ostest.test.metalkube.org" NetworkStack=1
```

bfournie:
Looks like -metal-ipi-ovn-dualstack is failing because only 1 internal IP is defined
(https://github.com/openshift/origin/blob/master/vendor/k8s.io/kubernetes/test/e2e/network/dual_stack.go#L65)

STEP: Destroying namespace "e2e-dualstack-497" for this suite.
fail [k8s.io/kubernetes@v1.23.0/test/e2e/network/dual_stack.go:65]: Expected
    <int>: 1
to equal
    <int>: 2

failed: (3.5s) 2022-01-20T20:07:15 "[sig-network] [Feature:IPv6DualStack] should have ipv4 and ipv6 internal node ip [Suite:openshift/conformance/parallel] [Suite:k8s]"

zaneb:
Even if this doesn't work it should make what we're trying to do explicit, instead of the previous code that was highly misleading about what would actually happen.
/lgtm
/hold
(holding to give us time to see dualstack test results, feel free to unhold once it's complete)

sadasu:
We were looking at apiServerInternalURI and apiServerURL fields in the Infrastructure CR to figure out if the installation was a dualstack one. The problem with that approach was that these two URLs always had only one IP associated with them even in dualstack scenarios. That resulted in CBO passing the wrong IP_OPTIONS to metal3 containers.

With this revised approach, we are looking at the ServiceNetwork in the Network CR . In dualstack installs, the ServiceNetwork  is expected to have 2 sets of IPs one for IPv4 and the other for IPv6.

zaneb:
> We were looking at apiServerInternalURI and apiServerURL fields in the Infrastructure CR to figure out if the installation was a dualstack one. The problem with that approach was that these two URLs always had only one IP associated with them even in dualstack scenarios. That resulted in CBO passing the wrong IP_OPTIONS to metal3 containers.
> 
> With this revised approach, we are looking at the ServiceNetwork in the Network CR . In dualstack installs, the ServiceNetwork is expected to have 2 sets of IPs one for IPv4 and the other for IPv6.

This would be good info for the commit message :)

cybertron:
/lgtm

For some reason it looks like the dual stack job is still passing ip=dhcp to the nodes, but I agree that this looks like a step in the right direction.

ardaguclu:
Latest ipv6 failure is not related to this job;

/test e2e-metal-ipi-ovn-ipv6

zaneb:
> /lgtm
> 
> For some reason it looks like the dual stack job is still passing ip=dhcp to the nodes, but I agree that this looks like a step in the right direction.

I don't see that on the workers. I see it passing `ip=auto`, which is the default - see https://github.com/coreos/fedora-coreos-tracker/issues/1000

I _do_ see that on the control plane, because obviously this change doesn't affect the installer.

However, none of the worker Nodes came up with an IPv6 address, so I suspect we will need to use `ip=dhcp,dhcp6`.

sadasu:
/test e2e-agnostic

current failures don't seem related to current changes

sadasu:
/override ci/prow/e2e-metal-ipi-ovn-dualstack
Expected top pass with this and https://github.com/openshift/machine-config-operator/pull/2927.

PR #240, opened 2022-01-25
Bug 2045927: Add proxy for image-customization-controller

zhouhao3:
Pass the proxy when starting the image-customization-controller container.
Avoid the issue that ironic agent image cannot be downloaded due to network proxy.

See https://github.com/openshift/installer/issues/5552 for details.

Signed-off-by: Zhou Hao <zhouhao@fujitsu.com>

### DIFF
--- a/provisioning/image_customization.go
+++ b/provisioning/image_customization.go
@@ -83,6 +83,7 @@ func getUrlFromIP(ipAddr string) string {
 }
 
 func createImageCustomizationContainer(images *Images, info *ProvisioningInfo, ironicIP string) corev1.Container {
+	envVars := envWithProxy(info.Proxy, []corev1.EnvVar{})
 	container := corev1.Container{
 		Name:  "image-customization-controller",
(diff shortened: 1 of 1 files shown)


dtantsur:
Would it be possible to update the tests to cover this addition?

zhouhao3:
>Would it be possible to update the tests to cover this addition?

It looks like there are currently no tests related to ImageCustomization. Maybe we need to create a separate PR to add a complete test to ImageCustomization.

elfosardo (approved):

rhjanders:
> > Would it be possible to update the tests to cover this addition?
> 
> It looks like there are currently no tests related to ImageCustomization. Maybe we need to create a separate PR to add a complete test to ImageCustomization.

@zhouhao3 I check with Dmitry, given the circumstances it's OK if you add tests in a separate PR. Please follow up as soon as possible (can be after 4.10 Code Freeze :) Thank you.

PR #241, opened 2022-01-25
Upgrade library-go to latest

ardaguclu:
This PR upgrades library-go to latest versions. Latest library-go includes fixes for webhooks.


ardaguclu:
Since there are a lot of changes between current library-go and latest one, it is not a simple uplift. According to the jobs, there will be a couple of changes in CBO code itself. 

This upgrade can be done in next releases instead assuming as bug fix.

/close

PR #243, opened 2022-02-11
Bug 2099928: Add UT for image_customization_test

zhouhao3:
Follow-up to #240

### DIFF
--- a/provisioning/image_customization_test.go
+++ b/provisioning/image_customization_test.go
@@ -1,0 +1,110 @@
+/*
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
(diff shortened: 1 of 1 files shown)


zhouhao3:
Follow-up to #240

zhouhao3:
@elfosardo @dtantsur It looks like this PR requires the bugzilla/valid-bug label to merge, but there is no corresponding bugzilla. Is there a way to merge directly?

fenggw-fnst:
This PR is for the main branch, rather than for a specific branch or bug, so can we skip the bugzilla/valid-bug label? Could you please take a look? @elfosardo @dtantsur Thanks a lot!

elfosardo:
@fenggw-fnst hi! Since we're in Feature Freeze period, this will indeed need a BZ, thanks!

PR #245, opened 2022-02-15
Drop the IRONIC_HTTPD compatibility option

dtantsur:
Removed from ironic-image in https://github.com/openshift/ironic-image/pull/255


p0lyn0mial:
fyi `ci/prow/e2e-metal-ipi-serial-ipv4 ` will be fixed by https://github.com/openshift/origin/pull/27121

PR #251, opened 2022-03-11
Allow access to InfraEnv resources from assisted service

dtantsur:
We will need to read and modify them in image-customization-controller:
- https://github.com/openshift/image-customization-controller/pull/42

### DIFF
--- a/config/rbac/role.yaml
+++ b/config/rbac/role.yaml
@@ -42,6 +42,21 @@ rules:
   - update
   - watch
 - apiGroups:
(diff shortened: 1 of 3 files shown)


sadasu:
Do you  need to enable rbacs for watch?

dtantsur:
I don't think so. ICC is not going to use a client cache for these resources, nor watch for their changes.

Maybe later... definitely not in the first iteration.

flaper87 (commented):
/lgtm

PR #254, opened 2022-03-18
Revert "Allow access to InfraEnv resources from assisted service"

dtantsur:
Reverts openshift/cluster-baremetal-operator#251

Judging by the discussion on https://github.com/openshift/enhancements/pull/1059 and on slack we will not need these access rights.


dtantsur:
/hold

Waiting for the final decision

PR #256, opened 2022-04-28
[WIP] Use go 1.18

elfosardo:
related release config change https://github.com/openshift/release/pull/28193

PR #257, opened 2022-04-28
bump golangci

honza:
The previous version doesn't work with golang 1.18.  The latest version
seems to break rule G109:

    Potential Integer overflow made by strconv.Atoi result conversion to
    int16/32

It's broken in that it doesn't respect our #nosec directive.  I disabled
that rule in .golangci.yaml.

We also update go.mod, and vendor.

cc #256


sosiouxme:
I can't easily tell if it's already doing this - can you test what happens against go 1.18 or does that need to be a second PR?

.golangci.yml
```diff
@@ -36,3 +36,6 @@ issues:
    - linters:
      - staticcheck
      text: "SA1019:"
+   - linters:
```
dtantsur, line 39: Could you leave a FIXME to remove it once goling is fixed?

honza:
> I can't easily tell if it's already doing this - can you test what happens against go 1.18 or does that need to be a second PR?

I just opened #259 that combines the two to see if this works.

@sosiouxme

PR #263, opened 2022-05-12
Fix a few papercuts

honza:
* Use portable shebangs
* Don't cd into GOPATH dir in hack/generate.sh
* Remove a cyclic dependency in Makefile


elfosardo:
probably need to wait for https://github.com/openshift/release/pull/30412

PR #266, opened 2022-05-27
Bug 2088428: Fix interpretation of Deployment Status Conditions

sadasu:
Update the way metal3 Deployment status is reported using guidance from : https://kubernetes.io/docs/concepts/workloads/controllers/deployment/#complete-deployment

The earlier implementation assumed only one Deployment Status Condition could be true at a time. This resulted in the metal3 deployment status being reported incorrectly and as a consequence, the "baremetal" ClusterOperator was reporting the wrong status.

### DIFF
--- a/provisioning/baremetal_pod.go
+++ b/provisioning/baremetal_pod.go
@@ -966,12 +966,26 @@ func EnsureMetal3Deployment(info *ProvisioningInfo) (updated bool, err error) {
 }
 
 func getDeploymentCondition(deployment *appsv1.Deployment) appsv1.DeploymentConditionType {
+	var progressing, available, replicaFailure bool
 	for _, cond := range deployment.Status.Conditions {
-		if cond.Status == corev1.ConditionTrue {
-			return cond.Type
+		if cond.Type == appsv1.DeploymentProgressing && cond.Status == corev1.ConditionTrue {
+			progressing = true
+		}
+		if cond.Type == appsv1.DeploymentAvailable && cond.Status == corev1.ConditionTrue {
+			available = true
+		}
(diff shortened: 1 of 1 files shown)


provisioning/baremetal_pod.go
```diff
@@ -966,10 +966,26 @@ func EnsureMetal3Deployment(info *ProvisioningInfo) (updated bool, err error) {
...
+		}
+	}
+	if progressing == true && available == false && replicaFailure == false {
+		return appsv1.DeploymentProgressing
+	}
+	if available == true && replicaFailure == false {
+		return appsv1.DeploymentAvailable
+	}
+	if replicaFailure == true {
+		return appsv1.DeploymentReplicaFailure
```
zaneb, line 988: What about:

```
switch {
    case replicaFailure && !progressing:
        return appsv1.DeploymentReplicaFailure
    case available:
        return appsv1.DeploymentAvailable
    default:
        return appsv1.DeploymentProgressing
}
```

I think that covers everything from https://kubernetes.io/docs/concepts/workloads/controllers/deployment/#deployment-status?

sadasu: replicaFailure == true && progressing == True, represents a transient failure. And in that case we want to just report Progressing as True.
For appsv1.DeploymentAvailable, shouldn't we also be checking if there is no replicaFailure?

zaneb: could change it to `available && !replicaFailure` to make it more general in case somebody reuses this for a different deployment in future. I agree that for the metal3 deployment that case shouldn't be possible.

PR #275, opened 2022-08-03
METAL-157: Don't run image cache unless required

dtantsur:
* Move the machine-os-images init container from cache to ICC
* Disable the cache if ProvisioningOSDownloadURL is empty
* Do not report cache state if it is disabled


dtantsur:
/retest

Mirror issues

PR #276, opened 2022-08-05
[release-4.11] Bump golang to 1.18 and regenerate CRDs

honza:


dtantsur:
/lgtm

We'll need approval from someone like @dhellmann if we want to land it before the release.

honza:
> We'll need approval from someone like @dhellmann if we want to land it before the release.

I think it should be a no-brainer since it's essentially the same as what ART was already doing.

hardys:
Do we need to bump all-the-dependencies here, or can we just regenerate the CRD that was breaking the test on https://github.com/openshift/cluster-baremetal-operator/pull/264 ?

PR #277, opened 2022-08-08
Add RBAC for metal3.io/hardwaredata

honza:
This blocks https://github.com/openshift/baremetal-operator/pull/238


honza:
/retest

cbo comes up fine; seems like a timeout somewhere

PR #278, opened 2022-08-08
Stop passing HTPASSWD variables to services

dtantsur:
Authentication is handled by httpd in our setup, these are ignored
because of the REVERSE_PROXY variables, but it's potentially confusing.


dtantsur:

Job timeout, I wonder if it persists

PR #279, opened 2022-08-11
Run a proxy for ironic when provisioning network is disabled

dtantsur:
Without a provisioning network, ICC uses IP address of the control plane
node with the metal3 pod. If the pod relocates, the ramdisk will lose
access to Ironic. To solve this issue, run a simple HTTP(s) reverse proxy
as a DaemonSet on all control plane nodes.


provisioning/utils.go
```diff
@@ -47,8 +48,27 @@ func getPodHostIP(podClient coreclientv1.PodsGetter, targetNamespace string) (st
...
-func GetIronicIP(client kubernetes.Interface, targetNamespace string, config *metal3iov1alpha1.ProvisioningSpec) (string, error) {
-	if config.ProvisioningNetwork != metal3iov1alpha1.ProvisioningNetworkDisabled && !config.VirtualMediaViaExternalNetwork {
+func GetServerInternalIP(osclient osclientset.Interface) (string, error) {
+	infra, err := osclient.ConfigV1().Infrastructures().Get(context.Background(), "cluster", metav1.GetOptions{})
+	if err != nil {
+		err = fmt.Errorf("Cannot get the 'cluster' object from infrastructure API: %w", err)
+		return "", err
+	}
+
+	// NOTE(dtantsur): do we need to handle non-BareMetal platforms here?
```
dtantsur, line 58: I guess we do. OpenStack and vSphere have the same field, for None we can fall back to the previous process.

I suggest we do it in a follow-up to reduce the scope of this (already large) change.

provisioning/utils.go
```diff
@@ -47,8 +48,27 @@ func getPodHostIP(podClient coreclientv1.PodsGetter, targetNamespace string) (st
...
+	if infra.Status.PlatformStatus.BareMetal == nil {
+		err = fmt.Errorf("Cannot detect server API VIP: not a baremetal platform")
+		return "", err
+	}
+
+	// FIXME(dtantsur): handle the new APIServerInternalIPs field and the dualstack case.
```
dtantsur, line 64: Can be a follow-up as well IMO to avoid updating vendored dependencies in the same change.

dtantsur:
/hold

I see a problem here (not sure why the CI fails to catch it): ICC reuses the same IP for inspector.

dtantsur:
Now requires https://github.com/openshift/image-customization-controller/pull/60

rhjanders:
> /hold
> 
> I see a problem here (not sure why the CI fails to catch it): ICC reuses the same IP for inspector.

I wonder if maybe it worked by a co-incidence? (IPs worked out to be the same, or inspector was listening on * and it worked in this case, or something) It's been puzzling me too why CI didn't pick this up.

dtantsur:
Okay, this works in my testing. Once the ICC change merges, we can check the virtualmedia job and unhold the PR.

dtantsur:
I suspect we need to rerun the job once the ICC change is in containers.

provisioning/ironic_proxy.go
```diff
@@ -0,0 +1,255 @@
...
+			UpdateStrategy: appsv1.DaemonSetUpdateStrategy{
+				RollingUpdate: &appsv1.RollingUpdateDaemonSet{
+					MaxUnavailable: &maxUnavail,
+				},
+			},
+		},
+	}, nil
+}
+
+func UseIronicProxy(config *metal3iov1alpha1.ProvisioningSpec) bool {
```
honza, line 198: nit: `s/UseIronicProxy/ShouldUseIronicProxy/`

provisioning/utils.go
honza: :+1:

PR #285, opened 2022-08-31
Added handling PlatformStatus.Type other than Baremetal

rhjanders:


provisioning/utils.go
```diff
@@ -48,21 +49,24 @@ func getPodHostIP(podClient coreclientv1.PodsGetter, targetNamespace string) (st
...
-		err = fmt.Errorf("Cannot detect server API VIP: not a baremetal platform")
+	// FIXME(dtantsur): handle the new APIServerInternalIPs field and the dualstack case.
+	switch infra.Status.PlatformStatus.Type {
+	case osconfigv1.BareMetalPlatformType:
+		return infra.Status.PlatformStatus.BareMetal.APIServerInternalIP, nil
+	case osconfigv1.OpenStackPlatformType:
+		return infra.Status.PlatformStatus.OpenStack.APIServerInternalIP, nil
+	case osconfigv1.VSpherePlatformType:
+		return infra.Status.PlatformStatus.VSphere.APIServerInternalIP, nil
+	default:
```
dtantsur, line 66: Let's return "" in case of platform None and fall back to the pod IP.

rhjanders: Thank you Dmitry, fixed now.

provisioning/utils.go
```diff
@@ -77,7 +83,7 @@ func GetIronicIP(client kubernetes.Interface, targetNamespace string, config *me
 	}
 
 	if UseIronicProxy(config) {
-		ironicIP, err = GetServerInternalIP(osclient)
+		ironicIP, err = getServerInternalIP(osclient)
```
dtantsur, line 86: Can be empty now, needs fallback to `inspectorIP`

PR #288, opened 2022-09-14
Pass IRONIC_EXTERNAL_URL_V6 to baremetal-operator

honza:


honza:
It needs more of this logic:

https://github.com/metal3-io/ironic-image/blob/main/scripts/configure-ironic.sh#L38

PR #295, opened 2022-09-21
OCPBUGS-781: Update the deprecated field APIServerInternalIP to APIServerInternalIPs

hroyrh:
In provisioning/utils.go we reference .PlatformStatus.BareMetal.APIServerInternalIP attribute from the config API but a recent [change](https://github.com/openshift/api/commit/51f399230d604fa013c2bb341040c4ad126e7309) deprecated this field in favour of .APIServerInternalIPs (note plural), this was done to better suit dual stack case.


rhjanders:
/lgtm
/cc dtantsur
@dtantsur given this is a continuation of METAL-256, I figured it may be worth your look, LGTM.

provisioning/utils.go
```diff
@@ -59,25 +59,25 @@ func getPodHostIP(podClient coreclientv1.PodsGetter, targetNamespace string) (st
 	return hostIP, err
 }
 
-func getServerInternalIP(osclient osclientset.Interface) (string, error) {
+func getServerInternalIP(osclient osclientset.Interface) ([]string, error) {
 	infra, err := osclient.ConfigV1().Infrastructures().Get(context.Background(), "cluster", metav1.GetOptions{})
 	if err != nil {
 		err = fmt.Errorf("Cannot get the 'cluster' object from infrastructure API: %w", err)
-		return "", err
+		return []string{""}, err
```
dtantsur, line 66: We used "" as an empty value for the `string` type. The slice type has an empty value as well - `nil`, so let's return it. It's especially true for the error case.

dtantsur:
In addition to one comment inline: how was the vendored file updated? I'm surprised to see neither go.mod nor vendor/modules.txt changes.

dtantsur (changes requested):
Yep, confirmed locally: running `go mod vendor` reverts your change. You cannot just copy updated files from upstream into `vendor`, you need to update `go.mod`.

rhjanders:
Thanks for your insights Dmitry. Removing LGTM for now, will re-review once the new revision is up.
/cancel lgtm

provisioning/utils.go
hroyrh: In case of an error should we fall back to Pod IP ( as is the case for NonePlatform ) or simply return ?

rhjanders:
Thank you for the revised patch @hroyrh.

I had a look at the issue with the updated dependency, I think an approach like this may help (need to add context as an argument in several places in the code, this is just an example):

```
[janders@carbon cluster-baremetal-operator]$ git diff provisioning/baremetal_pod.go
diff --git a/provisioning/baremetal_pod.go b/provisioning/baremetal_pod.go
index 9c3a92f3..ef10c436 100644
--- a/provisioning/baremetal_pod.go
+++ b/provisioning/baremetal_pod.go
@@ -961,7 +961,7 @@ func getMetal3DeploymentSelector(client appsclientv1.DeploymentsGetter, targetNa
        return nil, err
 }
 
-func EnsureMetal3Deployment(info *ProvisioningInfo) (updated bool, err error) {
+func EnsureMetal3Deployment(ctx context.Context, info *ProvisioningInfo) (updated bool, err error) {
        // Create metal3 deployment object based on current baremetal configuration
        // It will be created with the cboOwnedAnnotation
 
@@ -975,7 +975,7 @@ func EnsureMetal3Deployment(info *ProvisioningInfo) (updated bool, err error) {
        }
 
        deploymentRolloutStartTime = time.Now()
-       deployment, updated, err := resourceapply.ApplyDeployment(info.Client.AppsV1(),
+       deployment, updated, err := resourceapply.ApplyDeployment(ctx, info.Client.AppsV1(),
                info.EventRecorder, metal3Deployment, expectedGeneration)
        if err != nil {
                err = fmt.Errorf("unable to apply Metal3 deployment: %w", err)
[janders@carbon cluster-baremetal-operator]$ 
```

hroyrh:
@rhjanders @dtantsur I've merged changes from #305 and made changes accordingly. Can you please take a look again if things look okay. Also, should I first do a rebase first instead of the merge ( to keep the commit history clean ? ) ?

dtantsur:
@hroyrh I'd really appreciate a rebase on top of master to avoid reviewing unrelated changes

hroyrh:
@dtantsur I have rebased against my initial changes, can you take a look ?

vendor/github.com/openshift/api/config/v1/types_infrastructure.go
```diff
@@ -521,7 +521,7 @@ type OvirtPlatformStatus struct {
 	// by components inside the cluster, like kubelets using the infrastructure rather
 	// than Kubernetes networking. It is the IP that the Infrastructure.status.apiServerInternalURI
 	// points to. It is the IP for a self-hosted load balancer in front of the API servers.
-	APIServerInternalIP string `json:"apiServerInternalIP,omitempty"`
+	APIServerInternalIPs []string `json:"apiServerInternalIPs,omitempty"`
```
honza, line 524: Changes to this file need to accompanied by changes to `go.mod`, and `go.sum`

provisioning/utils.go
```diff
@@ -64,31 +64,25 @@ func getPodHostIP(podClient coreclientv1.PodsGetter, targetNamespace string) (st
 	return pod.Status.HostIP, nil
 }
 
-func getServerInternalIP(osclient osclientset.Interface) (string, error) {
```
honza, line 67: Consider changing the name to `getServerInternalIPs`

provisioning/utils.go
```diff
@@ -104,7 +98,8 @@ func GetIronicIP(client kubernetes.Interface, targetNamespace string, config *me
 	}
 
 	if UseIronicProxy(config) {
-		ironicIP, err = getServerInternalIP(osclient)
+		internalIPs, _ := getServerInternalIP(osclient)
```
honza, line 101: Let's handle the error here

provisioning/utils.go
```diff
@@ -104,7 +98,11 @@ func GetIronicIP(client kubernetes.Interface, targetNamespace string, config *me
 	}
 
 	if UseIronicProxy(config) {
-		ironicIP, err = getServerInternalIP(osclient)
+		internalIPs, err1 := getServerInternalIPs(osclient)
+		if err1 != nil {
+			return
```
honza, line 103: The naked `return` will return whatever is stored in `err` as the error value.

hroyrh: @honza should I just print the error using `fmt` or return an error message. And I think I missed the case of `internalIPs=nil`, so will handle it as well.

honza: I think you should be able to assign to `err`, and then do a naked return.

```go
internalIPs, err := getServerInternalIPs(osclient)
if err != nil {
    return
}
```

hroyrh: @honza does this look okay now ?

provisioning/utils.go
```diff
@@ -64,36 +64,31 @@ func getPodHostIP(podClient coreclientv1.PodsGetter, targetNamespace string) (st
...
 	case osconfigv1.OpenStackPlatformType:
-		return infra.Status.PlatformStatus.OpenStack.APIServerInternalIP, nil
+		return infra.Status.PlatformStatus.OpenStack.APIServerInternalIPs, nil
 	case osconfigv1.VSpherePlatformType:
-		return infra.Status.PlatformStatus.VSphere.APIServerInternalIP, nil
-	case osconfigv1.AWSPlatformType:
-		return "", nil
-	case osconfigv1.AzurePlatformType:
-		return "", nil
-	case osconfigv1.GCPPlatformType:
```
dtantsur, line 85: These 3 should be handled the same way as None, not removed

provisioning/utils.go
```diff
@@ -113,7 +117,7 @@ func GetIronicIP(client kubernetes.Interface, targetNamespace string, config *me
 		ironicIP = inspectorIP
 	}
 
-	return
+	return ironicIP, inspectorIP, err
```
dtantsur, line 120: nit: I don't think this change is needed

hroyrh: When I do that, the unit tests throw error : `naked return in func "GetIronicIP" with 32 lines of code (nakedret)` , and explicit return also improves readability I think.

provisioning/utils.go
hroyrh: Done

dtantsur:
/approve

Please check why the CI is failing

hroyrh:
@dtantsur is the sdn-upgrade job also required to pass before merging this pr ?

honza:
Could use a rebase, not a fan of "minor fixes" commits

hroyrh:
> Could use a rebase, not a fan of "minor fixes" commits

Done

PR #296, opened 2022-09-21
Update with current OWNERS

elfosardo:


hardys:
Manually applied `bugzilla/valid-bug` since this is a non-code change, and it's consistent with the OWNERS file back to 4.10

PR #302, opened 2022-10-03
Add port to IRONIC_EXTERNAL_URL_V6

honza:


dtantsur:
Still sleepy here, but shouldn't it be the value of VMEDIA_TLS_PORT in the end?

dtantsur:
Hmm, I also don't see HTTP_PORT used anywhere in BMO now, is it for some WIP patch?

honza:
> Still sleepy here, but shouldn't it be the value of VMEDIA_TLS_PORT in the end?

Good point; added that value as well.

> Hmm, I also don't see HTTP_PORT used anywhere in BMO now, is it for some WIP patch?

We  previously 08d1b7e241dc57ec7cf699b8a03b5f8d0d296326 started passing a new env var to the BMO here in the CBO. This var depends on other vars which need to be made available. Two of those vars weren't passed in in the original patch; we supply those values here.

PR #306, opened 2022-10-13
Golang 1.19

honza:
We have to upgrade golangci-lint to make 1.19 work.


honza:
Broken jobs haven't run since https://github.com/openshift/release/pull/33183 merged.

/retest

PR #309, opened 2022-11-28
Enable running CBO from Azure

elfosardo:
Handle Azure platform


giannisalinetti:
This PR is very important to accelerate the usage of CBO on Azure. Together with already merged PRs for AWS it will also help accelerate the adoption of RHACM as a ZTP hub for edge use cases. IHAC which is seeking exactly this scenario and looks forward to the merge of this PR and its GA downstream.

PR #310, opened 2022-11-28
Ensure we set ipv6 url in external url var

honza:


provisioning/utils.go
```diff
@@ -107,6 +107,62 @@ func GetIronicIP(client kubernetes.Interface, targetNamespace string, config *me
 	return
 }
 
+func GetPodIP(client kubernetes.Interface, targetNamespace string, networkType NetworkStackType) (string, error) {
```
dtantsur, line 110: This is essentially a copy of getPodHostIP, could these be refactored into one call?

provisioning/baremetal_pod.go
```diff
@@ -277,39 +278,34 @@ func setIronicExternalIp(name string, config *metal3iov1alpha1.ProvisioningSpec)
...
 }
 
-func setIronicExternalUrl(config *metal3iov1alpha1.ProvisioningSpec) []corev1.EnvVar {
+func setIronicExternalUrl(client kubernetes.Interface, config *metal3iov1alpha1.ProvisioningSpec, namespace string) corev1.EnvVar {
 	if config.ProvisioningNetwork != metal3iov1alpha1.ProvisioningNetworkDisabled && config.VirtualMediaViaExternalNetwork {
-		envVars := []corev1.EnvVar{
-			setIronicExternalIp(externalIpEnvVar, config),
+		ipv6PodIP, err := GetPodIP(client, namespace, NetworkStackV6)
+
+		if err != nil {
```
dtantsur, line 285: The IP won't be available in advance, so I assume you're relying on updating the pod after it's running?

PR #315, opened 2022-12-14
OCPBUGS-4712: delete ironic-proxy/image-cache when not needed

dtantsur:
Keeping ironic-proxy active (e.g. when switching from provisioning
network Disabled to Managed) causes a conflict with the metal3 pod.


dtantsur:
/hold

Will do local testing

dtantsur:
/hold cancel

I think this approach works

PR #316, opened 2022-12-19
Enable running CBO from GCP

elfosardo:
Handle GCP platform


dtantsur:
/approve

Will conflict with the Azure one.

PR #322, opened 2023-01-16
OCPBUGS-5461: Add ironic IP to no_proxy

derekhiggins:
Ensure internal ironic traffic bypasses any proxies. The NO_PROXY value created in createImageCustomizationContainer is what IPA will end up using.


dtantsur:
/approve

The linter fails, seems like you need to run `gofmt -s`

PR #324, opened 2023-01-27
OCPBUGS-4248: Set external Ironic URL to ironic-proxy IP

honza:
We need to set the external url when
	
1. IPv6 is enabled
2. `VirtualMediaViaExternalNetwork` is enabled
	
When `VirtualMediaViaExternalNetwork` is enabled, ironic-proxy is also enabled.  This means that we can set the external url to be the IPv6 version of ironic-proxy.

In this patch, we

1.  Update openshift dependencies to gain access to `APIServerInternalIPs` (note the plural)
2.  Update the `GetIronicIP` function to `GetIronicIPs` (and `getServerInternalIP` to `getInternalIPs`), and thereby force all callers to deal with the dual stack scenario
3.  Pass both IPv4 and IPv6 IPs to ironic-proxy so it can listen to both (Ironic image patch forthcoming)
4. Set the external url only when the proxy and virtual media are enabled.

### DIFF
--- a/provisioning/baremetal_pod.go
+++ b/provisioning/baremetal_pod.go
@@ -26,8 +26,9 @@ import (
 	"k8s.io/apimachinery/pkg/api/equality"
 	"k8s.io/apimachinery/pkg/api/resource"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
-	"k8s.io/client-go/kubernetes"
 	appsclientv1 "k8s.io/client-go/kubernetes/typed/apps/v1"
+	"k8s.io/klog/v2"
+	utilnet "k8s.io/utils/net"
 	"k8s.io/utils/pointer"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 	"sigs.k8s.io/controller-runtime/pkg/controller/controllerutil"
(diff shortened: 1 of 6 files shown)


provisioning/baremetal_pod.go
```diff
@@ -278,34 +278,64 @@ func setIronicExternalIp(name string, config *metal3iov1alpha1.ProvisioningSpec)
 	}
 }
 
-func setIronicExternalUrl(client kubernetes.Interface, config *metal3iov1alpha1.ProvisioningSpec, namespace string) corev1.EnvVar {
-	if config.ProvisioningNetwork != metal3iov1alpha1.ProvisioningNetworkDisabled && config.VirtualMediaViaExternalNetwork {
-		ipv6PodIP, err := GetPodIP(client.CoreV1(), namespace, NetworkStackV6)
+func setIronicExternalUrl(info ProvisioningInfo) corev1.EnvVar {
+	// We need to set the external url when
```
zaneb, line 282: This is ambiguous as to whether it is an AND or OR condition. (Assuming it's AND?)

provisioning/baremetal_pod.go
```diff
@@ -278,34 +278,64 @@ func setIronicExternalIp(name string, config *metal3iov1alpha1.ProvisioningSpec)
...
+	//
+	//   1. IPv6 is enabled
+	//   2. VirtualMediaViaExternalNetwork is enabled
+	//
+	// When VirtualMediaViaExternalNetwork is enabled, ironic-proxy is also
+	// enabled.  This means that we can set the external url to be the IPv6
+	// version of ironic-proxy.
```
zaneb, line 289: This seems too restrictive on the face of it.

`VirtualMediaViaExternalNetwork` is for the specific case where there _is_ a provisioning network, but we _don't_ want to use it for virtualmedia deploys (i.e. provisioning network is only for PXE).
It seems like the proxy, and also this, should also be enabled in the case where there is _not_ a provisioning network. In that case there is only the external network and it goes without saying that all virtualmedia deployments go over that network.

BTW does the logic in BMO know to only set the alternative URL on the node for virtualmedia deployments? (I know it only does it when the BMC has an IPv6 address.) If not that would mess up PXE deployments when `VirtualMediaViaExternalNetwork` is enabled and the PXE network is IPv6.

provisioning/baremetal_pod.go
```diff
@@ -278,34 +278,64 @@ func setIronicExternalIp(name string, config *metal3iov1alpha1.ProvisioningSpec)
...
-			return corev1.EnvVar{
-				Name:  externalUrlEnvVar,
-				Value: fmt.Sprintf(urlTemplate, "http", ipv6PodIP, baremetalHttpPort),
-			}
-		} else {
-			return corev1.EnvVar{
-				Name:  externalUrlEnvVar,
-				Value: fmt.Sprintf(urlTemplate, "https", ipv6PodIP, baremetalVmediaHttpsPort),
-			}
+	if err != nil {
```
zaneb, line 305: It's surprising to see this error suppressed.

provisioning/baremetal_pod.go
zaneb: > BTW does the logic in BMO know to only set the alternative URL on the node for virtualmedia deployments? (I know it only does it when the BMC has an IPv6 address.) If not that would mess up PXE deployments when `VirtualMediaViaExternalNetwork` is enabled and the PXE network is IPv6.

It looks like Ironic does not use the `external_http_url` for PXE, so we should be ok.

provisioning/utils.go
```diff
@@ -64,31 +64,31 @@ func getPodHostIP(podClient coreclientv1.PodsGetter, targetNamespace string) (st
...
 	if err != nil {
 		err = fmt.Errorf("Cannot get the 'cluster' object from infrastructure API: %w", err)
-		return "", err
+		return nil, err
 	}
 	// FIXME(dtantsur): handle the new APIServerInternalIPs field and the dualstack case.
 	switch infra.Status.PlatformStatus.Type {
 	case osconfigv1.BareMetalPlatformType:
-		return infra.Status.PlatformStatus.BareMetal.APIServerInternalIP, nil
+		return infra.Status.PlatformStatus.BareMetal.APIServerInternalIPs, nil
```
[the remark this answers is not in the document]
zaneb: What was the answer?

honza:
* Simplify comment
* Change the signature of `GetIronicIPs`
* Propagate errors to manager when setting extenal url

provisioning/utils.go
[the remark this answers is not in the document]
zaneb: It appears this is handled by https://github.com/openshift/cluster-network-operator/pull/1519

zaneb (commented):
/lgtm

provisioning/ironic_proxy.go
```diff
@@ -24,13 +25,66 @@ import (
...
 const (
-	ironicProxyService       = "ironic-proxy"
-	ironicPrivatePort        = 6388
-	ironicUpstreamIPEnvVar   = "IRONIC_UPSTREAM_IP"
-	ironicUpstreamPortEnvVar = "IRONIC_UPSTREAM_PORT"
+	ironicProxyService         = "ironic-proxy"
+	ironicPrivatePort          = 6388
+	ironicPrivatePortv6        = 6388
+	ironicUpstreamIPEnvVar     = "IRONIC_UPSTREAM_IP"
+	ironicUpstreamIPv6EnvVar   = "IRONIC_UPSTREAM_IP_V6"
```
dtantsur, line 32: The `_V6` variables do not exist (nor are they needed, one variable is enough).

dtantsur:
Please see https://github.com/openshift/cluster-baremetal-operator/pull/335

dtantsur:
I think this is the cause of confusion here:

> Pass both IPv4 and IPv6 IPs to ironic-proxy so it can listen to both

`IRONIC_UPSTREAM_*` variables do not define what ironic-proxy listens on, they define where it proxies to. I.e. the address of the ironic pod, which on dualstack may have any IP as long as this IP is reachable within the control plane.

PR #327, opened 2023-02-16
allow bmo to read namespace info

pixelsoccupied:
see https://github.com/openshift/baremetal-operator/pull/251 

/cc @dtantsur
/cc @rhjanders

### DIFF
--- a/config/rbac/role.yaml
+++ b/config/rbac/role.yaml
@@ -30,6 +30,12 @@ rules:
   - patch
   - watch
 - apiGroups:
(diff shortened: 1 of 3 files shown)


zaneb:
This looks fine, but I'm not certain we actually need to make use of these permissions. See comments on openshift/baremetal-operator#251.

/hold

pixelsoccupied:
We dont need this anymore. see https://github.com/metal3-io/baremetal-operator/pull/1226

PR #329, opened 2023-03-16
OCPBUGS-8597: Force mongodb version to 1.5.1

elfosardo:


dtantsur:
I don't mind this, but do we really care? We don't use anything mongo-related.

honza:
It would be nice to include the reason in the commit message.

elfosardo:
after a quick talk we realized this is not really needed as no code from mongodb is actually used from CBO or its dependencies

PR #335, opened 2023-04-13
Revert "proxy: pass ipv6 address to container"

dtantsur:
First and foremost, it made the proxy refer to itself, which only
happens to work when the API VIP is on the same master as the metal3
pod. Proxy must use the real pod IP.

The UPSTREAM_*_V6 variables do not exist in ironic-image, setting
them does nothing. On top of that, passing several upstream IPs does not
make much sense: we expect that Ironic is available on any of the valid
IPs, and we can make the proxy contact any of them.

This reverts commit ba7403e57e344e615d375af5cb7bfec5d56f03e9.

### DIFF
--- a/provisioning/ironic_proxy.go
+++ b/provisioning/ironic_proxy.go
@@ -13,7 +13,6 @@ import (
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/util/intstr"
 	appsclientv1 "k8s.io/client-go/kubernetes/typed/apps/v1"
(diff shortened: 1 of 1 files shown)


zaneb:
Do we expect everything to be correct after this revert, or is there something else we now need to do instead?

dtantsur:
> Do we expect everything to be correct after this revert, or is there something else we now need to do instead?

I hope that the external IP PR still works. This change only concerns how the proxy itself connects to the pod, while the Honza's PR was about using the proxy. Worth confirming though since virtual media is not tested with dualstack in the CI.

PR #338, opened 2023-04-19
[release-4.13] OCPBUGS-11985: use proxying for inspector in addition to ironic

dtantsur:
Apparently, inspector has the same issue as ironic: there is
a possibility to hit an outdated URI.

As a result, both Ironic and Inspector are always reachable via
the API VIP. I hope nobody asks the same for the image server.

Requires:
- https://github.com/openshift/ironic-image/pull/361

(cherry picked from commit 768b991bda3cf88dd7c4e789e0788e57b4ff2da8)


dtantsur:
/retest

The dependency has merged.

PR #341, opened 2023-05-04
OCPBUGS-4248: Avoid ironic-proxy when setting external url

honza:
The external url is the url the BMC will use to retrieve images from the httpd container inside metal3 in dual stack scenarios.

Each time the metal3 pod is started, we have to find any existing nodes, and rebuild the Ironic database.  During this process, we also tell each BMC where to get images.  This means that httpd doesn't have to be proxied, and therefore, when we construct the external url, we should ignore the ironic-proxy altogether.

This substantially changes the work done in https://github.com/openshift/cluster-baremetal-operator/pull/324.

### DIFF
--- a/provisioning/baremetal_pod.go
+++ b/provisioning/baremetal_pod.go
@@ -280,16 +280,7 @@ func setIronicExternalIp(name string, config *metal3iov1alpha1.ProvisioningSpec)
 }
 
 func setIronicExternalUrl(info ProvisioningInfo) (corev1.EnvVar, error) {
-	// We need to set the external URL to point to the ironic-proxy when IPv6
-	// is enabled and the proxy is present
-
-	if !UseIronicProxy(&info.ProvConfig.Spec) {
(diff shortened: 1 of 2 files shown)


honza:
/retitle OCPBUGS-4248: Avoid ironic-proxy when setting external url

zaneb:
This is using the API VIP as the address of the ironic image cache. Two problems with this:

* [OCPBUGS-23149](https://issues.redhat.com/browse/OCPBUGS-23149) - this doesn't work if the API VIP is not on the same host as the metal3 Pod, which is not the case unless by coincidence
* [OCPBUGS-23131](https://issues.redhat.com/browse/OCPBUGS-23131) - if there is a provisioning network and `virtualMediaViaExternalNetwork` is not set, then this is on the wrong network - we should use the provisioning VIP (which _is_ guaranteed to be on the same host as the metal3 Pod) instead

PR #342, opened 2023-05-08
OCPBUGS-14064: Split baremetal-operator into a separate pod

dtantsur:
There is no reason to keep it tied to Ironic. Actually, it causes
constant chicken-and-egg problems where BMO needs to know the location
of Ironic.


dtantsur:
/retest

I don't understand why github does not register the update here...

controllers/provisioning_controller.go
```diff
@@ -328,7 +329,7 @@ func (r *ProvisioningReconciler) Reconcile(ctx context.Context, req ctrl.Request
 		}
 	}
 
-	// Determine the status of the deployment
+	// Determine the status of the baremetal deployment
```
zaneb, line 332: metal3?

PR #343, opened 2023-06-13
NO-ISSUE: Stop linting when building the image

honza:
When we build the image, the code has already been through code review, and CI.  Everything has been generated, and linted, and addressed.  At the build stage, we just want to compile the binary.

Furthermore, the golangci-lint program seems to require attention every time we upgrade Go in the builder.  This breaks the build, and we have to do emergency upgrades to get it unstuck.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -19,7 +19,7 @@ IMAGES_JSON ?= /etc/cluster-baremetal-operator/images/images.json
 # Set VERBOSE to -v to make tests produce more output
 VERBOSE ?= ""
 
-all: cluster-baremetal-operator
+all: generate lint build
 
 # Run tests
 test: generate lint manifests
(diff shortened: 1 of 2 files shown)


Makefile
```diff
@@ -29,7 +29,7 @@ test: generate lint manifests
 unit: test
 
 # Build cluster-baremetal-operator binary
-cluster-baremetal-operator: generate lint
+build:
 	go build -o bin/cluster-baremetal-operator main.go
 
 # Run against the configured Kubernetes cluster in ~/.kube/config
```
elfosardo, line 35: remove lint from here and remove L114

elfosardo (changes requested):

Makefile
honza: Done

dtantsur:
/retitle NO-ISSUE: Stop linting when building the image
/lgtm

PR #344, opened 2023-06-15
Reduce container privilege

zaneb:
Many of the containers we run have to be run in privileged mode in order to access a host volume - not only so they can write files as root, but also to get an SELinux context that allows it.

However, some containers currently running as privileged do not require this.


zaneb:
I have tested these Deployment configurations locally and they appear to work, so it's not clear to me why CI is failing - particularly at such an early stage - and artifacts that might help debug it are not available.

dtantsur:
/retest

I think the CI was badly broken last week.

zaneb:
/reopen
Let's see if the tests are in better shape

zaneb:
Opened #396

PR #348, opened 2023-07-07
OCPBUGS-15232: Use Service instead of provisioning IP for BMO to talk to Ironic

dtantsur:
Using provisioning network addresses for communication between BMO and
Ironic seems to work but is a questionable design. Use the existing metal3-state
service instead.

Follow-up to c1bf1cba12c17990d43a15b4689189097f4f6116.

### DIFF
--- a/provisioning/baremetal_config.go
+++ b/provisioning/baremetal_config.go
@@ -91,8 +91,9 @@ func getIronicInspectorEndpoint() *string {
 	return &ironicInspectorEndpoint
 }
 
-func getRemoteEndpoints(info *ProvisioningInfo) (ironicEndpoint string, inspectorEndpoint string, err error) {
-	ironicIPs, inspectorIPs, err := GetIronicIPs(*info)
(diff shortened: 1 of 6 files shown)


dtantsur:
/retest

A bunch of different errors.

dtantsur:
/retest

Various issues, don't seem related.

dtantsur:
/retitle OCPBUGS-15232: Use Service instead of provisioning IP for BMO to talk to Ironic

zaneb:
Historical note: the name 'state' in the Service was meant to indicate that it held stateful stuff like MariaDB and Apache, and the intention was for stateless stuff like ironic-api to eventually be split out into its own (scalable) Deployment with its own Service.
Given that we are now running monolithic ironic, it seems unlikely that we are going to make that split in the near future. Still, I wouldn't be opposed to creating a new Service so that the naming can remain consistent no matter what deployment topology we have in the future. (You'll want to do this in your ironic operator I think.)

PR #350, opened 2023-07-21
OCPBUGS-16503: Uplift libraries to remove goproxy calls

elfosardo:


honza:
/hold 

Could you point out which goproxy calls we are eliminating? I think we could use a more fleshed out commit message as the change is a bit cryptic.

elfosardo:
> /hold
> 
> Could you point out which goproxy calls we are eliminating? I think we could use a more fleshed out commit message as the change is a bit cryptic.

@honza calls are not direct from CBO but from its dependencies controller-tools and kube-storage-version-migrator

honza:
> calls are not direct from CBO but from its dependencies controller-tools and kube-storage-version-migrator

All of the code is vendored.  I don't see any goproxy code being eliminated, and I assume I must be looking the wrong places. If we can't show in the diffs where these calls are, this PR is unnecessary.

elfosardo:
> > calls are not direct from CBO but from its dependencies controller-tools and kube-storage-version-migrator
> 
> All of the code is vendored. I don't see any goproxy code being eliminated, and I assume I must be looking the wrong places. If we can't show in the diffs where these calls are, this PR is unnecessary.

@honza from my environment I saw a diff with goproxy calls but I think there was some outdated content as I can't see that anymore, and also in the PR (I didn't realize it was not showing up in the commit)
thanks for the check, I'll close this PR

PR #351, opened 2023-07-27
reconcile status when clusteroperator changes

sanchezl:
Needed for openshift/cluster-kube-apiserver-operator#1423.

If status is modified, the operator must set status to proper values. This ensures that accidents cannot permanently reset status and gives a clear indication that the operator is "live". This came up as important when operators were NOT live, during cert rotations and we had no indication of problems.


controllers/provisioning_controller.go
```diff
@@ -200,14 +204,6 @@ func getSuccessStatus(imageCacheState appsv1.DaemonSetConditionType, ironicProxy
...
 // resource changes
 func (r *ProvisioningReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Result, error) {
-	// provisioning.metal3.io is a singleton
-	// Note: this check is here to make sure that the early startup configuration
-	// is correct. For day 2 operatations the webhook will validate this.
-	if req.Name != metal3iov1alpha1.ProvisioningSingletonName {
-		klog.Info("ignoring invalid CR", "name", req.Name)
-		return ctrl.Result{}, nil
-	}
-
```
sanchezl, line 210: I replaced this with a predicate (see method `SetupWithManager`), so only the expected resource will trigger.

controllers/provisioning_controller.go
```diff
@@ -603,14 +599,6 @@ func (r *ProvisioningReconciler) SetupWithManager(mgr ctrl.Manager) error {
...
 	// Check the Platform Type to determine the state of the CO
 	enabled := IsEnabled(r.EnabledFeatures)
-	if !enabled {
-		//Set ClusterOperator status to disabled=true, available=true
-		err = r.updateCOStatus(ReasonUnsupported, "Nothing to do on this Platform", "")
-		if err != nil {
-			return fmt.Errorf("unable to put %q ClusterOperator in Disabled state: %w", clusterOperatorName, err)
-		}
-		return nil
-	}
```
sanchezl, line 613: The `Reconcile()` method will handle this case. 
@deads2k  Note that commenting this out means that the controller will be started, but it will always reconcile correctly. The `deadman` controller currently has a special case to ignore `ClusterOperators` with `condition[@type=Disabled]==true`. Removing this here makes that special case superfluous for the baremetal operator.

controllers/provisioning_controller.go
dtantsur: The current version has a benefit of logging what went wrong though.

controllers/provisioning_controller.go
```diff
@@ -630,14 +618,18 @@ func (r *ProvisioningReconciler) SetupWithManager(mgr ctrl.Manager) error {
 
 	klog.InfoS("Network stack calculation", "NetworkStack", r.NetworkStack)
 
+	provisioningFilter := predicate.NewPredicateFuncs(func(object client.Object) bool { return object.GetName() == metal3iov1alpha1.ProvisioningSingletonName })
```
dtantsur, line 621: Would it be appropriate to log here if the name does not match?

sanchezl: I do not see a need. The operator is interested in this particular resource, otherwise it just doesn't care.

dtantsur: I'm worried about users creating a wrong resource and wondering why it's not reconciled.

elfosardo:
/lgtm
I doubt the error in CI is related to this patch

controllers/provisioning_controller.go
```diff
@@ -630,14 +618,24 @@ func (r *ProvisioningReconciler) SetupWithManager(mgr ctrl.Manager) error {
...
-		For(&metal3iov1alpha1.Provisioning{}).
+		For(&metal3iov1alpha1.Provisioning{}, builder.WithPredicates(provisioningFilter)).
 		Owns(&corev1.Secret{}).
 		Owns(&appsv1.Deployment{}).
 		Owns(&corev1.Service{}).
 		Owns(&appsv1.DaemonSet{}).
 		Owns(&osconfigv1.ClusterOperator{}).
 		Owns(&osconfigv1.Proxy{}).
 		Owns(&machinev1beta1.Machine{}).
+		Watches(&source.Kind{Type: &osconfigv1.ClusterOperator{}}, &handler.EnqueueRequestForObject{}, builder.WithPredicates(clusterOperatorFilter)).
```
zaneb, line 639: @sanchezl Does this do anything?
AIUI `EnqueueRequestForObject{}` makes the controller reconcile the Provisioning CR with the same name and namespace as the ClusterOperator that changed. Which of course doesn't exist, hence why you had to add a predicate to filter it out again, leaving us exactly where we started.
This appears to me to be just as ineffective as our previous attempt to watch the ClusterOperator on line 636, which has no effect because the Provisioning CR does not own any ClusterOperator CRs (and nor should it).

zaneb: https://issues.redhat.com/browse/OCPBUGS-36357

PR #352, opened 2023-07-30
Jira OCPBUGS-16919: Trigger reconcile if Secret openshift-config/pull-secret changes

andreaskaris:
When secret openshift-config/pull-secret changes, secret openshift-machine-api/pull-secret shall be updated. In order to achieve this:
* Add a new Watch to the ProvisioningReconciler.
* In method createRegistryPullSecret, recreate the secret in openshift-machine-api if it does not match the secret in openshift/config.

### Testing instructions

Spawn cluster with clusterbot:
~~~
launch openshift/cluster-baremetal-operator#352 ovn,metal
~~~

Modify secret openshift-config/pull-secret:
~~~
mkdir pull1
pushd pull1
oc extract secret/pull-secret -n openshift-config
vim .dockerconfigjson   # modify the pull-secret
oc create secret docker-registry --from-file .dockerconfigjson -n openshift-config pull-secret -o yaml --dry-run=true | oc apply -f -
popd
~~~

Check the log of the cluster-baremetal-operator-... pod, it should show:
~~~
$ oc logs -n openshift-machine-api -l  k8s-app=cluster-baremetal-operator | grep "content of secret"
Defaulted container "cluster-baremetal-operator" out of: cluster-baremetal-operator, baremetal-kube-rbac-proxy
I0823 17:17:18.778372       1 baremetal_secrets.go:154] content of secret openshift-config/pull-secret does not match content of secret openshift-machine-api/pull-secret, reconciling openshift-machine-api/pull-secret
~~~

Make sure that secret openshift-machine-api/pull-secret was updated:
~~~
mkdir pullverify
pushd pullverify
oc extract secret/pull-secret -n openshift-machine-api
cat .dockerconfigjson | base64 -d | less     # make sure that the pull-secret matches the changes
popd
~~~

### DIFF
--- a/controllers/provisioning_controller.go
+++ b/controllers/provisioning_controller.go
@@ -46,6 +46,7 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/controller/controllerutil"
 	"sigs.k8s.io/controller-runtime/pkg/handler"
 	"sigs.k8s.io/controller-runtime/pkg/predicate"
+	"sigs.k8s.io/controller-runtime/pkg/reconcile"
 	"sigs.k8s.io/controller-runtime/pkg/source"
 
 	baremetalv1alpha1 "github.com/metal3-io/baremetal-operator/apis/metal3.io/v1alpha1"
@@ -627,6 +628,28 @@ func (r *ProvisioningReconciler) SetupWithManager(mgr ctrl.Manager) error {
 	})
 	clusterOperatorFilter := predicate.NewPredicateFuncs(func(object client.Object) bool { return object.GetName() == clusterOperatorName })
 
+	// Watch Secret openshift-config/pull-secret. If this secret changes, we must requeue the provisioning Singleton,
+	// if it exists.
+	watchOCPConfigPullSecret := func(object client.Object) []reconcile.Request {
+		prov, err := r.readProvisioningCR(ctx)
+		// readProvisioningCR can return nil, nil upon IsNotFound, account for this as well.
+		if err != nil || prov == nil {
+			return []reconcile.Request{}
+		}
+		return []reconcile.Request{
+			{
+				NamespacedName: types.NamespacedName{
+					Namespace: prov.Namespace,
+					Name:      prov.Name,
+				},
+			},
+		}
+	}
+	secretFilter := predicate.NewPredicateFuncs(func(object client.Object) bool {
+		return object.GetNamespace() == provisioning.OpenshiftConfigNamespace &&
+			object.GetName() == provisioning.PullSecretName
+	})
+
 	return ctrl.NewControllerManagedBy(mgr).
 		For(&metal3iov1alpha1.Provisioning{}, builder.WithPredicates(provisioningFilter)).
 		Owns(&corev1.Secret{}).
@@ -637,5 +660,6 @@ func (r *ProvisioningReconciler) SetupWithManager(mgr ctrl.Manager) error {
 		Owns(&osconfigv1.Proxy{}).
 		Owns(&machinev1beta1.Machine{}).
 		Watches(&source.Kind{Type: &osconfigv1.ClusterOperator{}}, &handler.EnqueueRequestForObject{}, builder.WithPredicates(clusterOperatorFilter)).
+		Watches(&source.Kind{Type: &corev1.Secret{}}, handler.EnqueueRequestsFromMapFunc(watchOCPConfigPullSecret), builder.WithPredicates(secretFilter)).
 		Complete(r)
 }
--- a/provisioning/baremetal_secrets.go
+++ b/provisioning/baremetal_secrets.go
@@ -12,6 +12,7 @@ import (
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	utilerrors "k8s.io/apimachinery/pkg/util/errors"
 	coreclientv1 "k8s.io/client-go/kubernetes/typed/core/v1"
+	"k8s.io/klog/v2"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 	"sigs.k8s.io/controller-runtime/pkg/controller/controllerutil"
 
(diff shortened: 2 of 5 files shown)


controllers/provisioning_controller.go
```diff
@@ -639,5 +664,6 @@ func (r *ProvisioningReconciler) SetupWithManager(mgr ctrl.Manager) error {
 		Owns(&osconfigv1.ClusterOperator{}).
 		Owns(&osconfigv1.Proxy{}).
 		Owns(&machinev1beta1.Machine{}).
+		Watches(&source.Kind{Type: &corev1.Secret{}}, handler.EnqueueRequestsFromMapFunc(watchOCPConfigPullSecret)).
```
andreaskaris, line 667: AFAICT it's either that, or it's a completely new (independent) controller to watch the secret inside openshift-config

andreaskaris:
### Testing instructions

Modify secret openshift-config/pull-secret:
~~~
mkdir pull1
pushd pull1
oc extract secret/pull-secret -n openshift-config
vim .dockerconfigjson   # modify the pull-secret
oc create secret docker-registry --from-file .dockerconfigjson -n openshift-config pull-secret -o yaml --dry-run=true | oc apply -f -
popd
~~~

Check the log of the cluster-baremetal-operator-... pod, it should show:
~~~
I0730 22:02:35.074854       1 baremetal_secrets.go:154] content of secret openshift-config/pull-secret does not match content of secret openshift-machine-api/pull-secret, reconciling openshift-machine-api/pull-secret
~~~

Make sure that secret openshift-machine-api/pull-secret was updated:
~~~
mkdir pullverify
pushd pullverify
oc extract secret/pull-secret -n openshift-machine-api
cat .dockerconfigjson | base64 -d | less     # make sure that the pull-secret matches the changes
~~~

elfosardo:
cc @honza @dtantsur

elfosardo:
@andreaskaris PR needs rebase

controllers/provisioning_controller.go
zaneb: This appears to watch every Secret in every namespace, which unfortunately (a) is pretty unsafe (we'll end up caching every user's Secret data in CBO), and (b) will 100% for sure chew up all of the RAM and cause the cluster to fall over, not necessarily in that order.

If we can't control this here, we need to control the scope in the Cache builder.

A related (but different) example: https://github.com/openshift/baremetal-operator/blob/master/main.go#L207-L209

andreaskaris: Hey, thanks for the comment.

The security issue isn't a problem of my commit, but of the following line, which literally gives the cluster-baremetal-operator the right to do anything with any configmaps, secrets or services in the entire cluster:
https://github.com/openshift/cluster-baremetal-operator/blob/5b91a35afaf62c59c289c85bd561a7fa308352ce/controllers/provisioning_controller.go#L106

For the performance concern, check main.go:
```
-	        Namespace:          controllers.ComponentNamespace,
+		NewCache: cache.MultiNamespacedCacheBuilder(
			[]string{controllers.ComponentNamespace, provisioning.OpenshiftConfigNamespace}),
```
The operator used to only watch resources in its own namespace, and with the change, should only watch resources in the openshift-config and its own namespace.

But I'll do some further testing with the rebased commit once I manage to spawn a cluster with clusterbot

andreaskaris:
Status: I already rebased this but I'm currently waiting on Cluster Bot to spawn a system (and atm, it likes to fail for the clusters that I'm requesting)

andreaskaris:
I pushed the rebased version and ran a quick sanity check in a cluster bot built cluster. I'll run another manual test tomorrow just to be sure

controllers/provisioning_controller.go
zaneb: > the following line, which literally gives the cluster-baremetal-operator the right to do anything with any configmaps, secrets or services in the entire cluster

Unfortunately it has to be a superset of the RBAC roles it is deploying, so this is unavoidable.

> The operator used to only watch resources in its own namespace, and with the change, should only watch resources in the openshift-config and its own namespace.

Thanks, I indeed missed this. That should be sufficient AIUI.

andreaskaris:
I verified this manually once more and updating the secret works with this PR according to my limited testing, of course

andreaskaris:
/hold

there's a nil pointer issue

provisioning/baremetal_secrets.go
```diff
@@ -29,13 +30,16 @@ const (
...
-	openshiftConfigNamespace = "openshift-config"
 	openshiftConfigSecretKey = ".dockerconfigjson" // #nosec
-	pullSecretName           = "pull-secret"
 	// NOTE(dtantsur): this is kept here to be able to remove the old
 	// secret when a Provisioning is removed.
 	ironicrpcSecretName = "metal3-ironic-rpc-password" // #nosec
 	baremetalSecretName = "metal3-mariadb-password"    // #nosec
+
+	// OpenshiftConfigNamespace holds the name of the openshift-config namespace.
+	OpenshiftConfigNamespace = "openshift-config"
```
honza, line 40: nit: let's use a naming scheme consistent with the existing code

provisioning/baremetal_secrets.go
```diff
@@ -118,29 +122,62 @@ password = %s
...
-		return err
+		return fmt.Errorf("could not get secret %s/%s, err: %w", OpenshiftConfigNamespace, PullSecretName, err)
+	}
+	openshiftConfigSecretKeyData, ok := openshiftConfigSecret.Data[openshiftConfigSecretKey]
+	if !ok {
+		return fmt.Errorf("could not find key %q in secret %s/%s", openshiftConfigSecretKey, OpenshiftConfigNamespace, PullSecretName)
+	}
+
+	// Try to get the openshift-machine-api/pull-secret field .dockerconfigjson.
+	// The openshift-machine-api/pull-secret .dockerconfigjson field should be double encoded due to PR
```
honza, line 136: What do we mean by *double* encoded?

andreaskaris: In OCP, all secret data is base64 encoded, e.g.:
```
echo "string" | base64 -w0
```

That happens automatically when you write a string into Secret.StringData:
https://github.com/kubernetes/api/blob/v0.28.1/core/v1/types.go#L6541C3-L6541C3

It's then automatically base64 encoded and can be retrieved in Secret.Data:
https://github.com/kubernetes/api/blob/v0.28.1/core/v1/types.go#L6533

But the content of openshift-machine-api/pull-secret is actually 2x base64 encoded, the bash equivalent is:
```
echo "string" | base64 -w0 | base64 -w0
```

See https://github.com/openshift/cluster-baremetal-operator/pull/184/commits/067b314eebfc48befe4ad684fa1e9da1f91050d3#diff-fe5186bad7cb27f50e9cfd2a5ef8d36067e64477380926b9d6da32eeaa9f08ffR157

Unfortunately, that's existing logic that I didn't want to touch. It would IMO have been way better to store the secret correctly encoded ( = only once base64 encoded) and instead make sure that whatever component is responsible for sending the secret to the Configure-CoreOS init container (or the container itself) takes care of encoding this once more, if needed

provisioning/baremetal_secrets.go
andreaskaris: Thanks for the review!

The existing code used:
```
openshiftConfigNamespace = "openshift-config"
```

and:
```
pullSecretName           = "pull-secret"
```

I needed to make the variables public so that I could use them across packages, and the only way to make a variable in Golang public is by starting them with an upper case letter.

So I'm consistent with the preexisting code. Or what better names do you have in mind for these variables?

andreaskaris:
@honza  @zaneb  I addressed / answered to your respective comments, see above. wdyt about the PR now / any further input? Thanks!

andreaskaris:
@sgoveas That approved didn't work ;-)

PR #354, opened 2023-08-15
OCPBUGS-17693: adjust meta3-state service ports when proxy is on

dtantsur:


dtantsur:
/hold

The virtual media job must pass on this.

PR #357, opened 2023-08-22
[WIP] TLS host validation between BMO and Ironic/Inspector

dtantsur:
Now that BMO uses the metal3-state service for talking to Ironic and
Inspector, we can use the standard OpenShift service CA and turn on
TLS host validation.


dtantsur:
The naive attempt has failed. Will get back to it one day.

PR #374, opened 2023-10-17
OCPBUGS-20563: Add tls-cipher-suites to baremetal-kube-rbac-proxy

iurygregory:
This commit adds the tls-cipher-suites to the baremetal-kube-rbac-proxy


iurygregory:
Just as update, I've tested the PR via cluster-bot, and this solves the problem. Let's see if CI helps 🤞🏽

PR #380, opened 2023-11-10
OCPBUGS-23131,OCPBUGS-23149: fix IRONIC_EXTERNAL_URL_V6

dtantsur:
IRONIC_EXTERNAL_URL_V6 is currently calculated incorrectly:

1) It uses the API VIP, but the Ironic's httpd is not proxied in the
   same way as Ironic itself and thus may not be available there.
2) It uses an external IP even when a provisioning network is enabled
   and VirtualMediaViaExternalNetwork is false.

This change fixes both issues, while also cleaning up the provisioning
utils to ensure all functions are dualstack-aware. The split between
the BMO and Metal3 pods is an implicit requirement of this change.

This change does NOT fix the fact that ICC is not configured correctly
for deploying IPv6-only nodes from a dualstack cluster.

### DIFF
--- a/provisioning/baremetal_config_test.go
+++ b/provisioning/baremetal_config_test.go
@@ -115,6 +115,7 @@ type provisioningBuilder struct {
 }
 
 const testProvisioningIP = "172.30.20.3"
+const testProvisioningIPv6 = "fd2e:6f44:5dd8:b856::2"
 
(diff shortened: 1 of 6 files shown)


dtantsur:

Fixes a regression in IPv6 support on dualstack.

PR #382, opened 2023-11-10
Do not run lint when building CBO or generating manifests

dtantsur:
It wastes a lot of time for both local developers and the CI.
Furthermore, for me personally golangci-lint goes out of memory
("just" 16G laptop), so I cannot use it at all.


honza:
#343

PR #383, opened 2023-11-13
MGMT-16047: update dependencies

filanov:
Update controller-runtime and baremetal-operator to newer versions

this change is needed as part of [MGMT-16047](https://issues.redhat.com//browse/MGMT-16047), in order to update baremetal-operator that uses a newer version of controller-runtime as one of the dependencies this repo need to be updated as well


dtantsur:

This is blocking ACM from doing BMO rebases

PR #388, opened 2023-11-21
OCPBUGS-23504: fixes for deploying V6-only clusters from dualstack hubs

dtantsur:
This change backports all fixes done for [METAL-163](https://issues.redhat.com//browse/METAL-163) to provide the right callback URLs for IPv6-only BMCs from dualstack clusters where IPv4 is the primary stack.

Due to architectural reasons, we're forced to also backport the split of the Metal3 pod into BMO and the rest of Metal3.


dtantsur:
/label backport-risk-assessed
/hold

The pod split does pose certain risk. Before merging, ensure nothing suspicious in the CI jobs. I think we have enough coverage, and the change has been tested in 4.14 quite extensively.

dtantsur:
/close

The CI failures seem unrelated, but this is probably too much of a risk.

PR #389, opened 2023-11-21
OCPBUGS-23291: hack for deploying V6-only clusters from dualstack hubs

dtantsur:
In 4.12, we cannot calculate pod IPs of the Metal3 pod to pass to BMO
since BMO is a part of this same pod. As a 4.12-only hack, pass
the variables to compose the external URL to BMO. Another 4.12-only
hack will build the URL out of them.


dtantsur:
/retest

Timeouts

elfosardo:
/lgtm
failures do not seem to be related to this patch

PR #396, opened 2024-01-15
OCPBUGS-27145: Reduce container privileges

zaneb:
Many of the containers we run have to be run in privileged mode in order to access a host volume - not only so they can write files as root, but also to get an SELinux context that allows it.

However, some containers currently running as privileged do not require this.


zaneb:
Tests are failing because the fix for https://bugzilla.redhat.com/show_bug.cgi?id=1908302 in metal3-static-ip-manager - [toggling the addr_gen mode](https://github.com/openshift/ironic-static-ip-manager/blob/master/refresh-static-ip#L28) - requires extra capabilities.

zaneb:
Metal upgrade jobs were permafailing prior to this by the looks of it.

PR #399, opened 2024-01-24
OCPBUGS-28230: add terminationmessagepolicy to show last of error log in API

deads2k:
will improve visibility for reasons why a pod failed.


deads2k:
we're unable to debug without this information.  I'm manually labeling based on teh slack emoji since it's late on friday.

PR #400, opened 2024-01-25
OCPBUGS-24579: pass all IPs to image-customization-controller

dtantsur:
This is needed to handle the case when single-stack spoke clusters are
deployed from a dual-stack hub.

While here, make sure to pass ports as part of the URLs. CBO is the
authority on port numbers, ICC should not hardcode them.

Requires:
- https://github.com/openshift/image-customization-controller/pull/120


iurygregory:
/retest
We have a build with the PR that this depends on.

PR #401, opened 2024-02-02
OCPBUGS-27145: Drop unneccessary capabilities

zaneb:
Some containers need to run privileged so that we can mount a hostPath volume, but we can at least drop the capabilites we don't need.


zaneb:
@sgoveas [confirms](https://issues.redhat.com/browse/OCPBUGS-27145?focusedId=24606422&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-24606422) that regression tests are now passing with this PR
/retest

PR #403, opened 2024-02-15
OCPBUGS-27760: Update the leader election durations to be tolerant

deads2k:
In OCP, we tolerate about a minute of apiserver downtime before losing a lease by carefully setting the leader election values.  Combined with graceful lease release, this makes the default case fast, with a tolerant controller that doesnt' churn

### DIFF
--- a/main.go
+++ b/main.go
@@ -20,6 +20,7 @@ import (
 	"crypto/tls"
 	"flag"
 	"os"
+	"time"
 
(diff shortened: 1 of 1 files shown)


deads2k:
lease acquisition looks fine: 

> 2024-02-16T05:46:48.995650497Z I0216 05:46:48.995600       1 leaderelection.go:260] successfully acquired lease openshift-machine-api/cluster-baremetal-operator

/retest

PR #406, opened 2024-02-22
METAL-904: Move parts of ICC config to a new secret

dtantsur:
The assisted service needs to get the same configuration to embed IPA
into their images. Right now they embed CBO and ICC code to achieve
that. When we change anything on the CBO side (e.g. add ironic-proxy or
drop inspector), they need to update their code on all supported
releases.

This change adds a new secret with some of the IRONIC_* variables. It is
used for ICC itself and will be used by the assisted service instead of
CBO code directly.

### DIFF
--- a/provisioning/image_customization.go
+++ b/provisioning/image_customization.go
@@ -47,6 +47,7 @@ const (
 	deployISOFile                    = imageSharedDir + "/ironic-python-agent.iso"
 	deployInitrdEnvVar               = "DEPLOY_INITRD"
 	deployInitrdFile                 = imageSharedDir + "/ironic-python-agent.initramfs"
+	imageCustomizationConfigName     = "metal3-image-customization-config"
 )
 
 var (
@@ -242,10 +243,44 @@ func newImageCustomizationDeployment(info *ProvisioningInfo, ironicIPs []string,
 	}
 }
 
+func newImageCustomizationConfig(info *ProvisioningInfo, ironicIPs []string, inspectorIPs []string) *corev1.Secret {
+	return &corev1.Secret{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:        imageCustomizationConfigName,
+			Namespace:   info.Namespace,
+			Annotations: map[string]string{},
+			Labels: map[string]string{
+				"k8s-app":    metal3AppName,
+				cboLabelName: imageCustomizationService,
+			},
+		},
+		Type: corev1.SecretTypeOpaque,
+		StringData: map[string]string{
+			ironicAgentImage: info.Images.IronicAgent,
+			ironicBaseUrl:    getUrlFromIP(ironicIPs, baremetalIronicPort),
+			// TODO(dtantsur): when inspector is gone, we may be able to stop passing this URL
+			ironicInspectorBaseUrl: getUrlFromIP(inspectorIPs, baremetalIronicInspectorPort),
+			sshKeyEnvVar:           info.SSHKey,
+		},
+	}
+}
+
 func EnsureImageCustomizationDeployment(info *ProvisioningInfo) (updated bool, err error) {
 	ironicIPs, inspectorIPs, err := GetIronicIPs(info)
 	if err != nil {
 		return false, fmt.Errorf("unable to determine Ironic's IP to pass to the machine-image-customization-controller: %w", err)
+	}
+
+	configSecret := newImageCustomizationConfig(info, ironicIPs, inspectorIPs)
+	err = controllerutil.SetControllerReference(info.ProvConfig, configSecret, info.Scheme)
+	if err != nil {
+		err = fmt.Errorf("unable to set controllerReference on machine-image-customization config: %w", err)
+		return
+	}
+	_, _, err = resourceapply.ApplySecret(
+		context.Background(), info.Client.CoreV1(), info.EventRecorder, configSecret)
+	if err != nil {
+		return false, fmt.Errorf("failed to create the image customization config: %w", err)
 	}
 
 	imageCustomizationDeployment := newImageCustomizationDeployment(info, ironicIPs, inspectorIPs)
--- a/provisioning/image_customization_test.go
+++ b/provisioning/image_customization_test.go
@@ -64,6 +64,12 @@ func TestNewImageCustomizationContainer(t *testing.T) {
 			pullSecret,
 		},
 	}
+	secret1 := map[string]string{
+		"IRONIC_BASE_URL":           "https://192.168.0.2:6385",
+		"IRONIC_INSPECTOR_BASE_URL": "https://192.168.0.2:5050",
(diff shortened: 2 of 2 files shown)


carbonin:
Were you also going to add the secret name to an annotation somewhere?
I don't recall exactly where that was and by extension if it belongs in this PR.

dtantsur:
> Were you also going to add the secret name to an annotation somewhere?

@carbonin Any suggestions? I could add it to Provisioning if it makes the discovery easier for you (if you don't want to rely on the predefined name, for instance).

carbonin:
It depends on how unlikely you think it is that the name/namespace of this secret is to change.
I suppose we could just as easily depend on the presence of a secret as the presence of an annotation, but I definitely also don't want to have to check multiple locations depending on the OCP release or something.

dtantsur:
@carbonin I don't expect it to ever change, but who knows. I suspect adding it as annotation to Provisioning will result is somewhat more code than I have now. It's also going to be a bit weird because Provisioning is not namespaced, but the secret lives in openshift-machine-api. Would you still find an annotation useful?

carbonin:
Okay, I'm fine with just the secret then.

I suppose we're exposed to just about the same amount of risk either way.
The annotation name could change just as easily as the secret name so :shrug:

provisioning/image_customization.go
```diff
@@ -242,12 +239,46 @@ func newImageCustomizationDeployment(info *ProvisioningInfo, ironicIPs []string,
...
+			Annotations: map[string]string{},
+			Labels: map[string]string{
+				"k8s-app":    metal3AppName,
+				cboLabelName: imageCustomizationService,
+			},
+		},
+		Type: corev1.SecretTypeOpaque,
+		StringData: map[string]string{
+			ironicAgentImage: info.Images.IronicAgent,
+			ironicBaseUrl:    getUrlFromIP(ironicIPs, baremetalIronicPort),
```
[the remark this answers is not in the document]
dtantsur: Starting with https://github.com/openshift/image-customization-controller/commit/99105b7551e6605a3d423727214813e23451a96e (which you'll need to vendor first), this can be a comma-separated list.

provisioning/image_customization.go
```diff
@@ -242,12 +239,46 @@ func newImageCustomizationDeployment(info *ProvisioningInfo, ironicIPs []string,
...
 		return false, fmt.Errorf("unable to determine Ironic's IP to pass to the machine-image-customization-controller: %w", err)
 	}
 
+	configSecret := newImageCustomizationConfig(info, ironicIPs, inspectorIPs)
+	err = controllerutil.SetControllerReference(info.ProvConfig, configSecret, info.Scheme)
+	if err != nil {
+		err = fmt.Errorf("unable to set controllerReference on machine-image-customization config: %w", err)
+		return
+	}
+	_, _, err = resourceapply.ApplySecret(
```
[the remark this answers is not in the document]
dtantsur: Ugh, you're right.

@carbonin I assume it's not a problem for you because you read the information on demand anyway, right? Maybe we don't use the secret ourselves then and only create it for the assisted service to consume.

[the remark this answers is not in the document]
carbonin: Right we'd read the data out of the secret using the API.
And if the data is liable to change then we'd do it every time we need it.

carbonin (approved):

provisioning/image_customization.go
```diff
@@ -242,12 +243,46 @@ func newImageCustomizationDeployment(info *ProvisioningInfo, ironicIPs []string,
...
+		context.Background(), info.Client.CoreV1(), info.EventRecorder, configSecret)
```
zaneb, line 281: Last I checked you'll need to use this wrapper function: https://github.com/openshift/cluster-baremetal-operator/blob/master/provisioning/baremetal_secrets.go#L51-L73 instead of calling straight into library-go, otherwise it ends up writing the Secret on every reconcile.

zaneb: Because it doesn't check if anything is changing before it writes it.

dtantsur: It does though: https://github.com/openshift/library-go/blob/4109a9e7a43790f6bf68db64171b2257ae20d344/pkg/operator/resource/resourceapply/core.go#L423-L426. At least it intends to. Now, reading more the source, the problem could be the opposite: it assumes that the data never changes. I wish they had a usable documentation for these calls...

dtantsur: If I'm reading it right, the wrapper is useless: since ApplySecret only compares metadata, it just exits. Could you double-check? Am I reading it wrong?

dtantsur: This is the problematic part @zaneb: https://github.com/openshift/library-go/blob/4109a9e7a43790f6bf68db64171b2257ae20d344/pkg/operator/resource/resourceapply/core.go#L413. I suspect our only real option is to delete the secret...

zaneb: `.Data` is the content of the Secret, not the metadata, so it should be OK.

It looks like I misremembered the reason for adding our own wrapper, according to #127 it was so we could add custom logic like in https://github.com/openshift/cluster-baremetal-operator/blob/03222f4e09198bb7af9b7b8260fc2b8223f07553/provisioning/baremetal_secrets.go#L234-L240

I was definitely misled by https://github.com/openshift/cluster-baremetal-operator/blob/03222f4e09198bb7af9b7b8260fc2b8223f07553/provisioning/baremetal_secrets.go#L142-L175, which Get()s the Secret no less than 3 times - once before calling anything, again to pass to shallUpdateData() where it is ignored in favour of the first one, and finally in library-go which repeats the same comparison.

zaneb: #409

dtantsur: > existingCopy is the thing that [gets written](https://github.com/openshift/library-go/blob/4109a9e7a43790f6bf68db64171b2257ae20d344/pkg/operator/resource/resourceapply/core.go#L439) at the end, that's why the data is copied into it.

Yeah, but only [if the metadata is different](https://github.com/openshift/library-go/blob/4109a9e7a43790f6bf68db64171b2257ae20d344/pkg/operator/resource/resourceapply/core.go#L423)?

zaneb: That's comparing the whole object, not just the metadata.
It fetches the object from the API, makes a copy, merges the new data that we're requiring into the copy, then compares the two to see if they're different.

derekhiggins:
/retest-required
the way this job is failing isn't consistent, lets try again

dtantsur:
/test e2e-metal-ipi-ovn-ipv6

Fails radically different every time, this time without even getting to testing anything.

dtantsur:
/test e2e-metal-ipi-ovn-ipv6

Ansible Galaxy unavailable this time.

dtantsur:
/test e2e-metal-ipi-ovn-ipv6

Something in kubeapi this time...

PR #408, opened 2024-02-29
OCPBUGS-29687: Move Secrets from environment variables to VolumeMounts in metal3-httpd container

MahnoorAsghar:


elfosardo (commented):
just wondering it we want to adapt the test instead of just removing the secrets from the environment variables

MahnoorAsghar:
> just wondering it we want to adapt the test instead of just removing the secrets from the environment variables
I should adapt the test actually, will do!

dtantsur:
/hold

I don't believe this actually works. ironic-image expects the secrets to be mounted at `/auth/ironic/htpassd` (https://github.com/openshift/ironic-image/blob/master/scripts/auth-common.sh#L16-L18) while you're mounting them at the target file in `/etc`. Which is potentially confusing: you end up with `IRONIC_HTPASSWD` empty (confirmed via logs). The risk here is that something may rely on `IRONIC_HTPASSWD` being set instead of the file being populated, and "something" actually does: https://github.com/openshift/ironic-image/blob/master/ironic-config/httpd-ironic-api.conf.j2#L45. I suspect you end up without authentication.

dtantsur:
Looks plausible, let's see what the CI says.

MahnoorAsghar:
The metal3-httpd container is failing with this message:
`container create failed: time="2024-06-28T11:31:04Z" level=error msg="runc create failed: unable to start container process: error during container init: error mounting \"/var/lib/kubelet/pods/216ac305-72f8-44e3-849a-b88110938e0b/volume-subpaths/ironic-htpasswd/metal3-httpd/6\" to rootfs at \"/auth/ironic/htpasswd\": mount /var/lib/kubelet/pods/216ac305-72f8-44e3-849a-b88110938e0b/volume-subpaths/ironic-htpasswd/metal3-httpd/6:/auth/ironic/htpasswd (via /proc/self/fd/6), flags: 0x5001, data: context=\"system_u:object_r:container_file_t:s0:c189,c652\": not a directory"`

MahnoorAsghar:
The problem appears to come from the fact that the `metal3-ironic-basic-auth` volumeMount also uses the `/auth/ironic` directory, which is causing a conflict with the `ironic-htpasswd` volumeMount. 
I understand that I have two options: using a projected volume, or changing the htpasswd mount path to something different (which would require changes to ironic-image and BMO again).

https://kubernetes.io/docs/tasks/configure-pod-container/configure-projected-volume-storage/#configure-a-projected-volume-for-a-pod

dtantsur:
I don't think we should be changing upstream again. Realistically, we should have just one secret that contains both username/password and htpasswd (this is what I do in ironic-standalone-operator). Do you think it's going to be hard to change (taking into account upgrades)?

MahnoorAsghar:
From a quick look it appears that CBO itself is creating these secrets in provisioning/baremetal_secrets.go. It should be possible to use one Secret for the ironic (and inspector) username, password, and htpasswd, then. 
This should be backport-able too, but I am not sure if it will break any existing deployments.

MahnoorAsghar:
I did not use the inspectorCredentialsVolume because it is mounted at /auth/ironic-inspector, but the inspector htpasswd is mounted at /auth/inspector.

dtantsur:
> I did not use the inspectorCredentialsVolume because it is mounted at /auth/ironic-inspector, but the inspector htpasswd is mounted at /auth/inspector.

Sigh. Good that we're removing the inspector support anyway.

dtantsur:
/lgtm cancel

unable to apply Metal3 deployment: Deployment.apps \"metal3\" is invalid: [spec.template.spec.volumes[9].secret.items[0].path: Invalid value: \"/run/secrets/pull-secret\": must be a relative path, spec.template.spec.volumes[10].secret.items[0].path: Invalid value: \"/etc/ironic/htpasswd\": must be a relative path, spec.template.spec.volumes[11].secret.items[0].path: Invalid value: \"/etc/ironic-inspector/htpasswd\": must be a relative path, spec.template.spec.containers[0].volumeMounts[6].name: Not found: \"ironic-htpasswd\", spec.template.spec.containers[0].volumeMounts[7].name: Not found: \"inspector-htpasswd\"

Not sure what changed and why, but this does not work. A conflict with the other change of yours? Could you rebase and look into these issues?

dtantsur:
Hold on, it may be wrong reporting? Which jobs should I be looking at?

dtantsur:
Overriding jobs that no longer exist:

/override ci/prow/e2e-metal-ipi-sdn-serial-ipv4
/override ci/prow/e2e-metal-ipi-sdn-bm

Re-adding lgtm since the actual job is not suspicious:

/lgtm

PR #409, opened 2024-03-25
OCPBUGS-31341: Refactor Pull Secret application

zaneb:


elfosardo:
this looks ok to me
I don't think the failures we're seeing depend on the change, but I will give the metal jobs one more chance

zaneb:
Confirmed from the CI logs that the Secret is not churning (CreationTimestamp and update time are the same).

Not sure what is happening with the other CI jobs, but it appears unrelated.
/retest-required

PR #410, opened 2024-04-11
OCPBUGS-31961: bump x/net to 0.23.0

iurygregory:
Steps used:
$ go get golang.org/x/net@v0.23.0
$ go mod tidy
$ go mod vendor


flavianmissi:
Hi @iurygregory, it could be that the exposed http/2 server uses the built-in net/http instead (even though x/net is vendored). 
If that is the case, the problem should already have been fixed by new golang builds we got from ART.

iurygregory:
@flavianmissi  tks, I'm going to update this since it has conflicts, but I think the latest test done by Jad was working o/

iurygregory:
Closing this one, since it was pushed to master and we already have  x/net 0.25.0, I will open a new PR directly to release-4.16, and ask QE to verify 4.17 works.

PR #415, opened 2024-05-03
OCPBUGS-33254: Wait for bootstrapping to finish before starting

honza:


controllers/provisioning_controller.go
```diff
@@ -250,6 +250,10 @@ func (r *ProvisioningReconciler) Reconcile(ctx context.Context, req ctrl.Request
 			"unable to put %q ClusterOperator in Available state", clusterOperatorName)
 	}
 
+	if _, ok := baremetalConfig.Annotations["provisioning.metal3.io/paused"]; ok {
+		return ctrl.Result{}, fmt.Errorf("provisioning CR paused")
```
dtantsur, line 254: Maybe it's a case for RequeueAfter rather than an error? Especially since we don't want an immediate requeue.

controllers/provisioning_controller.go
```diff
@@ -250,6 +250,10 @@ func (r *ProvisioningReconciler) Reconcile(ctx context.Context, req ctrl.Request
 			"unable to put %q ClusterOperator in Available state", clusterOperatorName)
 	}
 
+	if _, ok := baremetalConfig.Annotations["provisioning.metal3.io/paused"]; ok {
+		return ctrl.Result{RequeueAfter: time.Minute}, fmt.Errorf("provisioning CR paused")
```
dtantsur, line 254: nit: `fmt.Errorf` is an overkill, `errors.New` is enough

dtantsur:
Looks good whenever ready.

controllers/provisioning_controller.go
honza: Done

controllers/provisioning_controller.go
honza: Done

PR #417, opened 2024-05-15
OCPBUGS-32304: Run metal3 containers as specific users and groups

MahnoorAsghar:
Run metal3 containers as specific users and groups. This will fix permission issues with the ironic-inspector container that it gets when it tries to create a directory, resulting in its error state.


MahnoorAsghar:
This needs to go into 4.16.0 as the bug it references is a release blocker.

PR #421, opened 2024-05-28
OCPBUGS-34119: Upgrade controller-tools and library-go

honza:


go.mod
```diff
@@ -1,147 +1,169 @@
 module github.com/openshift/cluster-baremetal-operator
 
-go 1.21
+go 1.22.0
+
+toolchain go1.22.3
```
dtantsur, line 5: This needs to be dropped

honza: Done

PR #422, opened 2024-05-30
OCPBUGS-34649: provisioning: add ownership for tls secret

vrutkovs:
Make sure all TLS artifacts have clear ownership

### DIFF
--- a/provisioning/baremetal_secrets.go
+++ b/provisioning/baremetal_secrets.go
@@ -15,6 +15,7 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/client"
 	"sigs.k8s.io/controller-runtime/pkg/controller/controllerutil"
 
+	"github.com/openshift/api/annotations"
 	"github.com/openshift/library-go/pkg/operator/events"
 	"github.com/openshift/library-go/pkg/operator/resource/resourceapply"
 )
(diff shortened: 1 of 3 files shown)


provisioning/baremetal_secrets.go
```diff
@@ -194,6 +195,9 @@ func createOrUpdateTlsSecret(info *ProvisioningInfo) error {
 		ObjectMeta: metav1.ObjectMeta{
 			Name:      tlsSecretName,
 			Namespace: info.Namespace,
+			Annotations: map[string]string{
+				annotations.OpenShiftComponent: "Bare Metal Hardware Provisioning / baremetal-operator ",
```
honza, line 199: nit: is the trailing space intentional?

provisioning/baremetal_secrets.go
```diff
@@ -29,6 +30,7 @@ const (
 	inspectorSecretName      = "metal3-ironic-inspector-password"
 	inspectorUsername        = "inspector-user"
 	tlsSecretName            = "metal3-ironic-tls" // #nosec
+	baremetalJiraComponent   = "Bare Metal Hardware Provisioning / baremetal-operator"
```
dtantsur, line 33: I think we have a separate component for cluster-baremetal-operator

vrutkovs: not sure what's the difference between the two, but sure, lets make it cluster-baremetal-operator

dtantsur: Two unfortunately named and related but still quite different projects.

PR #423, opened 2024-06-05
METAL-1040: Allow CBO to run without MAPI

eranco74:
Skip updateProvisioningMacAddresses in case the Machine CRD isn't registered
Remove Machine from the list of objects the provisioning controller manage

### DIFF
--- a/controllers/provisioning_controller.go
+++ b/controllers/provisioning_controller.go
@@ -31,6 +31,7 @@ import (
 	appsv1 "k8s.io/api/apps/v1"
 	corev1 "k8s.io/api/core/v1"
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
+	"k8s.io/apimachinery/pkg/api/meta"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/labels"
 	"k8s.io/apimachinery/pkg/runtime"
@@ -560,8 +561,14 @@ func (r *ProvisioningReconciler) updateProvisioningMacAddresses(ctx context.Cont
 	machines := machinev1beta1.MachineList{}
 	bmhNames := []string{}
 	labelReq, _ := labels.NewRequirement("machine.openshift.io/cluster-api-machine-role", selection.Equals, []string{"master"})
-	if err := r.Client.List(ctx, &machines, &client.ListOptions{LabelSelector: labels.NewSelector().Add(*labelReq)}); err != nil {
-		return errors.Wrap(err, "cannot list master machines")
+	err := r.Client.List(ctx, &machines, &client.ListOptions{LabelSelector: labels.NewSelector().Add(*labelReq)})
+	if err != nil {
+		if runtime.IsNotRegisteredError(err) || meta.IsNoMatchError(err) {
+			klog.Info("Machines CRD is not registered in the cluster, set provisioningMacAddresses if the metal3 pod fails to start")
+			return nil
+		} else {
+			return errors.Wrap(err, "cannot list master machines")
+		}
 	}
 	if len(machines.Items) < 1 {
 		klog.Info("No Machines with cluster-api-machine-role=master found, set provisioningMacAddresses if the metal3 pod fails to start")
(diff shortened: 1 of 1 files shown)


controllers/provisioning_controller.go
```diff
@@ -655,7 +662,6 @@ func (r *ProvisioningReconciler) SetupWithManager(mgr ctrl.Manager) error {
 		Owns(&appsv1.DaemonSet{}).
 		Owns(&osconfigv1.ClusterOperator{}).
 		Owns(&osconfigv1.Proxy{}).
-		Owns(&machinev1beta1.Machine{}).
```
zaneb, line 658: I agree that this was definitely having no useful effect in terms of triggering reconciles, since the Provisioning CR (thankfully) never owns any Machines.

The correct thing here I think would be to conditionally add `Watches(&machinev1beta1.Machine{}, handler.EnqueueRequestsFromMapFunc(watchOCPConfigPullSecret)` (note the map function could use a better name. Also the next line is [wrong as well](https://github.com/openshift/cluster-baremetal-operator/pull/351/files#r1628776168) I think.)

If removing this also has the effect of no longer caching Machines (I can't remember if this is the case, but it seems likely) then we won't actually ever be able to find the Machine when we look it up (even on hosts where the CRD exists), assuming we use the cached client.

zaneb: In the log I see:

```
2024-06-05T15:23:12.873346427Z W0605 15:23:12.873314       1 provisioning_controller.go:521] Machine cluster13-wzd5l-master-1 has no annotation linking it to BareMetalHost
2024-06-05T15:23:12.873392927Z W0605 15:23:12.873383       1 provisioning_controller.go:521] Machine cluster13-wzd5l-master-2 has no annotation linking it to BareMetalHost
2024-06-05T15:23:12.873420905Z W0605 15:23:12.873412       1 provisioning_controller.go:521] Machine cluster13-wzd5l-master-0 has no annotation linking it to BareMetalHost
2024-06-05T15:23:12.974889888Z I0605 15:23:12.974839       1 provisioning_controller.go:601] "Updating provisioning MAC addresses" CurrentMacAddresses=null NewMacAddresses=["F8:F2:1E:B2:DB:51","F8:F2:1E:B2:E7:81","F8:F2:1E:B2:E9:D1"]
```

so I guess this is not preventing us reading the Machines at least.

eranco74: Just to make sure, we want to have the Machine in the client cache (if the Machine CR is registered) we don't want to trigger reconcile, right?

controllers/provisioning_controller.go
```diff
@@ -646,17 +656,47 @@ func (r *ProvisioningReconciler) SetupWithManager(mgr ctrl.Manager) error {
...
+		Watches(&corev1.Secret{}, handler.EnqueueRequestsFromMapFunc(watchOCPConfigPullSecret), builder.WithPredicates(secretFilter))
+
+	// Check if Machine CRD is registered
+	registered, err := isMachineRegistered(mgr)
+	if err != nil {
+		return err
+	}
+
+	if registered {
+		ctrlBuilder = ctrlBuilder.Watches(&machinev1beta1.Machine{}, handler.EnqueueRequestsFromMapFunc(skipReconcile))
```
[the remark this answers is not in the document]
eranco74: See https://github.com/openshift/cluster-baremetal-operator/pull/423#discussion_r1628777876
Provisioning CR never owns any Machines, we have this just to keep Machines in the cached client.

controllers/provisioning_controller.go
JoelSpeed: Provided you are using the standard client that controller runtime provides, the first time an object type is listed or fetched, the controller will start to cache objects of that kind across the cluster.

There are options you can use to configure this behaviour, such as excluding certain types from being cached or restricting caching to all namespaces, best to check you aren't using any of those presently, but, the `Owns` here won't affect whether your machine objects are cached or not

dtantsur:
/retitle METAL-1040: Allow CBO to run without MAPI

PR #424, opened 2024-06-11
OCPBUGS-35211: Add capability to metal3-ramdisk-logs container

MahnoorAsghar:
The /shared directory is mounted to the ironic, ironic-inspector, ramdisk-logs, dnsmasq, and httpd containers. As the ironic container creates the /shared directory, the ramdisk-logs container was unable to delete files from the directory as it was running as the ironic user and group, but the ironic container (which created and owned the files) was running in privileged mode. 

To resolve this issue, the metal3-ramdisk-logs container is now run with the privileged SecurityContext.

### DIFF
--- a/provisioning/baremetal_pod.go
+++ b/provisioning/baremetal_pod.go
@@ -74,7 +74,6 @@ const (
 )
 
 var ironicUserID int64 = 1002
-var ironicGroupID int64 = 1003
 var inspectorGroupID int64 = 1004
 
 var podTemplateAnnotations = map[string]string{
@@ -700,10 +699,11 @@ func createContainerMetal3RamdiskLogs(images *Images) corev1.Container {
 		},
 		TerminationMessagePolicy: corev1.TerminationMessageFallbackToLogsOnError,
 		SecurityContext: &corev1.SecurityContext{
-			RunAsUser:  ptr.To(ironicUserID),
-			RunAsGroup: ptr.To(ironicGroupID),
 			Capabilities: &corev1.Capabilities{
 				Drop: []corev1.Capability{"ALL"},
+				Add: []corev1.Capability{
+					"CAP_DAC_OVERRIDE",
+				},
 			},
 		},
 	}


provisioning/baremetal_pod.go
```diff
@@ -782,6 +792,8 @@ func createContainerMetal3StaticIpManager(images *Images, config *metal3iov1alph
 					"FOWNER", // Needed for setting the addr_gen_mode
 				},
 			},
+			RunAsUser:  ptr.To(ironicUserID),
+			RunAsGroup: ptr.To(ironicGroupID),
```
dtantsur, line 796: This container is marked as privileged, we shouldn't/cannot use specific users with it.

provisioning/baremetal_pod.go
```diff
@@ -559,6 +565,8 @@ func createContainerMetal3Httpd(images *Images, config *metal3iov1alpha1.Provisi
 			Capabilities: &corev1.Capabilities{
 				Drop: []corev1.Capability{"ALL"},
 			},
+			RunAsUser:  ptr.To(ironicUserID),
+			RunAsGroup: ptr.To(ironicGroupID),
```
dtantsur, line 569: This container is marked as privileged, we shouldn't/cannot use specific users with it.

dtantsur:
I've started leaving comments on containers that are privileged, and realized that most of the remaining actually are. I'm afraid, we need to go down a less ideal approach and make the ramdisk-logs container privileged again.

provisioning/baremetal_pod.go
```diff
@@ -700,8 +699,7 @@ func createContainerMetal3RamdiskLogs(images *Images) corev1.Container {
 		},
 		TerminationMessagePolicy: corev1.TerminationMessageFallbackToLogsOnError,
 		SecurityContext: &corev1.SecurityContext{
-			RunAsUser:  ptr.To(ironicUserID),
-			RunAsGroup: ptr.To(ironicGroupID),
+			Privileged: pointer.BoolPtr(true),
```
zaneb, line 702: Have you tried adding CAP_FOWNER instead?
As far as I know, privileged mode should only be needed to tell SELinux to [allow us to use hostpath volumes](https://docs.openshift.com/container-platform/4.15/storage/persistent_storage/persistent_storage_local/persistent-storage-hostpath.html#persistent-storage-hostpath-pod_persistent-storage-hostpath), (which the shared volume is not an example of). In theory I believe everything else should be controllable by capabilities.

provisioning/baremetal_pod.go
zaneb: Is this actually a rule? (I don't know, I'm still trying to figure this stuff out.)

I don't know that we can rely on this to solve the problem, since we'll have to upgrade from previous releases where a bunch of files were created by root, but in principle running as the ironic user seems like a good thing to do.

provisioning/baremetal_pod.go
dtantsur: All for trying different things, but let's limit the experiments to a reasonable timeframe since this is a very high priority issue.

dtantsur: And if so, shouldn't it be `CAP_DAC_OVERRIDE` instead?

MahnoorAsghar: It appears CAP_DAC_OVERRIDE should cover it, and CAP_FOWNER is a level above that, which is not needed.

MahnoorAsghar:
Some metal jobs are broken, as is being tracked here: https://issues.redhat.com/browse/OCPBUGS-35564
Fix is in progress.

PR #428, opened 2024-06-19
OCPBUGS-29687: Move pull secret from env var to VolumeMount in metal3-image-customization container

MahnoorAsghar:


MahnoorAsghar:
Previous failures seemed to be because of the etcd container not being available

PR #429, opened 2024-06-20
OCPBUGS-29660: Ignore provisioning ip config if disabled

derekhiggins:
getProvisioningIPCIDR now returns nil if ProvisioningNetwork is disabled. The external HostIP will then be used (as intended when the prov network is disabled).

### DIFF
--- a/provisioning/baremetal_config.go
+++ b/provisioning/baremetal_config.go
@@ -61,6 +61,9 @@ func getDHCPRange(config *metal3iov1alpha1.ProvisioningSpec) *string {
 }
 
 func getProvisioningIPCIDR(config *metal3iov1alpha1.ProvisioningSpec) *string {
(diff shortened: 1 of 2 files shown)


derekhiggins:
Just want to double check this, the PR will mean we ignore any provisioning IP when the prov network is disabled,
and instead the Host IP is used.

Is this the correct behaviour? Could the user be potentially be providing a Provisioning IP from this situation mentioned in the readme?

> when the provisioning network is fully disabled. User can bring up the baremetal cluster using virtual media or assisted installation. If using metal3 for power management, BMCs must be accessible from the machine networks. User should provide two IPs on the external network that would be used for provisioning services.

derekhiggins:
/hold
waiting on clarification that this is ok to merge

zaneb:
This seems as plausible as anything.

derekhiggins:
thanks, unholding
/unhold

PR #430, opened 2024-06-21
METAL-503: Remove ironic-inspector

MahnoorAsghar:


MahnoorAsghar:
The e2e-metal-ipi-upgrade-ovn-ipv6 job failing consistently is a known issue, see https://issues.redhat.com/browse/METAL-1031

config/cluster-baremetal-operator/kustomization.yaml
```diff
@@ -1,7 +1,13 @@
 commonAnnotations:
+  capability.openshift.io/name: baremetal
   include.release.openshift.io/self-managed-high-availability: "true"
   include.release.openshift.io/single-node-developer: "true"
-  capability.openshift.io/name: baremetal
 
 resources:
 - cluster-baremetal-operator.yaml
+apiVersion: kustomize.config.k8s.io/v1beta1
```
dtantsur, line 8: I wonder why this gets added. Will the CI pass if you drop the changes to this file?

provisioning/baremetal_config.go
```diff
@@ -79,33 +77,23 @@ func getDeployKernelUrl() *string {
 	return &deployKernelUrl
 }
 
-// TODO(dtantsur): these two can be removed once we no longer have ironic/inspector split
-
 func getIronicEndpoint() *string {
```
dtantsur, line 80: My comment above claims that this function can be removed to. I wonder if IRONIC_ENDPOINT is actually only used by the removed inspector.

provisioning/baremetal_secrets.go
```diff
@@ -159,10 +157,6 @@ func EnsureAllSecrets(info *ProvisioningInfo) (bool, error) {
 	if err := createIronicSecret(info, ironicSecretName, ironicUsername, "ironic"); err != nil {
 		return false, errors.Wrap(err, "failed to create Ironic password")
 	}
-	// Create a Secret for the Ironic Inspector Password
-	if err := createIronicSecret(info, inspectorSecretName, inspectorUsername, "inspector"); err != nil {
```
dtantsur, line 163: I wonder if we should delete the existing secret.

provisioning/baremetal_config.go
dtantsur: Hmm, the variable itself is probably also used by BMO, but I highly doubt this function is used there - BMO cannot access Ironic over localhost.

config/cluster-baremetal-operator/kustomization.yaml
MahnoorAsghar: Ah yes, changes to this file must be because I ran some make command locally. I think they should be dropped, will do and retest.

provisioning/baremetal_secrets.go
MahnoorAsghar: Good point, have deleted it at the end of this function now

provisioning/baremetal_config.go
MahnoorAsghar: You're right, this function is not being used anywhere. I have removed it now.

dtantsur:
/approve
/hold

The change looks good. Holding it to let your htpasswd fix merge first. Feel free to unhold once ready.

dtantsur:
/hold cancel

Nice job. My approval still holds.

PR #441, opened 2024-08-30
OCPBUGS-30919: SCC-pinning for metal3-baremetal-operator

kramaranya:
[OCPBUGS-30919](https://issues.redhat.com/browse/OCPBUGS-30919)

Pin `metal3-baremetal-operator` pod to use the `hostnetwork-v2` SCC. This change resolves an issue where the operator pod was being assigned an incorrect SCC, causing it to get stuck in a CreateContainerConfigError state. 

Note for review: verify that the pod has the required SCC annotation by checking for the `openshift.io/required-scc: hostnetwork-v2` in the pod's metadata.

### DIFF
--- a/provisioning/bmo_pod.go
+++ b/provisioning/bmo_pod.go
@@ -209,9 +209,16 @@ func newBMOPodTemplateSpec(info *ProvisioningInfo, labels *map[string]string) (*
 
 	containers := injectProxyAndCA([]corev1.Container{container}, info.Proxy)
 
+	podAnnotations := make(map[string]string)
+	for key, val := range podTemplateAnnotations {
+		podAnnotations[key] = val
+	}
+
+	podAnnotations["openshift.io/required-scc"] = "hostnetwork-v2"
+
 	return &corev1.PodTemplateSpec{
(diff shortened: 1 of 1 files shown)


provisioning/bmo_pod.go
```diff
@@ -209,9 +209,14 @@ func newBMOPodTemplateSpec(info *ProvisioningInfo, labels *map[string]string) (*
 
 	containers := injectProxyAndCA([]corev1.Container{container}, info.Proxy)
 
+	podAnnotations := map[string]string{
```
dtantsur, line 212: Could you copy over `podTemplateAnnotations` and update the copy with the additional annotation? This way we won't forget to update BMO if podTemplateAnnotations ever gets updated.

derekhiggins:
/lgtm

```
            "kind": "Pod",
            "metadata": {
                "annotations": {
                    ...
                    "openshift.io/required-scc": "hostnetwork-v2",
                    "openshift.io/scc": "hostnetwork-v2",
                    "seccomp.security.alpha.kubernetes.io/pod": "runtime/default"
                },
                "name": "metal3-baremetal-operator-758b555788-6xfxn",
```

PR #450, opened 2024-10-15
OCPBUGS-43371: allow all features on platform=none

dtantsur:
These platforms are supposed to be highly configurable.


dtantsur:
/hold
/cc @zaneb 

Pending discussion.

PR #435, opened 2024-07-11
OCPBUGS-36357: Revert "reconcile on clusteroperator changes"

elfosardo:
This reverts commit d4e709bbfbae6d316f2e76bec18b0e10a45ac93e.

PR #452, opened 2024-10-18
OCPBUGS-36357: Replace watches with actual Watches

elfosardo:
Replace the watches for CO and Proxy with actual Watches
Remove watches for machines for the time being
Readd watches for secrets

Readd the function to allow triggering reconciliation of
the provisioning CR as mapToProvisioningSingleton.

### DIFF
--- a/controllers/provisioning_controller.go
+++ b/controllers/provisioning_controller.go
@@ -42,8 +42,12 @@ import (
 	"k8s.io/klog/v2"
 	utilnet "k8s.io/utils/net"
 	ctrl "sigs.k8s.io/controller-runtime"
+	"sigs.k8s.io/controller-runtime/pkg/builder"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 	"sigs.k8s.io/controller-runtime/pkg/controller/controllerutil"
+	"sigs.k8s.io/controller-runtime/pkg/handler"
+	"sigs.k8s.io/controller-runtime/pkg/predicate"
+	"sigs.k8s.io/controller-runtime/pkg/reconcile"
 
 	baremetalv1alpha1 "github.com/metal3-io/baremetal-operator/apis/metal3.io/v1alpha1"
 	osconfigv1 "github.com/openshift/api/config/v1"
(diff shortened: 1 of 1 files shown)


controllers/provisioning_controller.go
```diff
@@ -650,14 +654,45 @@ func (r *ProvisioningReconciler) SetupWithManager(mgr ctrl.Manager) error {
...
+		if object.GetName() == metal3iov1alpha1.ProvisioningSingletonName {
+			return true
+		}
+		klog.Info("ignoring provisioning config with unexpected name", "name", object.GetName(), "expected-name", metal3iov1alpha1.ProvisioningSingletonName)
+		return false
+	})
+
+	// Watch Secret openshift-config/pull-secret. If this secret changes, we must requeue the provisioning Singleton,
+	// if it exists.
+	watchOCPConfigPullSecret := func(ctx context.Context, object client.Object) []reconcile.Request {
```
dtantsur, line 667: nit: this variable is confusingly named: it's not used just for the pull secret

dtantsur:
Looks good to me, but I'd like like @zaneb to have a look.

controllers/provisioning_controller.go
zaneb: Agree, and it doesn't do a watch either.
We should call it something like `mapToProvisioningSingleton`.

controllers/provisioning_controller.go
```diff
@@ -650,14 +654,45 @@ func (r *ProvisioningReconciler) SetupWithManager(mgr ctrl.Manager) error {
...
+		return []reconcile.Request{
+			{
+				NamespacedName: types.NamespacedName{
+					Namespace: prov.Namespace,
+					Name:      prov.Name,
+				},
+			},
+		}
+	}
+	secretFilter := predicate.NewPredicateFuncs(func(object client.Object) bool {
```
zaneb, line 682: This could be called something like `pullSecretFilter`

controllers/provisioning_controller.go
```diff
@@ -650,14 +654,45 @@ func (r *ProvisioningReconciler) SetupWithManager(mgr ctrl.Manager) error {
...
+		return object.GetNamespace() == provisioning.OpenshiftConfigNamespace &&
+			object.GetName() == provisioning.PullSecretName
+	})
+
 	return ctrl.NewControllerManagedBy(mgr).
-		For(&metal3iov1alpha1.Provisioning{}).
+		For(&metal3iov1alpha1.Provisioning{}, builder.WithPredicates(provisioningFilter)).
```
zaneb, line 688: This is not needed.

It was added in https://github.com/openshift/cluster-baremetal-operator/pull/351/commits/d4e709bbfbae6d316f2e76bec18b0e10a45ac93e to cancel out the effect of using `handler.EnqueueRequestForObject{}` (instead of `handler.EnqueueRequestFromMapFunc`) on resources with different names to the provisioning CR. We don't need to re-add it.

zaneb (commented):
Some cosmetic suggestions, but this looks correct.

Note that it is essential to backport this to 4.18, because #435 reverted both d4e709bbfbae6d316f2e76bec18b0e10a45ac93e (which did nothing useful) and the operative part of 80f5e6fe41f82ae06c607823f09d7d26cd447868 (which is an essential bug fix for [OCPBUGS-16919](https://issues.redhat.com//browse/OCPBUGS-16919)).

zaneb:
This looks great now, so much more readable that what we used to have. And quite possibly correct as well.
/lgtm

zaneb:
I just realised that #435 also reverted part of #423. We need to either not watch Machines or watch them conditionally only after determining that the CRD exists.

/lgtm cancel

elfosardo:
let's wait for https://github.com/openshift/cluster-baremetal-operator/pull/457 to merge

elfosardo:
@zaneb when you have a moment this should be ready to go now

PR #460, opened 2025-01-05
OCPEDGE-1607: feat: add arbiter node selector

eggfoobar:
Per Arbiter EP: https://github.com/openshift/enhancements/pull/1674

Adding a label or check for arbiter as well as master when the provisioning controller is looking to update the mac address for the machines.


elfosardo:
/lgtm
failures are really not related to this change

PR #461, opened 2025-02-25
[4.19] OCPBUGS-51275: add missing relatedObjects

iurygregory:
This commit adds the missing relatedObjects to CBO


honza:
Probably needs an update here, too https://github.com/openshift/cluster-baremetal-operator/blob/main/controllers/clusteroperator.go#L75

PR #464, opened 2025-03-04
METAL-1404: Add annotation to enable CBO install via CVO in Hosted Clusters

MahnoorAsghar:
Add annotation to enable CBO install via CVO in Hosted Clusters


MahnoorAsghar:
/hold
Until the ability to make capabilities optional is introduced in hypershift, so that the installation of CBO is tied to the baremetal capability being enabled, and this capability can be made optional and non-default.
https://github.com/openshift/hypershift/pull/5456 is relevant in this regard.

MahnoorAsghar:
This installed CBO on the hosted cluster worker nodes, not the hosted cluster control plane, which is the desired effect.

PR #466, opened 2025-03-18
OCPBUGS-52483: Allow provisioningIP when network disabled

honza:
In 96b70d809bb608194407ebea334979a3e91a5a74, we decided to ignore the ProvisioningIP setting when the provisioning network is disabled.  It turns out that this has unintended consequences in specific scenarios.

Here, we are relaxing that requirement.  We pass the ProvisioningIP value to the metal3 containers.  The original bug was related to DHCP causing problems so we disable DHCP when the provisioning network is disabled.

UPDATE: We also take the opportunity to refactor the config retrieval logic around ProvisioningIP. We rename some functions to make things clearer and we no longer require network CIDR when provisioning IP is specified.


provisioning/baremetal_config.go
```diff
@@ -49,6 +49,9 @@ var (
 )
 
 func getDHCPRange(config *metal3iov1alpha1.ProvisioningSpec) *string {
+	if config.ProvisioningNetwork == metal3iov1alpha1.ProvisioningNetworkDisabled {
```
dtantsur, line 52: I think it should be != Managed because we also don't start DHCP in the Unmanaged case

provisioning/baremetal_config.go
```diff
@@ -61,9 +64,6 @@ func getDHCPRange(config *metal3iov1alpha1.ProvisioningSpec) *string {
 }
 
 func getProvisioningIPCIDR(config *metal3iov1alpha1.ProvisioningSpec) *string {
-	if config.ProvisioningNetwork == metal3iov1alpha1.ProvisioningNetworkDisabled {
-		return nil
-	}
 	if config.ProvisioningNetworkCIDR != "" && config.ProvisioningIP != "" {
```
dtantsur, line 67: I wonder if we need to split the logic around provisioning CIDR and provisioning IP. Despite very similar name, the latter is generically useful while the former is DHCP-specific.

Shameless plug: when writing IrSO, I actually split the two option groups entirely: https://github.com/metal3-io/ironic-standalone-operator/blob/main/api/v1alpha1/ironic_types.go#L61-L161

provisioning/baremetal_pod_test.go
```diff
@@ -357,13 +357,11 @@ func TestNewMetal3Containers(t *testing.T) {
 				withEnv(
 					containers["metal3-httpd"],
 					envWithValue("PROVISIONING_INTERFACE", ""),
-					envWithFieldValue("PROVISIONING_IP", "status.hostIP"),
```
dtantsur, line 360: I'd like to see one more test case where the provisioning IP is not set, and we fall back to status.hostIP

honza: Isn't that the test just below, L374? `DisabledSpecWithoutProvisioningIP`?

provisioning/baremetal_config.go
honza: Done

honza:
```
2025-05-20T14:56:12.785478251Z 2025-05-20 14:56:12.784 1 ERROR ironic.conductor.task_manager [None req-01e32944-c0b3-454d-8c87-5c662f574d29 - - - - - -] Node c86fc930-9b12-472b-9f1c-d8277e81831f moved to provision state "inspect failed" from state "inspecting"; target provision state is "manageable": ironic.common.exception.HardwareInspectionFailure: Failed to inspect hardware. Reason: unable to start inspection: Unable to connect to http://[fd2e:6f44:5dd8:c956::1]:8000/redfish/v1/Systems/fbcc7c34-ef06-48a4-962e-e15631765c50/VirtualMedia/Cd/Actions/VirtualMedia.InsertMedia. Error: HTTPConnectionPool(host='fd2e:6f44:5dd8:c956::1', port=8000): Read timed out. (read timeout=60)
```

.gitignore
```diff
@@ -10,9 +10,6 @@ bin
 # Test binary, build with `go test -c`
 *.test
 
-# Output of the go coverage tool, specifically when used with LiteIDE
-*.out
-
```
dtantsur, line 15: Unrelated (and potentially wrong?) change

honza: The `verify-deps` job was [previously complaining about this file](https://prow.ci.openshift.org/view/gs/test-platform-results/pr-logs/pull/openshift_cluster-baremetal-operator/466/pull-ci-openshift-cluster-baremetal-operator-main-verify-deps/1923328959299194880).

dtantsur: Oh lovely... Can we bump go-errors to include https://github.com/go-errors/errors/commit/7db060dd91303e294f1ee2855024347369f42c11?

honza: Done

PR #479, opened 2025-06-12
OCPBUGS-57419: log when resources are updated

dtantsur:


dtantsur:
/hold

This is not exactly what I expected

PR #481, opened 2025-06-13
OCPBUGS-57419: Do not log "new CO status" unnecessarily

dtantsur:
Currently, we log this even when the status is not actually new.


dtantsur:
/hold

y u no work

dtantsur:
/hold cancel

Works as expected

PR #482, opened 2025-06-13
OCPBUGS-56723: Treat ResourceNotFound as non-fatal

honza:
Status ReasonResourceNotFound is not a fatal error, it's a temporary condition while the Metal3 pod does not exist yet.  Setting the operator's status as Degraded is wrong since nothing bad is happening. We set Degraded=False, Progressing=True, Available=False instead.


iurygregory:
the errors are a bit strange

/retest

PR #483, opened 2025-06-16
METAL-1437: Add (unsupported) overrides for the ironic agent image

dtantsur:


dtantsur:
/hold
/cc @zaneb 

@zaneb we discussed this possibility a few weeks ago, wdyt about this approach?

PR #495, opened 2025-07-31
METAL-1508: Change deprecated pointer to ptr

elfosardo:
Also change an instance of io/ioutil to os as it just calls its function.


honza:
I have meant to do this for ages. Thanks!

/lgtm

PR #497, opened 2025-08-06
Add readOnlyRootFilesystem flag to baremetal containers deployed using CBO

hroyrh:
Add readOnlyRootFilesystem flag to containers metal3-httpd, metal3-dnsmasq, metal3-ironic and barmetal-operator. Also add writable mounts for ironic cert, data, conf and tmp mounts since we need these paths are writable for the deployment to succeed.

### DIFF
--- a/provisioning/baremetal_config.go
+++ b/provisioning/baremetal_config.go
@@ -46,6 +46,7 @@ const (
 	bootIsoSource                  = "IRONIC_BOOT_ISO_SOURCE"
 	useUnixSocket                  = "unix"
 	useProvisioningDNS             = "provisioning"
+	customConfigPath               = "CUSTOM_CONFIG_DIR" // TODO(hroy): do we need this
 )
 
 func getDHCPRange(config *metal3iov1alpha1.ProvisioningSpec) *string {
--- a/provisioning/baremetal_pod.go
+++ b/provisioning/baremetal_pod.go
@@ -60,6 +60,12 @@ const (
 	cboOwnedAnnotation               = "baremetal.openshift.io/owned"
 	cboLabelName                     = "baremetal.openshift.io/cluster-baremetal-operator"
 	externalTrustBundleConfigMapName = "cbo-trusted-ca"
+	ironicConfigVolume               = "metal3-ironic-conf"
+	ironicDataVolume                 = "metal3-ironic-data"
+	ironicConfigPath                 = "/conf"
+	ironicDataPath                   = "/data"
+	ironicTmpVolume                  = "metal3-ironic-tmp"
+	ironicTmpPath                    = "/tmp"
 )
 
 var podTemplateAnnotations = map[string]string{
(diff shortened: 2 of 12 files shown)


provisioning/baremetal_config.go
```diff
@@ -46,6 +46,7 @@ const (
 	bootIsoSource                  = "IRONIC_BOOT_ISO_SOURCE"
 	useUnixSocket                  = "unix"
 	useProvisioningDNS             = "provisioning"
+	customConfigPath               = "CUSTOM_CONFIG_DIR" // TODO(hroy): do we need this
```
hroyrh, line 49: I haven't used this var, but I'm just wondering if we need this and some other variables such as `ironicDataPath` as inputs.

provisioning/baremetal_pod.go
```diff
@@ -412,6 +466,11 @@ func createContainerMetal3Dnsmasq(images *Images, config *metal3iov1alpha1.Provi
 		VolumeMounts: []corev1.VolumeMount{
 			sharedVolumeMount,
 			imageVolumeMount,
+			//TEST(hroy): is the cert mount needed here
```
dtantsur, line 469: dnsmasq does not use HTTP(s)

provisioning/baremetal_pod.go
```diff
@@ -86,6 +92,27 @@ var ironicTlsMount = corev1.VolumeMount{
 	ReadOnly:  true,
 }
 
+var ironicCertMount = corev1.VolumeMount{
+	Name:      ironicTlsVolume,
+	MountPath: metal3TlsRootDir + "/ca/ironic",
```
dtantsur, line 97: This is not required to mount, please don't add it

provisioning/image_cache.go
```diff
@@ -93,14 +93,16 @@ func createContainerImageCache(images *Images) corev1.Container {
 		Image:           images.Ironic,
 		ImagePullPolicy: corev1.PullIfNotPresent,
 		SecurityContext: &corev1.SecurityContext{
+			// TODO(hroy): is readOnlyRootFS flag needed here
```
dtantsur, line 96: Ideally yes

provisioning/baremetal_pod.go
```diff
@@ -314,6 +371,9 @@ func createInitContainerStaticIpSet(images *Images, config *metal3iov1alpha1.Pro
 		Command:         []string{"/set-static-ip"},
 		ImagePullPolicy: "IfNotPresent",
 		SecurityContext: &corev1.SecurityContext{
+			ReadOnlyRootFilesystem: ptr.To(true),
```
hroyrh, line 374: Privileged container is needed for it to create a symlink for the host path volume for shared image cache. I have removed the `privileged: true` from the pod and individually added it to containers when needed.

provisioning/baremetal_pod.go
dtantsur: We should fix these mkdirs then

PR #510, opened 2025-09-23
OCPBUGS-60498: Always have a service for ironic-api port

elfosardo:
The current metal3-state service exposes different ports based on whether the proxy is enabled, port 6385 if without proxy, port 6388 (the private port) if with proxy.
Since port 6385 is exposed as a hostPort in both cases, there should always be a service exposing port 6385, regardless of the proxy configuration.

### DIFF
--- a/provisioning/state_service.go
+++ b/provisioning/state_service.go
@@ -36,6 +36,14 @@ func newMetal3StateService(info *ProvisioningInfo) *corev1.Service {
 			Port: int32(port),
 		},
 	}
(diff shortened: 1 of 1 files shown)


sgoveas:
/verified by sgoveas
[Tested with pre-merge build](https://issues.redhat.com/browse/OCPBUGS-60498?focusedCommentId=28131027&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-28131027)

PR #516, opened 2025-10-23
Remove machine-os-images from baremetal pod

honza:
We don't need to run the init container twice.  Only image customization depends on it.  This change also has the side effect of removing two volumes from the baremetal pod.

We introduced the machine-os-images init container in the baremetal pod in 07a9877a314e7c766d214d56067a9f84df779a61.  Then, machine-os-images was moved from the cache to image-customization in 2cede0a3319c8455bda346ac8dc6560cc25fa03b.

### DIFF
--- a/provisioning/baremetal_pod.go
+++ b/provisioning/baremetal_pod.go
@@ -116,8 +116,6 @@ var metal3Volumes = []corev1.Volume{
 		},
 	},
 	imageVolume(),
-	ironicAgentPullSecretVolume(),
-	userCABundleVolume(),
 	{
 		Name: ironicCredentialsVolume,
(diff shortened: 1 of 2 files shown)


zaneb:
> We don't need to run the init container twice. Only image customization depends on it.

I'm not sure that's correct. AFAIK image-customization provides only the ISO (for virtualmedia) and initramfs (for pxe). The kernel and rootfs for PXE must be available to ironic in the shared volume. I believe there may have been a change to make it possible for the image-customization to specify a kernel URL? I'm not sure how we use this. But still, the rootfs image would be missing.

CI failures show virtualmedia working ok for creating workers, but PXE broken.

PR #522, opened 2025-11-06
METAL-1239: Add ExternalIPs field to Provisioning CRD

MahnoorAsghar:
Assisted by: Claude


api/v1alpha1/provisioning_types.go
```diff
@@ -227,6 +227,9 @@ type ProvisioningSpec struct {
 	// ironic-prometheus-exporter container, and creates supporting resources
 	// (ServiceMonitor, Service ports) to expose hardware sensor metrics for Prometheus.
 	PrometheusExporter *PrometheusExporter `json:"prometheusExporter,omitempty"`
+
+	// IronicExternalIP is the external-facing IP address used to access the Ironic service.
```
dtantsur: We probably want to avoid "Ironic" in the name, especially since it's not just Ironic.

dtantsur:
GetIronicIPs needs updating too (instead of GetRealIronicIPs in a previous revision)

provisioning/baremetal_pod.go
```diff
@@ -204,18 +209,29 @@ func setIronicExternalIp(name string, config *metal3iov1alpha1.ProvisioningSpec)
...
-	}
-
+func setIronicExternalIPv6(info *ProvisioningInfo) (corev1.EnvVar, error) {
 	var ironicIPv6 string
 
-	for _, ironicIP := range ironicIPs {
-		if utilnet.IsIPv6String(ironicIP) {
-			ironicIPv6 = ironicIP
-			break
+	externalIPs, ironicIPs, err := GetImageServerIPs(info)
```
dtantsur: See below: you can simplify the output of GetImageServerIPs, and that will simplify this function as well.

provisioning/baremetal_pod_test.go
```diff
@@ -589,3 +589,242 @@ func TestProxyAndCAInjection(t *testing.T) {
...
+				Name:  externalUrlEnvVar,
+				Value: "https://[fd2e:6f44:5dd8:c956::16]:6183",
+			},
+			expectError: false,
+		},
+	}
+
+	for _, tc := range tCases {
+		t.Run(tc.name, func(t *testing.T) {
+			t.Logf("Testing tc : %s", tc.name)
```
dtantsur: nit: not needed (you'll see it in case of failures)

PR #529, opened 2025-11-24
METAL-1664: Remove regular cluster profile from hostedcluster CBO deployment

MahnoorAsghar:
Remove regular cluster profile from hostedcluster CBO deployment.
This should prevent creating the cluster-baremetal-operator-hostedcluster deployment on regular (non-hypershift/hosted clusters) openshift clusters.


hroyrh:
/lgtm
The metal-ipi-bm job seems to be having issues, it passed earlier, but started failing during reruns in my pr #497 as well.

PR #552, opened 2026-02-05
OCPBUGS-75933: Explicitly set ReadOnlyRootFilesystem to false for machine-os-images so that SCC don't end up enforcing it to true

hroyrh:


aygarg-rh:
/verified by aygarg [here](https://issues.redhat.com/browse/OCPBUGS-75933?focusedId=29010258&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-29010258)

PR #586, opened 2026-04-09
OCPBUGS-82531: Bind the plain HTTP 8080 metrics port to localhost

hroyrh:
Right now we bind the metrics server to 0.0.0.0 address which exposes it to all the interfaces, such as pod ip, localhost. This is a security risk, especially from the standpoint of enforcing centralized tls profile work.

With this fix, it will be only exposed on localhost and only the kube-rbac-proxy sidecar, running on HTTPS 8443 port, can listen to it.

Assisted-by: Claude Opus 4.6 <noreply@anthropic.com>

### DIFF
--- a/main.go
+++ b/main.go
@@ -76,7 +76,8 @@ func main() {
 	var imagesJSONFilename string
 
 	klog.InitFlags(nil)
-	flag.StringVar(&metricsAddr, "metrics-addr", ":8080", "The address the metric endpoint binds to.")
+	// Changing default metrics interface to loopback, so that we don't bypass kube-rbac-proxy
+	flag.StringVar(&metricsAddr, "metrics-addr", "[::1]:8080", "The address the metric endpoint binds to.")
 	flag.BoolVar(&enableLeaderElection, "enable-leader-election", false,
 		"Enable leader election for controller manager. Enabling this will ensure there is only one active controller manager.")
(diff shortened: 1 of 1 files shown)


main.go
```diff
@@ -125,6 +126,16 @@ func main() {
...
 		klog.Info("Applying cluster TLS profile to webhook server")
+		// Bind metrics port to localhost so that all the traffic goes through kube-rbac-proxy
+		// rather than explosing the plain HTTP metrics server directly.
+		metricsHost, metricsPort, err := net.SplitHostPort(metricsAddr)
+		if err != nil {
+			klog.ErrorS(err, "invalid metrics address")
+			os.Exit(1)
+		}
+		if metricsHost == "" {
+			metricsAddr = net.JoinHostPort("127.0.0.1", metricsPort)
```
dtantsur: I wonder if an IPv6-only cluster will like 127.0.0.1 (I vaguely recall that ::1 is safer in this regard)

hroyrh: Right, I also wasn't sure how to handle the loopback for both ipv4/v6. I was looking it up and someone mentioned that some tools can't work with `::1` address.

hroyrh: I was looking at this discussion https://groups.google.com/g/golang-nuts/c/iU7Kl2UemDw

PR #587, opened 2026-04-09
OCPBUGS-82526: Fix webhook server not starting due to missing WithValidator

hroyrh:
The controller-runtime upgrade 7ef1340a switched to CustomValidator but omitted the required WithValidator(r) call. Without it, no webhook handlers are registered and port 9443 never opens. FailurePolicy: Ignore caused all validation to silently pass.


dtantsur:
/approve

@hroyrh if the breaking change has been released already, could you file a bug and link it here?

PR #632, opened 2026-07-21
METAL-1795: Add TLS group/curve support

hroyrh:
## Summary by CodeRabbit

* **New Features**
  * Added TLS curve selection from TLS profile groups, propagating to both API and virtual media settings.
  * Apache now receives curve configuration derived from supported groups (with unsupported/forward-compatible groups filtered).
  * BMO now derives Go TLS curve preferences from supported groups and retains curve preferences even with TLS 1.3-only ciphers.
* **Observability**
  * Added logs to report TLS groups that couldn’t be applied for Apache/OpenSSL and BMO/Go.
* **Tests**
  * Expanded unit tests for curve translation and end-to-end propagation to Apache/BMO settings.
* **Chores**
  * Refreshed Kubernetes/OpenShift-related Go dependencies.


hroyrh:
Blocked by https://github.com/openshift/baremetal-operator/pull/510

provisioning/tls_profile.go
```diff
@@ -53,6 +53,53 @@ func splitCiphers(ciphers []string) (tls12, tls13 []string) {
...
+		switch g {
+		case configv1.TLSGroupX25519,
+			configv1.TLSGroupSecP256r1,
+			configv1.TLSGroupSecP384r1,
+			configv1.TLSGroupSecP521r1,
+			configv1.TLSGroupX25519MLKEM768,
+			configv1.TLSGroupSecP256r1MLKEM768,
+			configv1.TLSGroupSecP384r1MLKEM1024:
+			supported = append(supported, string(g))
+		default:
```
hroyrh, line 75: Should we fail in case of unsupported groups?

hroyrh:
@coderabbitai review

provisioning/tls_profile.go
dtantsur: Not sure. It may create a dangerous coupling between API server and us. Do we have any guidance on this topic?

PR #633, opened 2026-07-21
OCPBUGS-99278: Apply cluster TLS profile to ironic-proxy container

hroyrh:
The ironic-proxy container was missed when centralized TLS profile support was added. It terminates TLS via Apache but was not receiving the cluster TLS profile env vars (IRONIC_SSL_PROTOCOL, ciphers), falling back to ironic-image defaults instead of honoring the cluster-wide TLS security profile.

Pass TLSProfileSpec to createContainerIronicProxy() and inject the same tlsProfileToApacheEnvVars() env vars that metal3-httpd and metal3-ironic already receive.

Assisted-By: Claude Opus 4.6 <noreply@anthropic.com>

## Summary by CodeRabbit

* **New Features**
  * Added support for applying optional TLS profiles to the Ironic proxy.
  * When a TLS profile is provided, TLS- and Apache-related settings are automatically injected into the Ironic proxy container configuration, enabling secure connectivity without additional manual setup.

### DIFF
--- a/provisioning/ironic_proxy.go
+++ b/provisioning/ironic_proxy.go
@@ -17,6 +17,7 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/client"
 	"sigs.k8s.io/controller-runtime/pkg/controller/controllerutil"
 
+	configv1 "github.com/openshift/api/config/v1"
 	"github.com/openshift/library-go/pkg/operator/resource/resourceapply"
 	"github.com/openshift/library-go/pkg/operator/resource/resourcemerge"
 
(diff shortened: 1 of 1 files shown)


provisioning/ironic_proxy.go
```diff
@@ -17,10 +17,10 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/client"
 	"sigs.k8s.io/controller-runtime/pkg/controller/controllerutil"
 
+	configv1 "github.com/openshift/api/config/v1"
+	metal3iov1alpha1 "github.com/openshift/cluster-baremetal-operator/api/v1alpha1"
```
dtantsur: nit: normally, local imports are grouped separately, as it was before

PR #640, opened 2026-08-05
OCPBUGS-54957: Remove recreate strategy from BMO deployment

honza:
baremetal-operator runs with leader election enabled so we don't need the recreate strategy to ensure that only one copy of the deployment is running

The recreate strategy causes problems when the node on which the
deployment runs dies.   The cluster wrongly scales the deployment down
to zero.  This change fixes that problem.

## Summary by CodeRabbit

* **Bug Fixes**
  * Deployment rollouts now use the platform’s default strategy instead of forcing full recreation.
  * Updates can be applied more smoothly, helping reduce unnecessary disruption during deployment changes.

### DIFF
--- a/provisioning/bmo_pod.go
+++ b/provisioning/bmo_pod.go
@@ -283,9 +283,6 @@ func newBMODeployment(info *ProvisioningInfo) (*appsv1.Deployment, error) {
 			Replicas: ptr.To[int32](1),
 			Selector: selector,
 			Template: *template,
-			Strategy: appsv1.DeploymentStrategy{
-				Type: appsv1.RecreateDeploymentStrategyType,
-			},
 		},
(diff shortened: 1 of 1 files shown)


zaneb:
Nitpick about the commit message:

> The recreate strategy causes problems when the node on which the
deployment runs dies.   The cluster wrongly scales the deployment down
to zero.

This is what the bug report says, but it is untrue. The issue is that if the BMO Deployment is updated due to the metal3 Pod also moving, a replacement ReplicaSet is not created as long as the previous ReplicaSet still has a Pod that is Terminating, which it will be for quite some time after a node dies. At no time is the Deployment scaled down.

honza:
Commit message updated

PR #641, opened 2026-08-10
OCPBUGS-105455: Fix infinite reconciliation loop in EnsureMirrorConfig

elfosardo:
The IDMS list returned by the API includes server-set metadata (resourceVersion, uid, generation, etc.) that changes on every read. Serializing these fields into the ConfigMap caused ApplyConfigMap to detect a diff on every reconciliation, returning updated=true and preventing the controller from ever progressing past the ConfigMap step to deploy metal3/BMO.

Strip volatile metadata before marshaling so the serialized YAML is stable across API reads. This fixes HA clusters with workers getting stuck with machine-api unavailable because BMO was never deployed.

Assisted-By: Claude Opus 4.6

## Summary by CodeRabbit

* **Bug Fixes**
  * Stabilized mirror configuration processing by ignoring server-generated metadata when saving and comparing configurations.
  * Prevented unnecessary configuration changes caused by volatile resource metadata.
  * Ensured repeated mirror configuration operations produce consistent results and stable hashes.
* **Tests**
  * Added coverage confirming that required names and specifications remain intact while transient metadata is removed.

### DIFF
--- a/provisioning/mirror_config.go
+++ b/provisioning/mirror_config.go
@@ -12,6 +12,7 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/controller/controllerutil"
 	"sigs.k8s.io/yaml"
 
+	osconfigv1 "github.com/openshift/api/config/v1"
 	"github.com/openshift/library-go/pkg/operator/resource/resourceapply"
 )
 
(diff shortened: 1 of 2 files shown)


bfournie:
/lgtm
Thanks for the quick turnaround @elfosardo !

